Issue #362, opened 2014-06-30
Sample operation


mrocklin:
Blaze should support sampling.  Many of the current backends support it naturally.

This requires both a symbolic operation in `blaze/expr/table.py` and implementations for the various backends in `blaze/compute/*.py`

postelrich:
Closing old issues. Feel free to reopen with up to date info.

PR #591, opened 2014-09-04
#362 Sample operation initial work

nevermindewe:
Initial version of Sample() operation implemented and tested for numpy and pandas.
Sample() should randomly select some items from a given datasource and return them.
For numpy and pandas, the returned data are new arrays or dataframes or series.


cowlicks:
xref #362

PR #733, opened 2014-10-11
IsNull and DropNA expressions

### LINKED ISSUE #697: Break isnull type operations out into a separate expression
Currently the `isnan` method uses `np.isnan` or `math.isnan`. We should add an additional `isnull` that checks for null values.

Most backends have something that corresponds to this operation.

cc @quasiben @chdoig @mrocklin

cpcloud:
closes #697


cpcloud:
@mrocklin what's your feeling about this PR, in general?

cpcloud:
i'd be okay with a single function `nonnull` that drops all `null` values instead of a predicate and a `dropna` function. thoughts?

mrocklin:
I would be fine with that.  I think it's probably easier to implement and reason about.

cpcloud:
@llllllllll I'd like to revisit this soon. What do you think about calling `DropNA` `Valid` and removing the `IsNull` expression?

llllllllll:
Why the name change to `Valid`? I think having the predicate is still really useful. For exploring a new data set, it is often nice to see how many missing values you have or something. Also, if we have the predicate, we don't really need a dropna because it just becomes `expr[~expr.isnull()]`

cpcloud:
True. I feel the same way about exploring a new dataset. So let's ... drop ... dropna :) and keep isnull

llllllllll:
Do we have code in place that lets us do an optimzation like `Selection(IsNull)` on `DataFrame` turns into `df.dropna()` or no?

cpcloud:
Yep, you could do that with `optimize`

cpcloud:
I think changing `isnull` to `notnull` is the way to go here. As a user of blaze I think that I'd want to be able to say "give me the data that are `notnull`" rather than say "give me the data that are not `isnull`"

@llllllllll thoughts?

cpcloud:
@llllllllll i don't know if that optimization can be done with existing code

it might be possible to steal some ideas from here to do this: https://github.com/blaze/dask/blob/master/dask/rewrite.py

llllllllll:
I agree that `notnull` is more frequently what you want. You can still express the inverse with a unary not so I am +1.

About the rewrite stuff, this looks really cool. This matches the use case I had a while ago about calling compute_down on each subexpr. I think we are sort of missing this ast-level transformation step.

PR #742, opened 2014-10-14
dynd: add comparison test

mrocklin:
cc @mwiebe


cpcloud:
this could also go in with an xfail

mrocklin:
Historically this was here to drive dynd development.  It has gone stale.  Closing.

PR #758, opened 2014-10-16
relabel uses kwargs

mrocklin:
Before

```
>>> accounts.relabel({'amount': 'balance'})
```

After

```
>>> accounts.relabel(amount='balance')
```


TomAugspurger:
Maybe make a note in the docs that you need to use a dictionary for integer-like fields. e.g. if `d.fields`is `['0', '1']`, `d.relabel(0='balance')` won't work. You need `d.replabe({'0': 'balance'})`

PR #835, opened 2014-11-17
Server thread

mrocklin:
Servers don't have to block

```Python
In [1]: from blaze import Server

In [2]: server = Server({})

In [3]: server.run(separate_thread=True)
Out[3]: <FunctionRunner(Thread-1, started 139788731410176)>
 * Running on http://127.0.0.1:6363/

In [4]: # Control returned
```

### DIFF
--- a/blaze/server/server.py
+++ b/blaze/server/server.py
@@ -15,6 +15,7 @@ from ..api import discover, Data
 from ..expr import Expr, Symbol, Selection, Broadcast, Symbol
 from ..expr.parser import exprify
 from .. import expr
+from ..utils import example, FunctionRunner
 
 from ..compatibility import map
 from datashape import Mono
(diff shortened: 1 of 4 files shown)


cpcloud:
bumping this thread (pun intended!)

going to play with this and would like to get it in

hhuuggoo:
I think easiest way is to write a function which returns a flask app because then people can wrap it with gunicorn.  My opinion is just write the flask part and then others can use whichever webserver and deployment pattern they want

cpcloud:
@hhuuggoo can you show how this can be done? i think it'd be swell to incorporate an example of this in the docs.

mrocklin:
See also https://github.com/ContinuumIO/multiuserblazeserver

hhuuggoo:
https://github.com/ContinuumIO/multiuserblazeserver/blob/master/mbs/app.py#L9

llllllllll:
The way I got multithreading working with blaze was to create a new Flask app and register the blaze blueprint. I then take that new app create a singleton instance of a `gunicorn.app.base.BaseApplication` and call the run method on this new server object. An example looks like:

```python
    from gunicorn.app.base import BaseApplication
    from metautils import Singleton

    from my_code import build_app  # the function that creates the app and registers the API 
    from my_code import gunicorn_config  # gunicorn settings, workers, host, port, etc...

    class app(BaseApplication, metaclass=Singleton()):
        def load(self, *, _underlying=build_app(settings):
            return _underlying

        def load_config(self, *, _cfg=gunicorn_config):
            for k, v in key_map(str.lower, _cfg).items():
                if k in self.cfg.settings and v is not None:
                    self.cfg.set(k.lower(), v)


    app.start()
```

This lets us get multi-threading for free with gunicorn.

cpcloud:
Closing this PR as it's going stale. I think the threaded server issue is worth discussing. I'll open an issue.

PR #909, opened 2014-12-31
Implement numba ufunc via pyfunc

### LINKED ISSUE #912: hash(expr) != hash(expr2) does not imply that expr.isidentical(expr2) is False
somewhat esoteric bug where a `._subs(...)` call isn't substituting the correct values because this predicate is testing `True` when it should be `False` here: https://github.com/ContinuumIO/blaze/blob/master/blaze/compute/core.py#L352-L353

```
ipdb> inputs
[Broadcast(_children=(chunk.a, chunk.b), _scalars=(a, b), _scalar_expr=a < b)]
ipdb> exprs
(Broadcast(_children=(a, b), _scalars=(a, b), _scalar_expr=a < b),)
ipdb> hash(inputs[0]) == hash(exprs[0])
False
ipdb> inputs[0].isidentical(exprs[0])
True
```

cpcloud:
closes #912

### DIFF
--- a/blaze/compute/numba.py
+++ b/blaze/compute/numba.py
@@ -76,7 +76,7 @@ def get_numba_type(dshape):
     >>> get_numba_type(datashape.dshape('10 * {a: int64}'))
     Traceback (most recent call last):
       ...
-    TypeError: Invalid datashape to numba type: dshape("{ a : int64 }")
+    TypeError: Invalid datashape to numba type: dshape("{a: int64}")
 
     See Also
     --------
--- a/blaze/compute/numba.py
+++ b/blaze/compute/numba.py
@@ -4,11 +4,12 @@ import numpy as np
 
 from .core import compute
 from ..expr import Broadcast, symbol, UTCFromTimestamp, DateTimeTruncate
+from ..expr import Map
 from ..dispatch import dispatch
 from toolz import memoize
 import datashape
 import numba
-from .pyfunc import lambdify
+from .pyfunc import funcstr
 
 
 def get_numba_type(dshape):
@@ -132,12 +133,12 @@ def _get_numba_ufunc(expr):
     compute_signature
     """
     leaves = expr._leaves()
-    func = lambdify(leaves, expr)
+    s, scope = funcstr(leaves, expr)
+    scope = {k: numba.jit(nopython=True)(v) if callable(v) else v
+             for k, v in scope.items()}
+    func = eval(s, scope)
     sig = compute_signature(expr)
-
-    # we need getattr(..., 'func', func) for Map expressions which can be passed
-    # directly into numba
-    return numba.vectorize([sig], nopython=True)(getattr(expr, 'func', func))
+    return numba.vectorize([sig], nopython=True)(func)
 
 
 # do this here so we can run our doctest
--- a/blaze/compute/numba.py
+++ b/blaze/compute/numba.py
@@ -1,0 +1,63 @@
+from __future__ import absolute_import, division, print_function
+
+import numpy as np
+
+from .core import compute
+from ..expr import Broadcast, symbol, UTCFromTimestamp, DateTimeTruncate
+from ..dispatch import dispatch
+from toolz import memoize
+import datashape
+import numba
+from .pyfunc import lambdify
+
+
+def get_numba_type(dshape):
+    measure = dshape.measure
+    if measure == datashape.bool_:
+        restype = numba.bool_  # str(bool_) == 'bool' so we can't use getattr
+    elif measure == datashape.date_:
+        restype = numba.types.NPDatetime('D')
+    elif measure == datashape.datetime_:
+        restype = numba.types.NPDatetime('us')
+    elif measure == datashape.timedelta_:
+        restype = numba.types.NPTimedelta(measure.unit)
+    else:
+        restype = getattr(numba, str(measure))
+    return restype
+
+
+def compute_signature(expr):
+    restype = get_numba_type(expr.schema)
+    argtypes = [get_numba_type(e.schema) for e in expr._leaves()]
+    return restype(*argtypes)
+
+
+@memoize
+def get_numba_ufunc(expr):
+    leaves = expr._leaves()
+    func = lambdify(leaves, expr)
+    sig = compute_signature(expr)
+
+    # we need getattr(..., 'func', func) for Map expressions which can be passed
+    # directly into numba
+    return numba.vectorize([sig], nopython=True)(getattr(expr, 'func', func))
+
+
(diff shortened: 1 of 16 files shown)


blaze/compute/numba.py
```diff
@@ -0,0 +1,164 @@
...
+    Examples
+    --------
+    >>> from blaze import symbol
+    >>> s = symbol('s', 'int64')
+    >>> t = symbol('t', 'float32')
+    >>> d = symbol('d', 'datetime')
+
+    >>> expr = s + t
+    >>> compute_signature(expr)
+    float64(int64, float32)
```
mrocklin: These are really nice doctests.  Thanks!

blaze/compute/numba.py
```diff
@@ -0,0 +1,164 @@
...
+    >>> s = symbol('s', 'float64')
+    >>> t = symbol('t', 'float64')
+
+    >>> x = np.array([1.0, 2.0, 3.0])
+    >>> y = np.array([2.0, 3.0, 4.0])
+
+    >>> f = get_numba_ufunc(s + t)
+
+    >>> f(x, y)
+    array([ 3.,  5.,  7.])
```
mrocklin: This is also fantastic.  @seibert and @sklam might be interested in seeing this.

blaze/compute/numba.py
```diff
@@ -0,0 +1,63 @@
...
+
+@memoize
+def get_numba_ufunc(expr):
+    leaves = expr._leaves()
+    func = lambdify(leaves, expr)
+    sig = compute_signature(expr)
+
+    # we need getattr(..., 'func', func) for Map expressions which can be passed
+    # directly into numba
+    return numba.vectorize([sig], nopython=True)(getattr(expr, 'func', func))
```
mrocklin: Can @seibert or @sklam verify that this is the optimal approach and releases the GIL?

blaze/compute/numba.py
```diff
@@ -0,0 +1,164 @@
...
+    --------
+    get_numba_type
+    compute_signature
+    """
+    leaves = expr._leaves()
+    func = lambdify(leaves, expr)
+    sig = compute_signature(expr)
+
+    # we need getattr(..., 'func', func) for Map expressions which can be passed
+    # directly into numba
```
mrocklin: Doesn't `lambdify` cover the map case already?

conda.recipe/meta.yaml
```diff
@@ -45,6 +46,7 @@ requirements:
     - psutil
     - into
     - pymongo
+    - numba
```
cpcloud: i've moved this code to `compute/numba.py`.

blaze/compute/numba.py
```diff
@@ -0,0 +1,164 @@
...
+    return numba.vectorize([sig], nopython=True)(getattr(expr, 'func', func))
+
+
+# do this here so we can run our doctest
+get_numba_ufunc = memoize(_get_numba_ufunc)
+
+
+@dispatch(Broadcast, np.ndarray)
```
mrocklin: We sill need to apply this and optimize to multiple inputs.

Multiple dispatch doesn't handle variadic args.  The best solution I've found is a for loop up to a reasonably high number of inputs.  See https://github.com/ContinuumIO/dask/blob/master/dask/obj.py#L166-L174 for an example.

mrocklin:
This looks like excellent work.  I'm glad the solution came out so simply.

blaze/compute/numba.py
pitrou: Yes, it does release the GIL in nopython mode.

blaze/compute/numba.py
```diff
@@ -0,0 +1,164 @@
...
+    assert len(t._scalars) == 1
+    scalar = t._scalars[0]
+    fields = scalar.fields
+    d = dict((scalar[c], symbol(c, getattr(scalar, c).dshape))
+             for i, c in enumerate(fields))
+    expr = t._scalar_expr._subs(d)
+
+    if isinstance(expr, (UTCFromTimestamp, DateTimeTruncate)):
+        # numba segfaults here
+        return compute(t._scalar_expr, x)
```
mrocklin: What is the cause here.  Can we report this upstream?  Also what happens if expr isn't one of these but one of the sub-expressions is.  Do we still seg-fault?

cpcloud: I've already reported this upstream: https://github.com/numba/numba/issues/917

i'll add a test for a subexpr that's a datetime

mrocklin: Broadcast collect can also be told what operations to bring into the broadcast operation.  Perhaps we need a set of excluded operations as well.

blaze/compute/numba.py
```diff
@@ -0,0 +1,203 @@
...
+
+# do this here so we can run our doctest
+get_numba_ufunc = memoize(_get_numba_ufunc)
+
+
+@dispatch(Broadcast, np.ndarray)
+def compute_up(t, x, **kwargs):
+    assert len(t._scalars) == 1
+    scalar = t._scalars[0]
+    fields = scalar.fields
```
mrocklin: Hrm, we should also test this with non-record arrays.  Datashapes like `10 * 10 * int` will also be a common use case.

blaze/compute/numba.py
cpcloud: @mrocklin without support for record dtypes in `numba`, I don't think there's a way for us to support multiple table like children using numba. Additionally, the `funcstr` generated for a record dtype child is specific to the streaming Python backend implementation, for example if I have an expression

```python
t.a
```
with `dshape`:

```python
{a: int64, b: int64}
```
this generates the following `lambda`:

```python
lambda t: t[1]
```

where it should really be

```python
lambda t: t.a
```

cpcloud: we *could* support multiple children that are *not* record dshaped (i.e., arrays), but i don't think there's a clean, general way to make arrays work with multiple children and table-likes only work when there's a single child. i'm all ears if there's something you think i'm missing

mrocklin: If we remove `Field` from the types that `broadcast_collect` collects in the `optimize` call then maybe we can assume that all children are non-table-like.  

See https://github.com/ContinuumIO/blaze/blob/master/blaze/expr/broadcast.py#L102-L122

mrocklin: I'm also happy to fool around with this once I finish up other things.

mrocklin: I'd actually like to play with it regardless and get a sense for the performance effects.

mrocklin:
I'm +1 on this once everything is in and tests pass.

PR #914, opened 2015-01-09
optimize needs to be defined for multiple args

cpcloud:


cpcloud:
@mrocklin this could use your review when you get a chance

blaze/expr/broadcast.py
```diff
@@ -99,8 +100,8 @@ def scalar_symbols(exprs):
     return scalars
 
 
-Broadcastable = (Arithmetic, Math, Map, Field, DateTime)
-WantToBroadcast = (Arithmetic, Math, Map, DateTime)
+Broadcastable = (Arithmetic, Math, Map, Field, DateTime, USub)
+WantToBroadcast = (Arithmetic, Math, Map, DateTime, USub)
```
mrocklin: `Not` and `USub` often go together.  Perhaps this should be `UnaryOp`?

cpcloud: true, then we can remove `Math`, as `UnaryOp` is a superclass of `Math`

cpcloud: hm but something like `hypot` wouldn't be a unary op nevermind on that last point

mrocklin:
Looks good otherwise.

blaze/expr/broadcast.py
cpcloud: should probably consider decoupling `math` from `unaryop` at some point

mrocklin: That bit of the class hierarchy could definitely be improved.  Not sure what the right way is.

cpcloud:
turns out `Not` wasn't implemented for pyfunc either

cpcloud:
@mrocklin added unaryop and `Not` implementation to `pyfunc`

any other comments?

PR #917, opened 2015-01-10
PR to cleanup a few things in pandas

cpcloud:
* `bcolz` backend now gets its `chunksize` argument from the value passed in or it computes it by taking the minimum `chunklen` of all the columns. the latter method is how `bcolz` itself does it
* `test_pandas_compute.py` now uses either `test_series_equal` or `test_frame_equal` for testing, rather than `assert str(df) == str(df2)`. this exposed a few minor issues that i fixed.
* more doctests
* tests now pass that were failing (in `test_pandas_compute.py`)
* relabeling fields that don't exist raises an error
* map with rename test was redundant, now it's not
* remove unused bcolz chunks function and single associated test,
  this is now in the [`into`](https://github.com/ContinuumIO/into) project
* robustify valid identifier checking
* compute_up(symbol, DataFrame) was removed because it was never executed:
  compute will look for expressions in scope before dispatching on compute_up
  and will subsequently return the frame (because that symbol is in scope)

### DIFF
--- a/blaze/compute/bcolz.py
+++ b/blaze/compute/bcolz.py
@@ -4,8 +4,8 @@ from toolz import curry, concat, first
 from multipledispatch import MDNotImplementedError
 
 from ..expr import (Selection, Head, Field, Projection, ReLabel, ElemWise,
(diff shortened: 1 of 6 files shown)


mrocklin:
+!

PR #925, opened 2015-01-13
Fix numba/numpy TypeError

### LINKED ISSUE #920: Numba type error
A failing example that generates a numba dtype error.  Looks like the types are misaligned.
```Python
In [1]: from blaze import *
x
In [2]: x = np.array([[(4, 180), (4, 184), (4, 188), (4, 192), (4, 196)],
       [(4, 660), (4, 664), (4, 668), (4, 672), (4, 676)],
       [(4, 1140), (4, 1144), (4, 1148), (4, 1152), (4, 1156)],
       [(4, 1620), (4, 1624), (4, 1628), (4, 1632), (4, 1636)],
       [(4, 2100), (4, 2104), (4, 2108), (4, 2112), (4, 2116)]], 
      dtype=[('count', '<i4'), ('total', '<i8')])

In [3]: aggregate = Data(x)

In [4]: compute((sum(aggregate.total, axis=(0,))) / (sum(aggregate.count, axis=(0,))))
TypeError: ufunc 'ufunc' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
> /home/mrocklin/workspace/blaze/blaze/compute/numba.py(208)broadcast_numba()
    206         return compute(t, dict(zip(t._leaves(), data)))
    207     else:
--> 208         return ufunc(*data)
ipdb> print ufunc.types 
['il->d']
ipdb> print data
(array([20, 20, 20, 20, 20]), array([5700, 5720, 5740, 5760, 5780]))
```

cc @cpcloud

cpcloud:
closes #920

depends on ContinuumIO/datashape#125

### DIFF
--- a/blaze/compute/numpy.py
+++ b/blaze/compute/numpy.py
@@ -88,14 +88,17 @@ inat = np.datetime64('NaT').view('int64')
 
 @dispatch(count, np.ndarray)
 def compute_up(t, x, **kwargs):
+    result_dtype = to_numpy_dtype(t.dshape)
     if issubclass(x.dtype.type, (np.floating, np.object_)):
-        return pd.notnull(x).sum(keepdims=t.keepdims, axis=t.axis)
+        return pd.notnull(x).sum(keepdims=t.keepdims, axis=t.axis,
+                                 dtype=result_dtype)
     elif issubclass(x.dtype.type, np.datetime64):
-        return (x.view('int64') != inat).sum(keepdims=t.keepdims, axis=t.axis)
+        return (x.view('int64') != inat).sum(keepdims=t.keepdims, axis=t.axis,
+                                             dtype=result_dtype)
     else:
-        return np.ones(x.shape,
-                       dtype=to_numpy_dtype(t.dshape)).sum(keepdims=t.keepdims,
-                                                           axis=t.axis)
+        return np.ones(x.shape, dtype=result_dtype).sum(keepdims=t.keepdims,
+                                                        axis=t.axis,
+                                                        dtype=result_dtype)
 
 
 @dispatch(nunique, np.ndarray)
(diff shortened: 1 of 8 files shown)


blaze/compute/numpy.py
```diff
@@ -109,7 +112,12 @@ def compute_up(t, x, **kwargs):
 
 @dispatch(Reduction, np.ndarray)
 def compute_up(t, x, **kwargs):
-    return getattr(x, t.symbol)(axis=t.axis, keepdims=t.keepdims)
+    reducer = getattr(x, t.symbol)
+    try:
+        return reducer(axis=t.axis, keepdims=t.keepdims,
+                       dtype=to_numpy_dtype(t.schema))
+    except TypeError:  # some reductions don't take the dtype argument
```
mrocklin: Can we check this with `inspect` as with the `blaze.utils.keywords` function?

blaze/compute/tests/test_sql_compute.py
```diff
@@ -348,11 +348,11 @@ def test_by_summary_clean():
 
 
 def test_by_summary_single_column():
-    expr = by(t.name, n=t.name.count(), biggest=t.name.max())
+    expr = by(t.name, n=t.name.count(), biggest=t.amount.max())
```
mrocklin: The intent of this test is to check a case where the child of the by is a single column.  Previously this was `t.name`, now the child will be just `t`.

mrocklin: If you have an issue with computing the maximum of a string then I suggest changing all of the columns from `t.name` to `t.amount.`

blaze/expr/reductions.py
```diff
@@ -100,7 +101,9 @@ class all(Reduction):
 
 
 class sum(Reduction):
-    pass
+    @property
+    def schema(self):
+        return DataShape(datashape.maxtype(super(sum, self).schema))
```
mrocklin: Should this operate on `self.child` rather than `self`?

mrocklin: nm, I see how it works now.

mrocklin:
+1 other than `test_by_summary_single_column` issue

mrocklin:
Thanks!

PR #932, opened 2015-01-16
Slice by list

mrocklin:
Just reinventing more of NumPy (probably this is a bad sign).

Will be useful for dask.

```Python
In [1]: from blaze import *
x = 
In [2]: x = np.arange(20)

In [3]: d = Data(x)

In [4]: compute(d[[5, 3, 5]])
Out[4]: 
array([5, 3, 5])
```

### DIFF
--- a/blaze/compute/pandas.py
+++ b/blaze/compute/pandas.py
@@ -493,7 +493,7 @@ def compute_up(expr, df, **kwargs):
     index = expr.index
     if isinstance(index, tuple) and len(index) == 1:
         index = index[0]
-    if isinstance(index, _inttypes):
+    if isinstance(index, _inttypes + (list,)):
         return df.iloc[index]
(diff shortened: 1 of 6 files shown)


cpcloud:
should this work for table-like things?

```python
In [11]: from blaze import *

In [12]: df = pd.DataFrame({'a': np.random.randn(5)})

In [13]: t = symbol('t', discover(df))

In [14]: compute(t, df)
Out[14]:
          a
0 -1.058250
1 -0.306955
2  0.988621
3 -2.285035
4  1.309147

In [15]: compute(t[[0, 1]], df)  # NotImplementedError
```

mrocklin:
Presumably it should work for all things.  I'm being pretty narrow about this at the moment.  But yes, will put up a commit supporting Pandas.

PR #934, opened 2015-01-18
Expr supports __array__

mrocklin:
Support standard protocols like `__array__`, `__float__`, etc. to improve standard interoperation with normal Python code.

```Python
In [1]: from blaze import *
x
In [2]: x = np.arange(6)

In [3]: d = Data(x)

In [4]: np.array(d * 10)
Out[4]: array([ 0, 10, 20, 30, 40, 50])

In [5]: int(d.sum())
Out[5]: 15

In [6]: x[:3] = d[3:] * 10

In [7]: x
Out[7]: array([30, 40, 50,  3,  4,  5])
```

Should have done this long ago.


blaze/compute/tests/test_pandas_compute.py
```diff
@@ -40,7 +40,7 @@ def df_all(a_df, b_df):
     This method doesn't check the index which can be manipulated during
     operations.
     """
-    assert all(a_df.columns == b_df.columns)
+    assert np.all(a_df.columns == b_df.columns)
```
cpcloud: Any reason we shouldn't gut this function and just use `tm.assert_frame_equal(a_df, b_df)`? The use of this later in this file looks redundant.

blaze/expr/tests/test_arithmetic.py
```diff
@@ -64,8 +64,8 @@ def test_dir():
     i = symbol('i', '10 * int')
     d = symbol('d', '10 * datetime')
 
-    assert i + 1
-    with pytest.raises(Exception):
+    i + 1  # this works
```
cpcloud: i like having the assert her

blaze/expr/tests/test_collections.py
```diff
@@ -23,7 +24,8 @@ def test_transform():
     expr = transform(t, z=t.x + t.y)
     assert expr.fields == ['x', 'y', 'z']
 
-    assert (t.x + t.y) in expr._subterms()
+    assert builtins.any((t.x + t.y).isidentical(node)
```
cpcloud: this is one reason i don't like star imports, just a nit though

blaze/interactive.py
```diff
@@ -281,9 +281,24 @@ def into(a, b, **kwargs):
...
+
+def intonumpy(data, dtype=None, **kwargs):
+    # TODO: Don't ignore other kwargs like copy
+    result = into(np.ndarray, data)
+    if dtype:
+        result = result.astype(dtype)
+    return result
+
+Expr.__array__ = intonumpy
+Expr.__int__ = lambda x: int(compute(x))
```
cpcloud: i think these could also be written as `into(int)`, etc. not a big deal, just a comment that can probably be ignored

blaze/expr/tests/test_table.py
```diff
@@ -201,7 +189,7 @@ def test_path_issue():
     t2 = transform(t, sizes=t.result.map(lambda x: (x - MIN)*10/(MAX - MIN),
                                          schema='float64', name='size'))
 
-    assert t2.sizes in t2.children
+    assert builtins.any(t2.sizes.isidentical(node) for node in t2.children)
```
cpcloud: should we just override `__contains__` so that `in` works like this?

cpcloud: hm nevermind, we often want to check a specific list of things, not the actual expression

blaze/expr/tests/test_arithmetic.py
mrocklin: What should we assert?

blaze/compute/tests/test_pandas_compute.py
mrocklin: Gutted.

blaze/interactive.py
mrocklin: We don't have `Expr` in the `convert` network.  Hence the explicit call to `compute`.

blaze/expr/tests/test_arithmetic.py
cpcloud: Nothing in particular, what was wrong with what was there? Bare code that isn't assigned to anything or used anywhere increases the likelihood that I will delete it and that could potentially hide regressions. Likely a personal shortcoming of mine, but I think it's easy just to keep the assert

mrocklin: `assert` checks the non-zero-ness of the expressions which now triggers a call to `compute`

cpcloud: How about is not None?

mrocklin: I've added a type check

blaze/expr/tests/test_collections.py
mrocklin: Ironically this would have worked fine if we had star-imported from `blaze` rather than `blaze.expr`.

mrocklin:
ok to merge?

cpcloud:
fine by me

PR #943, opened 2015-01-26
Add pre_compute methods for new into.JSON objects

mrocklin:
We've recently added json support back into `into`.   This supplies `pre_compute` implementations for these types .


mrocklin:
Merging soon if tests pass and no comment.

PR #945, opened 2015-01-27
Iterator is no longer in into

cpcloud:
in anticipation of aws work

we also shouldn't make other projects that depend on into depend on the stdlib
imports that it yields, rather we should import from the stdlib


mrocklin:
Yeah, I need to become less sloppy about imports.

PR #949, opened 2015-01-30
Fix SQL repring bug

cpcloud:
fixes an issue with repring of the sql engine, where calling `t.head(2).head(3)` would take the most recent call rather than the minimum


cpcloud:
@mrocklin any comments? would like to merge this soon if possible

mrocklin:
Looks great to me.  Thanks!

PR #951, opened 2015-02-03
Is in

beauzeaux:
Although I can't think of any usages outside of a selection, subclassing from expressions.Map seems to give a more general solution. 

That being said, it is only used as a selector in sqlalchemy (presumably mongo as well) and I'm not sure if this breaks any future compatibility with those libraries. 

Not sure how to test given it just piggybacks on Map.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,7 @@ blaze.egg-info
 
 
 __conda_version__.txt
+
+# Vagrant related ignores
+Vagrantfile
+.vagrant/*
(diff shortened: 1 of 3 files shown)


mrocklin:
`Map` is not widely supported and when it is it tends to be slow.  If `IsIn` genuinely isn't widely supported then Blaze probably should have it and users should just do what you've done here.

However, my guess is that there are clever ways to do this in various backends.  Things like `semi-join`s in SQL or `pandas.isin` are good examples.

beauzeaux:
Ah, i was unaware that `Map` wasn't widely supported. 

I guess then I need to narrow down the usage of IsIn to only be a Selection.

PR #955, opened 2015-02-08
SparkSQL Refresh

cpcloud:


blaze/compute/tests/conftest.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+
+
+@pytest.fixture(scope='session')
```
mrocklin: The `scope='session'` bit is nice.

cpcloud: yeah, i did this so that `test_spark` and `test_sparksql` can share the instance. In fact, spark whines if you have more than one instance active at the same time.

blaze/compute/sparksql.py
```diff
@@ -1,135 +1,136 @@
...
-    schema = discover(rdd).subshape[0]
-    columns = dshape_to_alchemy(schema)
-    for column in columns:
-        if column.name == primary_key:
-            column.primary_key = True
+@dispatch(Field, SQLContext)
+def compute_up(expr, data, **kwargs):
+    name = expr._name
+    columns = dshape_to_alchemy(expr.dshape)
+    return sa.Table(name, sa.MetaData(), *columns)
```
mrocklin: In this particular case it seems like we're not actually using this in the normal way of `compute_up`.  Could we call it something else and still have the code work the same?

The thing that it returns isn't really the expression mapped on to the data.  We wouldn't be able to recover an actual SparkSQL dataframe here.

blaze/compute/sparksql.py
```diff
@@ -1,135 +1,136 @@
...
-    scope = kwargs.pop('scope', dict())
-    scope = dict((t, q.mapping.get(data, data)) for t, data in scope.items())
+    # compute using sqlalchemy
+    scope = dict(zip(new_leaves, sa_tables))
+    query = compute(expr, scope)
 
-    q2 = compute_up(expr, q.query, scope=scope, **kwargs)
-    return SparkSQLQuery(q.context, q2, q.mapping)
+    # interpolate params
+    qs = str(literalquery(query))
```
mrocklin: We'll want to try to use the `hive` dialect here if possible.

Also, at least in older versions there seemed to be several `sql` methods, some of which may have been hive focused.  Hopefully they've been merged by now, might be worth looking in to.

blaze/compute/sparksql.py
```diff
@@ -1,135 +1,136 @@
...
+    java_import(ctx._jvm, 'scala')
 
-@dispatch(Expr, SparkSQLQuery)
-def post_compute(expr, query, scope=None):
-    result = query.context.sql(sql_string(query.query))
-    if iscollection(expr.dshape) and isscalar(expr.dshape.measure):
-        result = result.map(lambda x: x[0])
-    return result
+    # grab Scala's set converter and convert to a Python set
+    return set(ctx._jvm.scala.collection.JavaConversions.setAsJavaSet(x))
```
mrocklin: In some perverse future universe `into` does this.

blaze/compute/tests/conftest.py
mrocklin: I am intimately familiar with this.

blaze/compute/sparksql.py
cpcloud: ha! only if `into` wants to deal with total opaque `JavaObject`s (this comes from `py4j`). I basically had to manually chain the methods one at a time while simultaneously following the docs for each resulting type to find out how to do this.

blaze/compute/sparksql.py
cpcloud: changed this to `make_sqlalchemy_table`

cpcloud:
@mrocklin whenever you get a chance, this is ready for review

blaze/compute/tests/test_sparksql.py
```diff
@@ -0,0 +1,248 @@
...
+                                                 reduction)())
+    result = compute(expr, ctx)
+    expected = compute(expr, {db: {'t': df}})
+    assert (set(map(frozenset, into(list, result))) ==
+            set(map(frozenset, into(list, expected))))
+
+
+@pytest.mark.parametrize(['reducer', 'reduction'],
+                         itertools.product(['id', 'name'],
+                                           ['count', 'nunique']))
```
mrocklin: The use of parameters seems appropriate to me here.  Might even want to go further with expression up above if it can be done simply.

blaze/compute/sparksql.py
```diff
@@ -1,135 +1,88 @@
...
 
-@dispatch(Head, SparkSQLQuery)
-def post_compute(expr, query, scope=None):
-    result = query.context.sql(sql_string(query.query))
-    if iscollection(expr.dshape) and isscalar(expr.dshape.measure):
-        result = result.map(lambda x: x[0])
-    return result.collect()
+    # interpolate params
+    compiled = literalquery(query, dialect=HiveDialect())
+    return data.sql(str(compiled))
```
mrocklin: I've probably said this already, but this `compute_down` implementation is really slick.

mrocklin:
Seems generally fine to me

blaze/compute/sparksql.py
cpcloud: ah thanks!

cpcloud:
here goes nothing :bomb:!

PR #957, opened 2015-02-10
Scalar results from SQL repr nicely

mrocklin:
The solution feels a bit hacky.

### DIFF
--- a/blaze/interactive.py
+++ b/blaze/interactive.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import datashape
+import datetime
 from datashape import (discover, Tuple, Record, dshape, Fixed, DataShape,
     to_numpy_dtype, isdimension, var)
 from datashape.predicates import iscollection, isscalar, isrecord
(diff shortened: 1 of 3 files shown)


blaze/interactive.py
```diff
@@ -224,7 +233,19 @@ def expr_repr(expr, n=10):
 
     # Scalars
     if ndim(expr) == 0 and isscalar(expr.dshape):
-        return repr(compute(expr))
+        result = compute(expr)
+        if 'float' in str(expr.dshape):
```
cpcloud: i think there's a stronger way to check this:

```python
expr.dshape.measure in datashape.floating.types
```

mrocklin: Hrm, didn't know that those existed.  What about option types?

cpcloud:
```python
getattr(expr.dshape.measure, 'ty', expr.dshape.measure) in datashape.floating.types
```

mrocklin: I'm actually tempted to stick with `'float' in str(expr.dshape)`.  Given that we're under the `isscalar` check this is actually decently strong.  It's also fairly clear.

mrocklin:
I apologize for merging this.  Feeling time pressured.

cpcloud:
no problem, was really just a suggestion

PR #965, opened 2015-02-26
Default pmap

### LINKED ISSUE #964: Allow setting default for map parameter of compute()
It would be nice if it was possible to set a default value for the map parameter of the compute command.

Currently I am always specifying `compute(..., map=pool.map)`. Just specifying `compute(...)` would be cleaner.

I tried setting `blaze.map = pool.map` but this does not seem to have the desired effect.

mrocklin:
This adds a new function allowing users to set a default `pmap` function to be used in chunking computations.

```python
In [1]: import blaze as bz

In [2]: from multiprocessing import Pool

In [3]: p = Pool(4)

In [4]: bz.set_default_pmap(p.map)
```

Fixes #964

### DIFF
--- a/blaze/__init__.py
+++ b/blaze/__init__.py
@@ -25,6 +25,7 @@ from .expr.arrays import (tensordot, transpose)
 from .expr.functions import *
 from .index import create_index
 from .interactive import *
+from .compute.pmap import set_default_pmap
 from .compute.csv import *
 from .compute.json import *
 from .compute.python import *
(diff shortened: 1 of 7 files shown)


cpcloud:
not to bog this down in bureaucracy, but what about having a `settings.py.in` that generates a `settings.py` during `python setup.py install`?

cpcloud:
i imagine that this won't be the only thing that people want to set

mrocklin:
It's reasonable.  I imagine some of the Blaze-rolled solutions will move into dask which will have a better mechanism for this.  To that end I'm a little hesitant to add bureaucracy.

mrocklin:
Whoops, forgot to add pmap file.

mrocklin:
@cpcloud ok to merge?

cpcloud:
Yep

PR #968, opened 2015-02-26
Don't name things that don't have a record dshape

### LINKED ISSUE #966: The .fields attribute should be empty on non-record datashapes
We currently have a default `fields` attribute even on datasets with a homogenous datashape 

```python
In [1]: import blaze as bz

In [2]: bz.Data([1, 2, 3]).fields
Out[2]: ['_1']
```

This should probably just be `[]`

cc @fpliger

cpcloud:
closes #966

### DIFF
--- a/blaze/tests/test_interactive.py
+++ b/blaze/tests/test_interactive.py
@@ -1,5 +1,6 @@
 from blaze.interactive import Data, compute, concrete_head, expr_repr, to_html
 
+import datetime
 from into import into, append
 from into.backends.csv import CSV
 from blaze import discover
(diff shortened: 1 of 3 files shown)


blaze/tests/test_interactive.py
```diff
@@ -332,3 +333,23 @@ def test_scalar_sql_compute():
             dshape=dshape('var * {name: string, amount: int}'))
     d = Data(t)
     assert repr(d.amount.sum()) == '300'
+
+
+def test_no_name_for_simple_data():
+    d = Data([1, 2, 3])
+    assert repr(d) == '    \n0  1\n1  2\n2  3'
+    assert not str(d)
```
mrocklin: It might be reasonable for these to have a string representation in the future (or even right now).  I suggest explicitly asserting `not d._name`

mrocklin:
Looks good

cpcloud:
@mrocklin changed the test, merging after passing

mrocklin:
Great

PR #969, opened 2015-02-27
Doc on column metadata

cpcloud:


mrocklin:
This is good stuff.  Thoughts coming in line shortly.

docs/source/column-metadata.rst
```diff
@@ -0,0 +1,185 @@
...
+Column Metadata
+***************
+
+This document is heavily inspired by the KDB+ DBMS and its driver language ``q``.
+
+In ``q``, tables are constructed like this:
+
+   .. code-block:: sh
+
+      q) t: ([] name: `Alice`Bob`Joe; amount: 100 200 -400)
```
mrocklin: Note to self, always start document with `q` code.  Quickly gets rid of the riff-raff.

docs/source/column-metadata.rst
```diff
@@ -0,0 +1,185 @@
...
+============
+To use a primary key, we construct the table by filling in the square brackets.
+
+   .. code-block:: sh
+
+      q) t: ([id: 1 2 3] name: `Alice`Bob`Joe; amount: 100 200 -400)
+
+Foreign Keys
+============
+Many (if not all) database systems provide a way to relate columns in possibly different tables to each other. When a column in one table refers to a column in another table we call this a *foreign key*.
```
mrocklin: Definitely not all database systems

docs/source/column-metadata.rst
```diff
@@ -0,0 +1,185 @@
...
+      c
+      a
+      b
+      c
+
+
+Other Metadata
+==============
+
+Additionally, KDB+ exposes metadata regarding the structure of a particular column. In particular it allows one to attribute different properties to a column. Those properties are:
```
mrocklin: Might also want to look at SQL constraints http://www.w3schools.com/sql/sql_constraints.asp

docs/source/column-metadata.rst
```diff
@@ -0,0 +1,185 @@
...
+   .. code-block:: python
+
+      >>> t = symbol('t', """var * {
+      ...     id: int64,
+      ...     name: categorical[type=string, categories=["Alice", "Bob", "Joel"]],
+      ...     amount: float64
+      ... }""")
+      >>> t.primary_key = t.id
+      >>> s = symbol('s', "var * {t_id: int64, height: float64}")
+      >>> s.relations = {s.t_id: t.id}
```
mrocklin: I'm generally hesitant about mutation of symbols.

docs/source/column-metadata.rst
```diff
@@ -0,0 +1,185 @@
...
+
+A more general approach might be to pass around a dictionary of arbitrary metadata. This has the benefit of not tying the notion of relations and keys to blaze's ``Symbol`` objects, which are meant to be able to describe many different kinds of data other than tables.
+
+
+   .. code-block:: python
+
+      >>> t = symbol('t', dshape=..., meta={'keys': {'primary': ['id']}})
+      >>> s = symbol('s', dshape=..., meta={'keys': {'foreign': {'t_id': t.id}}})
+
+This approach also has the benefit of being low tech. We're passing around core Python data structures (``list`` and ``dict``) that fully describe primary and foreign key relationships.
```
mrocklin: Unsurprisingly I tend to favor this kind of solution.  Even though things might be arbitrary it would become important to have a few very particular standards and generally supported forms of metadata, like what you've listed above with sorted, distinct, ...

docs/source/column-metadata.rst
```diff
@@ -0,0 +1,185 @@
...
+
+
+Validation can also take place at construction time. A few things can be validated:
+
+* The ``keys`` dictionary key is only valid on ``Record`` dshapes
+* The primary key is actually in the datashape
+* The primary key is a valid primary key type (this would likely be limited to integers at first)
```
mrocklin: Primary keys are often non-integral.

docs/source/column-metadata.rst
```diff
@@ -0,0 +1,185 @@
...
+* Foreign keys all reference primary keys
+
+
+In any case we want to be able to access implicit join functionality:
+
+   .. code-block:: python
+
+      >>> compute(s.t_id.amount, {s: <sdata>, t: <tdata>})
```
mrocklin: It might be worth looking at SQLAlchemy's ORM to see how they spell these things.

docs/source/column-metadata.rst
```diff
@@ -0,0 +1,185 @@
...
+      >>> select t_id.amount from s
+
+In the case of an interactive expression, when we construct objects we need to have tables with foreign keys keep a reference to the tables from which keys originate.
+
+   .. code-block:: python
+
+      >>> t = Data('sqlite:///path/to/db::t')
+      >>> s = Data('sqlite:///path/to/db::s')  # s has a foreign key into t
+
+      >>> compute(s.t_id.name)  # generates SQL similar to the above
```
mrocklin: Presumably things like `Data` would discover the appropriate metadata from the SQL database.  An interesting way to approach building this idea would be to start implementing this function and find out what sorts of things exist on various databases.

mrocklin:
Stating interdependence between multiple symbols seems tricky.  We either have to mutate or create them all at the same time somehow.

Systems like SQLAlchemy bundle all of the tables in to a central metadata instance.  We ended up going this route a bit too with `compute_down` mostly supporting databases as leaves rather than acting on each table independently.  I wonder if it makes sense to localize metadata to only one fairly complex symbol and use that symbol to represent the entire system.

mrocklin:
In general I think that this document could use the second voice of SQL in addition to `q`.  Having both in the mix will probably clarify what is standard and what is unique to one system.

jreback:
I think it would be nice to show a short discource on how ``DataFrames`` (to-some-extent) do this as well.

e.g.

```
In [1]: df = DataFrame(np.random.randn(3,2),columns=list('AB'))         

# in-effect an auto-join foreign key for matching named columns in another table (e.g. will join on the nameed column 'foo')
In [2]: df.index.name = 'foo'

# sorted (or not) and sorting direction
In [4]: df.index.is_monotonic_increasing
Out[4]: True

In [5]: df.index.is_monotonic_decreasing
Out[5]: False

# nullness
In [7]: df.index.hasnans
Out[7]: False

# uniqueness
In [8]: df.index.is_unique
Out[8]: True
```

Other thing to note is that, these are all computed on-demand. These are NOT pre-computed (nor does q I believe). These *can* be expensive to compute so these in general are not *always* available indicators.

docs/source/column-metadata.rst
cpcloud: I'll clarify this by stating at the outset that I'm only talking about relational DBMSs

docs/source/column-metadata.rst
cpcloud: I left this in there to show the evolution of my thoughts, but now that it's in the commit history I'll blow it away. Hooray git!

docs/source/column-metadata.rst
cpcloud: Yep. Once key support has solidified I think enforcing certain metadata would be valuable.

docs/source/column-metadata.rst
cpcloud: I did look at it, just didn't want to open up a can of worms on the first pass :smile:

cpcloud:
> Stating interdependence between multiple symbols seems tricky. We either have to mutate or create them all at the same time somehow.

Indeed, could this be addressed by computing on the database as a leaf? IIRC this is already happening in blaze's SQL land.

PR #978, opened 2015-03-06
Update to use odo

cpcloud:


cpcloud:
@mrocklin this could your review whenever you get a chance.

mrocklin:
How about build requirements in requirements.txt and the conda recipes?  Should we wait on this until things like anaconda are updated?

cpcloud:
ah right

mrocklin:
There is also no need to rush this.  Into is still around.

cpcloud:
i kind of wanted to get everything synced up all at once for projects that depend on blaze and into

jreback:
my 2c I would make this change in say blaze 0.8
eg s major version change

cpcloud:
hm ok

cpcloud:
@jreback @mrocklin i'm going to kick up some dust with this here PR. marked this for 0.8.0 and will tag as such when said dust has settled

mrocklin:
Sounds fine to me.  Merge at will.

PR #980, opened 2015-03-06
Changes to fix windows builds

cpcloud:


mrocklin:
I'm totally just going to +1 anything you do here.  So feel free to bypass the PR system if that will reduce friction.

PR #981, opened 2015-03-07
Proper scalar conversion when blaze server things do not compute to return a scalar

cpcloud:


cpcloud:
@mrocklin can u review when you get a chance? this is necessary for things like sqlalchemy and sparksql, whose backends don't return a `base` object, but need conversion to work correctly with blaze server

mrocklin:
+1

PR #988, opened 2015-03-11
Export absolute value function

cpcloud:


cpcloud:
merging after pass if no comment

shoyer:
There is a built-in function `abs`, corresponding to the `__abs__` special method, so you didn't really need this. No harm, though.

PR #990, opened 2015-03-11
Implement Floordiv for SQL

mrocklin:
Ensure that Math functions operate as expected in SQL

Also implement floor division using a combination of `floor` and division.  This enables truncation operations in SQL.


cpcloud:
@mrocklin can you rehash and push when you get a chance? i think we can get this in

PR #994, opened 2015-03-15
Add string length expression

cpcloud:
```python
In [1]: df = Data(pd.DataFrame({'a': ['Alice', 'Bob', 'Josephine'], 'b': [1, 2, 3]}))

In [2]: df
Out[2]:
           a  b
0      Alice  1
1        Bob  2
2  Josephine  3

In [3]: df.a.strlen()
Out[3]:
   a
0  5
1  3
2  9
```

### DIFF
--- a/blaze/expr/strings.py
+++ b/blaze/expr/strings.py
@@ -3,7 +3,7 @@ from __future__ import absolute_import, division, print_function
 import datashape
 from datashape import String
 from datashape.predicates import isrecord, iscollection
-from .expressions import Expr, schema_method_list
+from .expressions import Expr, schema_method_list, ElemWise
 
 __all__ = ['Like', 'like', 'strlen', 'UnaryStringFunction']
 
@@ -38,21 +38,15 @@ def like(child, **kwargs):
 like.__doc__ = Like.__doc__
 
 
-class UnaryStringFunction(Expr):
+class UnaryStringFunction(ElemWise):
 
     """String function that only takes a single argument.
     """
     __slots__ = '_hash', '_child'
 
-    @property
-    def dshape(self):
-        if iscollection(self._child.dshape):
-            return datashape.var * self._dtype
-        return self._dtype
-
 
 class strlen(UnaryStringFunction):
-    _dtype = datashape.int64
+    schema = datashape.int64
 
 
 def isstring(ds):
--- a/blaze/expr/strings.py
+++ b/blaze/expr/strings.py
@@ -1,11 +1,15 @@
 from __future__ import absolute_import, division, print_function
 
 import datashape
+from datashape import String
+from datashape.predicates import isrecord
 from .expressions import Expr, schema_method_list
 
-__all__ = ['Like', 'like']
+__all__ = ['Like', 'like', 'length', 'StringFunction']
+
 
 class Like(Expr):
+
     """ Filter expression by string comparison
 
     >>> from blaze import symbol, like, compute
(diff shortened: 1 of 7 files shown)


mrocklin:
This looks cool.  Two thoughts:

1.  What other string operations are commonly implemented?
2.  The term `length` might be ambiguous in the case of collections of strings

cpcloud:
Pretty much all of the operations that are on python's strings are implemented in many databases. Things like indexing (the ith element of every string in a column), lowercase-ing uppercase-ing, and containment might be others that are immediately useful.

mysql: http://dev.mysql.com/doc/refman/5.7/en/string-functions.html
oracle: https://docs.oracle.com/cd/E17952_01/refman-5.1-en/string-functions.html
pandas: http://pandas.pydata.org/pandas-docs/dev/text.html
postgres: http://www.postgresql.org/docs/9.3/static/functions-string.html#FUNCTIONS-STRING-OTHER

cpcloud:
i was hoping to slowly add functions, `length` happened to be something that I wanted for hive work

mrocklin:
Filling these out as needed makes sense.

cpcloud:
as far as point 2 is concerned, is there another name that would be more appropriate? `length` isn't a typical API in Python land (usually this is written as `len(x)` or `x.shape[0]` or `x.size`, but all of the databases i listed above call this operation `length`

mrocklin:
This is where I'm envious of pandas' str namespace.  A cheap version might
be stringlength.

I do expect people to use length where they meant count.

cpcloud:
how about `strlength`?

jreback:
you should prob do some kind of accessor

e.g. ``df.a.str.length()`` (and similarly for datetimelikes, e.g. ``.dt``). 

it just makes the API much more intuitive, and allows you to have names that match with function.
granted, now you have an extra accessor in the mix, my 2c

cpcloud:
for datetimes it's easier, because `day`, `month`, `year` can't possibly be anything but what they sound like

cpcloud:
and since we have the type of expression, we can just have those methods/attributes show up only where appropriate

jreback:
true, but string ``len,get,slice,cat`` all are meaninful and yet you might want for something else (in a top-level namespace)

jreback:
my 2c that the API namespace gets cluttered very quickly, and unless you start going the: ``str_len`` route, its problematic (``str_len`` just sucks though)

cpcloud:
ah ha! but it *won't* be cluttered, because we only expose attributes and methods based on the type. e.g., `mean` won't show up at *all* as an attribute on a string column.

cpcloud:
and yes `str_len` is annoying

cpcloud:
looking at the db string functions above, i actually think that `strlength` is one of the few cases where we need to differentiate explicitly in the method name

jreback:
``.join`` is a problem as you have a top-level which is meaninful (well you expose it as a function so maybe it doesn't matter)

cpcloud:
`join` also isn't an operation that happens on a column, not saying there isn't the possibility of confusion, i just think it's lower there because `join` is operation between two tables

cpcloud:
i wonder if having an `str` accessor is possible ... it's repr could even be the list of available string methods

mrocklin:
If most string operations other than length are unambiguous then I'd prefer to keep things flat and just special case length to something like strlength.

cpcloud:
@mrocklin ok 2 merge?

blaze/expr/strings.py
```diff
@@ -33,6 +37,35 @@ def like(child, **kwargs):
...
+
+    """String function that only takes a single argument.
+    """
+    __slots__ = '_hash', '_child'
+
+    @property
+    def dshape(self):
+        if iscollection(self._child.dshape):
+            return datashape.var * self._dtype
+        return self._dtype
```
mrocklin: return DataShape(*[self.shape + [self._dtype]])

Or alternatively define this as a subclass of `Elemwise` and rely on its `dshape` use.  This will also make these operations broadcast nicely with the chunking backends.

mrocklin:
+1 other than the elemwise bit

PR #996, opened 2015-03-19
Move some generic utils out of sparksql.py

cpcloud:


cpcloud:
@mrocklin any comments here? i want to use these in another library that can't depend on spark so i thought moving these out of sparksql would be useful

mrocklin:
Sure.  Seems fine to me.

PR #999, opened 2015-03-23
Add multicolumn sort to sql

cpcloud:


cpcloud:
i'm going to add a test for something that uses the sql backend so that we can confirm this works against a compute resource

cpcloud:
@mrocklin ok 2 merge?

PR #1003, opened 2015-03-25
Allow DateTimes to be groupers in sql

cpcloud:
Also adds a default implementation of SQL date functionality

- [x] use `extract` + `compiles` to generalize date access across the various backends

### DIFF
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -20,6 +20,7 @@ import sqlalchemy as sa
 import sqlalchemy
 from sqlalchemy import sql, Table, MetaData, Column
 from sqlalchemy.sql import Selectable, Select
+from sqlalchemy.ext.compiler import compiles
 from sqlalchemy.sql.elements import ClauseElement, ColumnElement
 from sqlalchemy.engine import Engine
 from operator import and_, eq
@@ -594,7 +595,13 @@ def compute_up(expr, data, **kwargs):
 
 @dispatch(DateTime, (ClauseElement, sa.sql.elements.ColumnElement))
 def compute_up(expr, data, **kwargs):
-    return getattr(sa.func, expr.attr)(data).label(expr._name)
+    return sa.extract(expr.attr, data).label(expr._name)
+
+
+@compiles(sa.sql.elements.Extract, 'hive', 'mysql', 'mssql')
+def compile_extract_to_date_function(element, compiler, **kwargs):
+    func = getattr(sa.func, element.field)(element.expr)
+    return compiler.visit_function(func, **kwargs)
 
 
 @dispatch((Millisecond, Microsecond),
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -32,7 +32,8 @@ from odo.backends.sql import metadata_of_engine
 from ..dispatch import dispatch
 from ..expr import Projection, Selection, Field, Broadcast, Expr
 from ..expr import (BinOp, UnaryOp, USub, Join, mean, var, std, Reduction,
-                    count, FloorDiv, UnaryStringFunction)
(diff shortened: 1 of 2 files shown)


blaze/compute/sql.py
```diff
@@ -585,6 +592,17 @@ def compute_up(expr, data, **kwargs):
     return table_of_engine(data, expr._name)
 
 
+@dispatch(DateTime, (ClauseElement, sa.sql.elements.ColumnElement))
+def compute_up(expr, data, **kwargs):
+    return getattr(sa.func, expr.attr)(data).label(expr._name)
```
mrocklin: Well that was easier than I expected it to be :)

cpcloud: It's a bit bittersweet because sqlalchemy itself doesn't abstract away date component access, so this solution will only work if the RDBMs supports access like this

```sql
select day(t.dates) from t
```

cpcloud: that includes

hive, mysql, and mssql

postgres and oracle do it like this:

```sql
select extract(day from t.dates) from t
```

cpcloud: if there was a way to get the dialect out from an arbitrary sqlalchemy expression then we could support everything

cpcloud: Oh! I just realized I might be able to do this with `compiles`

mrocklin: Select -> Select -> Table via .froms attribute
Table -> engine -> via .bind attribute
engine -> dialect via .dialect attribute

Or there might be a more direct way.  I've saved many an hour by poring over the list of attributes and private attributes on sqlalchemy expressions

mrocklin: Probably you've already seen this, but sqlalchemy does support extract commands http://stackoverflow.com/questions/12019766/how-to-get-month-and-year-from-date-field-in-sqlalchemy

cpcloud: Actually, I hadn't seen that! Excellent. Thanks. So I should be able to do something like this:

1. blaze will compile everything to an `extract` call
2. then we compile extract to different things based on the dialect (default does whatever the default of sqlalchemy does, version below compiles to a function call)

```python
@compiles(Extract, 'mysql', 'hive', 'mssql')
def compile_date_extract(element, compiler, **kwargs):
    return compiler.visit_function(getattr(sa.func, element.field)(ex.expr.effective_value), **kwargs)
```

cpcloud:
@mrocklin this turned out better than i thought it would! thanks for the link

any more comments?

blaze/compute/sql.py
```diff
@@ -585,6 +593,24 @@ def compute_up(expr, data, **kwargs):
     return table_of_engine(data, expr._name)
 
 
+@dispatch(DateTime, (ClauseElement, sa.sql.elements.ColumnElement))
+def compute_up(expr, data, **kwargs):
+    return sa.extract(expr.attr, data).label(expr._name)
+
+
+@compiles(sa.sql.elements.Extract, 'hive')
```
mrocklin: What happens on impala?

cpcloud: impala has an extract function which is what the expression gets compiled to by default:

http://www.cloudera.com/content/cloudera/en/documentation/cloudera-impala/v2-0-x/topics/impala_datetime_functions.html?scroll=datetime_functions

cpcloud: the `compiles` decorator lets you selectively compile things differently based on the dialect and the expression type

mrocklin: Yeah, when I saw that at first I assumed it was one of ours.  Neat to see other dispatching systems out in the wild.

cpcloud:
i think this is ready to go....

@mrocklin any more comments?

PR #1005, opened 2015-03-25
Proper length function

### LINKED ISSUE #1004: Use char_length instead of length
String length is currently using `length` which in a few databases returns the number of bytes in a string. This is problematic for multibyte characters, such as `é`. The proper function to use is `char_length`.

cpcloud:
Closes #1004

depends on #1003

### DIFF
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -569,6 +569,16 @@ string_func_names = {
 }
 
 
+@compiles(sa.sql.functions.GenericFunction, 'hive')
+def compile_char_length_on_hive(element, compiler, **kwargs):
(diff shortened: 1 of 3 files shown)


cpcloud:
This should be fixed in https://github.com/dropbox/PyHive/pull/20, closing. I'll reopen if it won't be fixed there.

cpcloud:
ah darn, nevermind we still need this

cpcloud:
it'll fail until the pyhive PR is merged, or we stick the compilation from the pyhive PR into blaze

cpcloud:
i'm going to implement the pyhive version here and i'll remove it if that gets merged in over there

PR #1009, opened 2015-03-29
Make sure we do not ignore repring failures caused by incorrect computation

cpcloud:


cpcloud:
@mrocklin are you ok with this? we only want to try to repr things as tables that have all these properties

1. one dimensional
2. only record or scalar measure

but we never want to fail, because we'll always show something rather than an error message, which makes it really hard to know what is going wrong when things inevitably *do* go wrong

mrocklin:
Yeah, this seems like a good change to me

cpcloud:
merging on pass

PR #1010, opened 2015-03-30
Arithmetic expressions should use their name as a label if possible

cpcloud:


cpcloud:
@mattr ok with this?

mrocklin:
Looks good.

PR #1013, opened 2015-04-03
Mongo as a first class citizen

### LINKED ISSUE #944: Support Mongo Databases as first class citizens
A while ago we started supporting database like objects as data.  For example with SQLAlchemy we have implementations like 

    @dispatch(Field, Engine)
        ...

And it's now common to give an engine to `Data` and use field access to navigate various tables.

We left behind Mongo when we did this and it has now caught up to us in #941 .  In particular `Data('...::mytable')` now *assumes* that it can turn this in to `Data('...').mytable`.

So, it would be nice if we knew how to operate on Mongo Databases themselves, not just collections.  This generally requires making a `compute_up` function that selects a collection given a database (easy) and a `discover` function which finds all of the collections of a database (also usually easy).  When looking a little while ago at the pymongo api I didn't find a "list all of the valid collections" method, but presumably there is some way to do this.

cpcloud:
Closes #944

### DIFF
--- a/blaze/compute/mongo.py
+++ b/blaze/compute/mongo.py
@@ -354,7 +354,7 @@ datetime_terms = {Day: 'dayOfMonth',
 
 @dispatch(Field, Database)
 def compute_up(expr, data, **kwargs):
-    return MongoQuery(getattr(data, expr._name), [])
+    return getattr(data, expr._name)
 
 
 @dispatch(Expr, Collection)
--- a/blaze/compute/mongo.py
+++ b/blaze/compute/mongo.py
@@ -48,6 +48,7 @@ import numbers
 
 try:
     from pymongo.collection import Collection
+    from pymongo.database import Database
 except ImportError:
     Collection = type(None)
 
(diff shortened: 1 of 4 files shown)


blaze/compute/mongo.py
```diff
@@ -57,27 +58,33 @@
...
 
+
 @dispatch(Expr, Collection)
 def pre_compute(expr, data, scope=None, **kwargs):
     return MongoQuery(data, [])
 
 
+@dispatch(Expr, Database)
+def pre_compute(expr, data, **kwargs):
+    return data
```
mrocklin: Was this required?

blaze/compute/mongo.py
```diff
@@ -347,6 +352,11 @@ def compute_up(t, q, **kwargs):
                   Second: 'second'}
 
 
+@dispatch(Field, Database)
+def compute_up(expr, data, **kwargs):
+    return MongoQuery(getattr(data, expr._name), [])
```
mrocklin: Ideally you would only need to do `return getattr(data, expr._name)`.  Pre_compute should trigger every time the type of the data changes.

mrocklin:
Could spot check that `Data` works as expected as reported in https://github.com/ContinuumIO/blaze/issues/941

blaze/compute/mongo.py
cpcloud: Yes because of the annoying `__iter__` definition on `Database` objects. The pure python backend has a `pre_compute` implementation that dispatches on `Sequence`.

blaze/compute/mongo.py
cpcloud: Yep, I meant to remove this. Done.

cpcloud:
`Data` doesn't work yet. I need to get around this stupid `Iterator` business

cpcloud:
Oh. I see you a have `not_an_iterator` `list`. Excellent

cpcloud:
Fixed:

```
In [1]: d = Data(resource('mongodb://localhost/db'))

In [2]: d
Out[2]:
Data:       Database(MongoClient('localhost', 27017), u'db')
DataShape:  {
  _test_coll: 5 * {amount: int64, id: int64, name: string},
  _test_comprehensive: 5 * {amount: int64, id: int64, name: string},
  coll: 2 * {int: int64, string: string},
  coll2: 2 * {int: int64, string: int64},
  df: 3 * {a: int64, b: string},
  no_date: 3 * {amount: int64, id: int64, name: string},
```

mrocklin:
Woot.

cpcloud:
merging on pass

PR #1017, opened 2015-04-05
Label expressions on Selectables should use sqlalchemy's alias method, not label

cpcloud:


cpcloud:
gah, this is wrong. the labeling actually should be happening with `label` otherwise we don't get the correct name

PR #1019, opened 2015-04-05
Rip out ScalarSelects from Merge operations in the SQL backend

### LINKED ISSUE #1018: Merge operations on sqlalchemy selectables should consolidate scalar selects' from clauses with from clauses in the toplevel select
When doing things like this

```python
t = symbol('t', 'var * {name: string, amount: float64, id: int64}')
subset = t[t.amount < 0]
r = transform(subset, abs_amount=abs(subset.amount))
```
the expression generated by sqlalchemy is something like this

```sql
select name, amount, id, (select abs(amount) as abs_amount from t where amount < 0)
from
t where amount < 0
```

when really what we want is this:

```sql
select name, amount, id, abs(amount) as abs_amount from t where amount < 0
```

cpcloud:
closes #1018

### DIFF
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -536,12 +536,88 @@ def compute_up(t, s, **kwargs):
     return select(columns)
 
 
-@dispatch(Merge, Selectable)
-def compute_up(t, s, **kwargs):
-    subexpression = common_subexpression(*t.children)
-    children = [compute(child, {subexpression: s}, post_compute=False)
-                for child in t.children]
-    return select(children)
+@dispatch(sa.sql.FromClause)
+def get_inner_columns(sel):
+    return list(map(lower_column, sel.c.values()))
+
+
+@dispatch(sa.sql.elements.ColumnElement)
+def get_inner_columns(c):
+    return [c]
+
+
+@dispatch(sa.sql.selectable.ScalarSelect)
+def get_inner_columns(sel):
+    inner_columns = list(sel.inner_columns)
+    assert len(inner_columns) == 1, 'ScalarSelect should have only ONE column'
+    return list(map(lower_column, sel.inner_columns))
+
+
+@dispatch(sa.Table)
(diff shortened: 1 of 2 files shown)


cpcloud:
@mrocklin this could use a review when you get a chance. thanks!

blaze/compute/sql.py
```diff
@@ -524,12 +536,88 @@ def compute_up(t, s, **kwargs):
...
+    # we need these getattrs if data is a ColumnClause or Table
+    return sa.select(columns,
+                     from_obj=get_all_froms(data),
+                     whereclause=getattr(data, '_whereclause', None),
+                     distinct=getattr(data, '_distinct', False),
+                     having=getattr(data, '_having', None),
+                     group_by=get_clause(data, 'group_by'),
+                     limit=getattr(data, '_limit', None),
+                     offset=getattr(data, '_offset', None),
+                     order_by=get_clause(data, 'order_by'))
```
mrocklin: Oh wow, I wonder if this is a better approach to do in lots of places.

blaze/compute/tests/test_sql_compute.py
```diff
@@ -1130,6 +1119,84 @@ def test_join_count():
...
+    WHERE accounts.id = :id_1
+    """)
+
+    try:
+        # we're exactly equal
+        assert result == expected
+    except AssertionError:
+        # we're equal modulo additional transform columns because the **kwargs
+        # of transform are a dict and python functions don't have access to the
+        # order in which arguments were passed to the function
```
mrocklin: This should be deterministic though shouldn't it?  Don't we sort?

cpcloud: Yep, I think we should sort so that we can get a predictable datashape out of a `Merge` expression

cpcloud:
@mrocklin any more comments before merge?

mrocklin:
Nope.

Issue #637, opened 2014-09-16
A Blaze equivalent for: SELECT * WHERE t.column IN list_values


chdoig:
I'd like to have a way to select rows from values in a given list without a Join hack. ;)

An equivalent to SQL:
SELECT * WHERE t.column IN list_values

TomAugspurger:
Just to piggy-back on this issue, an is *not* in operation would be useful too. Right now it seems difficult to recast `not_isin` as a join, like you can (always?) do with `isin`.

```python
In [42]: df1 = pd.DataFrame({'A': [1, 2, 3, 4], 'B': ['a', 'b', 'c', 'd']})

In [43]: df2 = pd.DataFrame({'A': [3, 4, 5]})

In [44]: d1 = Data(df1)

In [45]: d2 = Data(df2)
```

I'm not sure of the best way to get the rows of of `d1` where 'A' is not in `d2` (rows 0 and 1).
With SQL it's something like

```sql
select d1.*
from d1
left join d2
  on d1.a = d2.a
where d2.a is null
order by a;
```

With blaze, a `join(d1, d2, how='left')`, I don't have a chance to add the `where` clause.

I guess this is an anti-join? I'm not too familiar with the relational algebra. Maybe this warrants a new issue/API?

cpcloud:
I think using `join` for this computation is overkill in the case that we want to check membership against a small list of values. For example,

```python
t.column.isin(['a', 'b', 'c'])
```

`['a', 'b', 'c']` isn't a table, so co-opting join in the case would be awkward. Certainly in the case of `t.column.isin(s.column)` we could use join for backends where this expression would be difficult to implement.

mrocklin:
Yeah, I guess I was thinking of `isin` as more of a binary operator with some other expression as in `A[A.column.isin(B.column)]` (though presumably this spelling could be improved).  In that case leveraging `join` is likely computationally ideal.

I'm not sure how to cleanly do the small number of elements case.

mrocklin:
Looks like I was over-thinking it.  I was unaware of `SELECT ... WHERE ... IN ...` statements in SQL

PR #1021, opened 2015-04-06
WIP: Add isin

cowlicks:
xref #637 #951 

This is my attempt at adding an `isin` expression. I still need a lot of help.

### DIFF
--- a/blaze/compute/python.py
+++ b/blaze/compute/python.py
@@ -33,7 +33,7 @@ from datashape.predicates import isscalar, iscollection
 from ..dispatch import dispatch
 from ..expr import (Projection, Field, Broadcast, Map, Label, ReLabel,
                     Merge, Join, Selection, Reduction, Distinct,
-                    By, Sort, Head, Apply, Summary, Like,
+                    By, Sort, Head, Apply, Summary, Like, IsIn,
                     DateTime, Date, Time, Millisecond, ElemWise, symbol,
                     Symbol, Slice, Expr, Arithmetic, ndim, DateTimeTruncate,
                     UTCFromTimestamp)
@@ -488,6 +488,18 @@ def reduce_by_funcs(t):
             return tuple(c(x, y) for c, x, y in zip(combiners, a, b))
 
         return grouper, binop2, combiner, tuple(inits)
+
+
+@dispatch(IsIn, Sequence)
+def compute_up(t, seq, **kwargs):
+    def isin(keys, elem):
+        if elem in keys:
+            return True
+        return False
+    isin_with_keys = partial(isin, t._key)
+    row_isin = lambda x: list(map(isin_with_keys, x))
+    out = list(map(row_isin, seq))
+    return out
 
 
 @dispatch(By, Sequence)
--- a/blaze/expr/collections.py
+++ b/blaze/expr/collections.py
@@ -8,8 +8,8 @@ from datashape.predicates import isscalar, iscollection, isrecord
 from .core import common_subexpression
 from .expressions import Expr, ElemWise, label
 
-__all__ = ['Sort', 'Distinct', 'Head', 'Merge', 'distinct', 'merge',
-           'head', 'sort', 'Join', 'join', 'transform']
+__all__ = ['Sort', 'Distinct', 'Head', 'Merge', 'IsIn', 'distinct', 'merge',
+           'head', 'sort', 'Join', 'join', 'transform', 'isin']
 
 class Sort(Expr):
     """ Table in sorted order
@@ -424,10 +424,33 @@ def join(lhs, rhs, on_left=None, on_right=None, how='inner'):
 
 join.__doc__ = Join.__doc__
 
+class IsIn(Expr):
+    """
+    TODO
+    """
+    __slots__ = '_hash', '_child', '_key'
+
+    @property
+    def dshape(self):
+        """ self._child.dshape * number of matched results
+        """
+        return datashape.var * self._child.dshape.measure
+
+
+    def __str__(self):
+        return "%s.isin({%s})" % (self._child, repr(self._key))
+
+
+def isin(child, key):
+    if not hasattr(key, '__iter__'):
+        key = frozenset([key])
+    key = frozenset(key)
+    return IsIn(child, key)
+
 
 from .expressions import dshape_method_list
 
 dshape_method_list.extend([
-    (iscollection, set([sort, head])),
+    (iscollection, set([sort, head, isin])),
     (lambda ds: len(ds.shape) == 1, set([distinct])),
     ])
--- a/blaze/expr/collections.py
+++ b/blaze/expr/collections.py
@@ -441,11 +441,11 @@ class IsIn(Expr):
         return "%s.isin({%s})" % (self._child, repr(self._key))
 
 
-def isin(child, key):
+def isin(child, key, **kwargs):
     if not hasattr(key, '__iter__'):
         key = frozenset([key])
     key = frozenset(key)
-    return IsIn(child, key)
+    return IsIn(child, key, **kwargs)
 
 
 from .expressions import dshape_method_list
--- a/blaze/compute/numpy.py
+++ b/blaze/compute/numpy.py
@@ -9,7 +9,7 @@ from numbers import Number
 
 from ..expr import Reduction, Field, Projection, Broadcast, Selection, ndim
 from ..expr import Distinct, Sort, Head, Label, ReLabel, Expr, Slice, Join
-from ..expr import std, var, count, nunique, Summary
+from ..expr import std, var, count, nunique, Summary, IsIn
 from ..expr import BinOp, UnaryOp, USub, Not, nelements
 from ..expr import UTCFromTimestamp, DateTimeTruncate
 from ..expr import Transpose, TensorDot
@@ -295,6 +295,11 @@ def compute_up(expr, x, **kwargs):
 @dispatch(TensorDot, np.ndarray, np.ndarray)
 def compute_up(expr, lhs, rhs, **kwargs):
     return np.tensordot(lhs, rhs, axes=[expr._left_axes, expr._right_axes])
+
+
+@dispatch(IsIn, np.ndarray)
+def compute_up(t, ndarr, **kwargs):
+    return np.in1d(ndarr.ravel(), tuple(t._key))
 
 
 @compute_up.register(Join, DataFrame, np.ndarray)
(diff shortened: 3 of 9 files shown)


blaze/expr/collections.py
```diff
@@ -424,10 +424,35 @@ def join(lhs, rhs, on_left=None, on_right=None, how='inner'):
 
 join.__doc__ = Join.__doc__
 
+class IsIn(Expr):
```
cpcloud: The expression I had in mind works like this:

```python
In [1]: def isin(a, b):
   ...:     return [el in b for el in a]
   ...:

In [2]: isin([1, 2, 3], [1, 2, 4, 5, 6])
Out[2]: [True, True, False]
```
That's the python sequence version. The pandas version for `Series` objects would look like this:

```python
In [3]: s = pd.Series([1, 2, 3])

In [4]: s.isin([1, 2, 4, 5, 6])
Out[4]:
0     True
1     True
2    False
dtype: bool
```
So you might have a recipe like this in `python.py`:

```python
@dispatch(IsIn, Sequence)
    key = expr._key
    return [el in key for el in data]
```

blaze/expr/collections.py
```diff
@@ -424,10 +424,35 @@ def join(lhs, rhs, on_left=None, on_right=None, how='inner'):
...
+
+    @property
+    def _len(self):
+        """ number of matched results """
+        pass
+
+    def __str__(self):
+        return "%s.isin(%s)" % (self._child, self._key)
+
+    def _name(self):
```
cpcloud: I think you can leave this method out.

blaze/expr/collections.py
```diff
@@ -424,10 +424,35 @@ def join(lhs, rhs, on_left=None, on_right=None, how='inner'):
...
 join.__doc__ = Join.__doc__
 
+class IsIn(Expr):
+    """
+    """
+    __slots__ = '_hash', '_child', '_key'
+
+    @property
+    def dshape(self):
+        """ self._child.dshape * number of matched results
```
cpcloud: This should be something like

```python
return DataShape(*(self._child.shape + (datashape.bool_,)))
```

The `datashape` programming interface leaves a little to be desired.

blaze/expr/collections.py
```diff
@@ -424,10 +424,35 @@ def join(lhs, rhs, on_left=None, on_right=None, how='inner'):
...
+        """
+        pass
+
+    @property
+    def _len(self):
```
cpcloud: I don't think this is necessary.

blaze/expr/collections.py
cowlicks: I have something similiar [here](https://github.com/cowlicks/blaze/commit/61979fb40486ef78f35ede56b84b4ce09f1654a3#diff-c9f69944277033e73dc99af44b3e7a86R493). But for some reason the breakpoint does not work when I try:
```python
blaze.Data(range(6)).isin([1, 2])
```

blaze/compute/python.py
```diff
@@ -490,6 +490,12 @@ def combiner(a, b):
         return grouper, binop2, combiner, tuple(inits)
 
 
+@dispatch(IsIn, Sequence)
+def compute_up(t, seq, **kwargs):
+    import pdb; pdb.set_trace()
+    return tuple(i for i in t if i in seq)
```
cpcloud: `seq` is the computed result of `t._child`, so you need to do this

```python
key = t._key
return tuple(el for el in seq if el in key)
```

blaze/expr/collections.py
```diff
@@ -424,10 +424,36 @@ def join(lhs, rhs, on_left=None, on_right=None, how='inner'):
...
+    __slots__ = '_hash', '_child', '_key'
+
+    def __init__(self, child, key):
+        super(IsIn, self).__init__(child, frozenset(key))
+
+    @property
+    def dshape(self):
+        """ self._child.dshape * number of matched results
+        """
+        return DataShape(*(self._child.shape + (datashape.bool_,)))
```
cpcloud: this should return something like this:

```python
return var * self.shape[1:] * self._child.dshape
```

blaze/expr/collections.py
```diff
@@ -424,10 +424,36 @@ def join(lhs, rhs, on_left=None, on_right=None, how='inner'):
...
+
+    @property
+    def _len(self):
```
cpcloud: this isn't needed

blaze/expr/collections.py
```diff
@@ -424,10 +424,36 @@ def join(lhs, rhs, on_left=None, on_right=None, how='inner'):
...
+        """ number of matched results """
+        pass
+
+    def __str__(self):
+        return "%s.isin(%s)" % (self._child, self._key)
```
cpcloud: small nitpick:

it might be nice to repr `self` like this:

```
'%s.isin({%s})' % (self._child, ', '.join(map(repr, self._key)))
```
so that `_key` always looks like a `set`.

blaze/expr/collections.py
cpcloud: it's probably easier to do this though

```
from itertools import chain
return DataShape(*chain([var], self._child.shape[1:], [self._child.dshape.measure]))

File: blaze/expr/collections.py [hunk 9]
```diff
@@ -424,10 +424,36 @@ def join(lhs, rhs, on_left=None, on_right=None, how='inner'):
 
 join.__doc__ = Join.__doc__
 
+class IsIn(Expr):
+    """
+    """
+    __slots__ = '_hash', '_child', '_key'
+
+    def __init__(self, child, key):
+        super(IsIn, self).__init__(child, frozenset(key))
```
cpcloud: sorry for the churn, your original idea of putting this in the `isin` function is better :)

blaze/expr/collections.py
cowlicks: I'm basically copying what your telling me here. What does `dshape` need to do? And once I know this, I guess I can figure out how to express what I want by reading the datashape docs?

cowlicks:
So there is some ambiguity about what `isin` should do, should it return a boolean array of the same shape? Should it return the rows that contain matching elements? Should it just return the elements that match?

Suggestionns?

cpcloud:
There are two issues here.

1. I originally thought that it should do what `Series.isin` does. I then changed my mind and thought that it would be nicer to return the elements in the left hand side that are in the right hand side.

2. The issue of what the datashape of this expression will be is dependent on the datashape of `self._child`. If `isin` is called on table-like thing, then the result will be a table like thing. If it's called on vector-like thing then the result will be a vector-like thing etc.

cowlicks:
@cpcloud By 'return the elements' do you mean 'return the rows' of a 2D thing? In the ND case it would be the matching slice along the zeroth dimension?

cpcloud:
I'd probably only support columns and and table-like things (tables have a `Record` datashape) for now.

cpcloud:
I'm not totally sure what `DataFrame.isin` does. I can take a look later.

mwiebe:
I read this first as "integer sin", then I looked a bit closer. :)

cowlicks:
`DataFrame.isin` demo, it returns a dataframe of the same shape as the lhs. Where each element is True/False depending on if it matches the rhs argument.
```python
In [1]: from pandas import DataFrame as df

In [2]: a = df([['alice', 1],['bob', 2], ['alice',3]])

In [3]: a
Out[3]: 
       0  1
0  alice  1
1    bob  2
2  alice  3

In [4]: a.isin(['alice'])
Out[4]: 
       0      1
0   True  False
1  False  False
2   True  False

In [5]: a.isin(['alice', 3])
Out[5]: 
       0      1
0   True  False
1  False  False
2   True   True

In [6]: a.isin(['alice', 2])
Out[6]: 
       0      1
0   True  False
1  False   True
2   True  False
```

cpcloud:
@cowlicks make sure you merge master into your local branch, I've just added a fix for the recent pymongo update that preserves back compat. You'll get a failing build because of this even if the tests relevant to this PR are passing.

cowlicks:
What is the desired API for the SQL isin. Numpy, and Pandas return a column full of boolean's, True if isin, False if not isin. But SQL as it currently is, would just return the matching elements. Which isn't very useful (since you already know the value of the matching elements). And I think returning a column of bools in SQL would be difficult, but it is probably possible. #637 suggests returning the matching rows, but this differs from Numpy and Pandas.

asmeurer:
> I read this first as "integer sin", then I looked a bit closer. :)

Would that be something like signsin?

cpcloud:
One option might be to perform the boolean indexing in the numpy and pandas case, which is similar to how SQL uses `in`:

```python
# pandas
df.loc[df.amount.isin(key)]

# numpy
x[np.in1d(x['amount'], key)]

# sql
select * from t where amount in key
```

cpcloud:
I've not found a use case for column of booleans in SQL that aren't better utilized as the predicate in a where clause.

mrocklin:
For the text comparison `like` operator we chose to unite the downselection.  Seems to have worked well.

cowlicks:
@cpcloud all these would translate to:
```python
# blaze
t.amount.isin(key)
```

@mrocklin "unite the dowselection" means return the whole row?

mrocklin:
> @mrocklin "unite the dowselection" means return the whole row?

Sorry for the delay, but yes, that was the thought.  Could take a look at [like](https://github.com/ContinuumIO/blaze/blob/master/blaze/expr/strings.py#L15-L23).  Please don't take this as anything more than just an idea though.  I don't have strong thoughts on this.

cowlicks:
I'm waiting for input on what the desired API should be, but while trying to implement the "return whole rows" part I had a question:

So pandas' data `blaze_data.field_name.isin(stuff)` dispatches to
```python
@dispatch(IsIn, Series)
def compute_up(expr, s, **kwargs):
```
where `s` is the series. But I want to return subset of the DataFrame the series was taken from. How do I access this?

mrocklin:
The approach taken in in like would be that IsIn (or whatever it would be
called (probably not isin since semantics are different from pandas)) would
operate on the table itself

>>> blaze_data.containing(field_name=[1, 2, 3])

But this decision should probably be informed by whatever makes this
operation easy to implement in systems like sqlalchemy.

cpcloud:
I like `containing`.

One could imagine multiple `in` clauses:

```python
data.containing(a=[1, 2, 3], b=['a', 'b', 'c'])

# rough sqlalchemy:
data.where(data.a.in_([1, 2, 3]) & data.b.in_(['a', 'b', 'c']))
```

cpcloud:
closing in favor of #1034

PR #1022, opened 2015-04-07
WIP: Rip out inner_columns and scalarselects from groupbys on selects

cpcloud:
- [x] check for multiple `froms`
- [x] check that we can handle functions with multiple arguments


cpcloud:
This feels a bit magical. I need to write some more tests and let this mull over a bit before merging.

cpcloud:
@mrocklin can you take a look at this when you get a chance? thanks.

blaze/compute/sql.py
```diff
@@ -487,18 +495,27 @@ def compute_up(t, s, **kwargs):
...
+                                                sa.sql.selectable.Join))):
+        from_obj = s.froms[0]
+    else:
+        from_obj = None
+    return sa.select(columns=columns,
+                     whereclause=getattr(s, 'element', s)._whereclause,
+                     from_obj=from_obj,
+                     distinct=getattr(s, 'element', s)._distinct,
+                     group_by=grouper,
+                     having=None,  # TODO: we don't have an expression for this
```
mrocklin: What would having look like?

mrocklin:
I haven't loaded sqlalchemy back into my head so I'm not able to really certify anything.  Things seem pretty sane though.

blaze/compute/sql.py
cpcloud: names and account totals for people who have more than $1000 in their accounts

```sql
select name, sum(amount) as total from accounts group by name having sum(amount) > 1000
```

blaze could do pretty much the same thing syntactically:

```python
by(t.name, total=t.amount.sum()).having(t.amount.sum() > 1000)
```

cpcloud: We'd also have to require `isinstance(child, By)`.

mrocklin: Hrm, so it's like an after-the-fact `Selection`?

cpcloud: It's like `where` for the sets of tuples in a `group by`, like do the group by, but only return the result of the group by that satisfies the predicate

cpcloud: Said predicate is some kind of reduction

cpcloud:
punted on multiple argument functions and multiple from columns, added `assert`s for both

PR #1027, opened 2015-04-10
Some refactoring in `test_sql_compute.py`

cowlicks:
This just consolidates some of the tests to use the same data and cleans up the imports.

### DIFF
--- a/blaze/compute/tests/test_sql_compute.py
+++ b/blaze/compute/tests/test_sql_compute.py
@@ -19,6 +19,33 @@ from blaze.expr import (symbol, discover, transform, summary, by, sin, join,
 from blaze.compatibility import xfail
 from blaze.utils import tmpfile
 
+
+@pytest.fixture(scope='module')
+def data():
+    # make the engine
+    engine = sa.create_engine('sqlite:///:memory:')
+    metadata = sa.MetaData(engine)
+
+    # name table
+    name = sa.Table('name', metadata,
+                    sa.Column('id', sa.Integer),
+                    sa.Column('name', sa.String),
+                    )
+    name.create()
+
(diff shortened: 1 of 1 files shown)


blaze/compute/tests/test_sql_compute.py
```diff
@@ -1,22 +1,51 @@
...
+                        count, exp, nelements)
 from blaze.compatibility import xfail
-from toolz import unique
-from pandas import DataFrame
-from odo import into, resource
 from blaze.utils import tmpfile
 
+
+@pytest.fixture(scope='module')
+def data():
```
cpcloud: Are we sure there's no inserts on this table during tests that would affect other tests?

cpcloud: If there are, then we need to use a `yield_fixture` with function scope (function scope is the default in pytest fixtures) and drop the table after the `yield` statement

cowlicks: No table inserts or changes that I know of. Maybe I should just switch it now since it might catch someone off guard in the future?

I don't think we're ready to add the *yield* fixture yet, since this just uses an sqlite db.

cpcloud: i like the idea of having a single fixture that makes as few assumptions as possible. A new contributor can simply use the `data` argument and knows that they are getting a fresh database and test table. The reason to use `yield_fixture` is because it lets you clean up.

cpcloud:
@cowlicks olooks like there's a merge conflict

cowlicks:
@cpcloud rebased

cpcloud:
feel free to merge after passing

PR #1031, opened 2015-04-13
Native join in sparksql

### LINKED ISSUE #1012: Spark DataFrame Join should use the native implementation
Right now we do something like this

```python

@dispatch(Join, SparkDataFrame, SparkDataFrame)
def compute_up(expr, lhs, rhs, **kwargs):
    return compute_up(expr, lhs.rdd, rhs.rdd, **kwargs)
```

Which doesn't use the native Spark DataFrame join implementation. I don't think it would be that difficult to use the native implementation. We just have to be careful about how we deal with post-join column selection. The code that does this for plain RDDs can probably be abstracted into a separate function. It may already exist in that form as well.

cpcloud:
closes #1012

### DIFF
--- a/blaze/compute/spark.py
+++ b/blaze/compute/spark.py
@@ -138,13 +138,6 @@ def jgetattr(data, attr, default=None):
         return default
 
 
-@compute_up.register(Join, SparkDataFrame, SparkDataFrame)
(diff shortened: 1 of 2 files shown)


cpcloud:
@mrocklin ok with this?

mrocklin:
Sure

PR #1029, opened 2015-04-13
Add dask array backend

cowlicks:

PR #1035, opened 2015-04-16
Fix for LLVM crashes

### LINKED ISSUE #1030: Blaze can trigger crashes in the LLVM
```python
import numpy as np
 
from odo import discover
from dask.array import from_array
from dask.array.blaze import compute, symbol
 
nx = np.arange(600).reshape((20, 30))
dx = from_array(nx, blockshape=(4, 5))
sx = symbol('x', discover(dx))
 
expr = sx.mean(axis=0)
 
dask_ns = {sx: dx}
 
result = compute(expr, dask_ns)
result2 = np.array(result)
print(result2)
```
We get the output
```bash
python crash_llvm.py 
python: LeaksContext.h:54: void llvm::LeakDetectorImpl<T>::addGarbage(const T*) [with T = llvm::Value]: Assertion `Ts.count(o) == 0 && "Object already in set!"' failed.
[1]    18320 abort      python test.py
```

cowlicks:
This closes #1030 thanks for the comments from @gmarkall

### DIFF
--- a/blaze/compute/numba.py
+++ b/blaze/compute/numba.py
@@ -197,8 +197,6 @@ def _get_numba_ufunc(expr):
     else:
         leaves = expr._leaves()
 
-    # we may not have a Broadcast instance because arithmetic expressions can
-    # be vectorized so we use getattr
     s, scope = funcstr(leaves, expr)
 
     scope = dict((k, numba.jit(nopython=True)(v) if callable(v) else v)
--- a/blaze/compute/numba.py
+++ b/blaze/compute/numba.py
@@ -208,9 +208,8 @@ def _get_numba_ufunc(expr):
     sig = compute_signature(expr)
     # vectorize is currently not thread safe. So lock the thread.
     # TODO FIXME remove this when numba has made vectorize thread safe.
-    lock.acquire()
-    ufunc = numba.vectorize([sig], nopython=True)(func)
-    lock.release()
+    with lock:
+        ufunc = numba.vectorize([sig], nopython=True)(func)
     return ufunc
 
 
--- a/blaze/compute/numba.py
+++ b/blaze/compute/numba.py
@@ -1,14 +1,15 @@
 from __future__ import absolute_import, division, print_function
 
 import numpy as np
+from toolz import memoize
+import datashape
+import numba
+from numba.npufunc.ufuncbuilder import UFuncBuilder
 
 from .core import compute, optimize
 from ..expr import Expr, Arithmetic, Math, Map, UnaryOp
 from ..expr.strings import isstring
 from ..expr.broadcast import broadcast_collect, Broadcast
-from toolz import memoize
-import datashape
-import numba
 from .pyfunc import funcstr
 
 
--- a/blaze/compute/numba.py
+++ b/blaze/compute/numba.py
@@ -1,10 +1,10 @@
 from __future__ import absolute_import, division, print_function
+import threading
 
 import numpy as np
 from toolz import memoize
 import datashape
 import numba
-from numba.npufunc.ufuncbuilder import UFuncBuilder
 
 from .core import compute, optimize
 from ..expr import Expr, Arithmetic, Math, Map, UnaryOp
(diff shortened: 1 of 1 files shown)


blaze/compute/numba.py
```diff
@@ -1,14 +1,15 @@
 from __future__ import absolute_import, division, print_function
 
 import numpy as np
+from toolz import memoize
+import datashape
+import numba
+from numba.npufunc.ufuncbuilder import UFuncBuilder
```
gmarkall: I think this might be a typo - `npyufunc` instead?

cowlicks: Oh yeah, so this was not even run when I tested it >___<

cpcloud:
Can you add a test for this?

mrocklin:
This is kind of a PITA to test.  The error arose when two threads ran the function before compilation, triggering concurrent JIT compilations.  We'll get indirect coverage from the dask PR when it's in.

cpcloud:
ok, fair enough

cpcloud:
@mrocklin can you confirm that this fixes the issue you're having in dask?

cowlicks:
Okay this does it for me. Tested locally:
```bash
(blaze-dev)➜  ~  python crash_llvm.py
[ 285.  286.  287.  288.  289.  290.  291.  292.  293.  294.  295.  296.
  297.  298.  299.  300.  301.  302.  303.  304.  305.  306.  307.  308.
  309.  310.  311.  312.  313.  314.]
```

mrocklin:
@cowlicks does #1029 pass with this fix?

cowlicks:
It does locally, I rebased #1029 on top of this so y'all can see it works with travis (hopefully it will). After this is in I'll rebase #1029 on master again.

blaze/compute/numba.py
```diff
@@ -196,15 +198,20 @@ def _get_numba_ufunc(expr):
...
+    # get the func
     func = eval(s, scope)
+    # get the signature
     sig = compute_signature(expr)
-    return numba.vectorize([sig], nopython=True)(func)
+    # vectorize is currently not thread safe. So lock the thread.
+    # TODO FIXME remove this when numba has made vectorize thread safe.
+    lock.acquire()
+    ufunc = numba.vectorize([sig], nopython=True)(func)
+    lock.release()
```
mrocklin: Just PSA (not a request to change anything) but are you aware that locks function as context managers?

```python
with lock:
    ufunc = ...
```

mrocklin:
@cpcloud ok to merge after 1029 passes tests?

blaze/compute/numba.py
cpcloud: What would happen if `numba.vectorize` raises? Would the lock be held indefinitely? If so, then this should definitely use the context manager functionality.

cpcloud:
Yep, modulo my comment on the context manager issue

mrocklin:
Looks like this hangs

cowlicks:
Yeah, I've been working on it. Still have not figured it out.

cpcloud:
This might be sqlalchemy related.

cpcloud:
I'm looking into it.

cowlicks:
I think so too. When I remove the sql source from `test_comprehensive` it doesn't hang. sparksql stuff fails. And it hangs later on `test_sql_compute.py::test_computation_directly_on_sqlalchemy_tables`

cowlicks:
@cpcloud What changed recently with sqlalchemy? IsIn stuff?

cowlicks:
Oh #1040 is hanging too. Lol I thought this had something to do with the thread lock never releasing this whole time.

cpcloud:
Sqlalchemy recently released 1.0.0

cpcloud:
Not unheard of to have breaking changes in a 1.0 release :)

cpcloud:
@cowlicks i think you need to pull from master, rehash and push:

```sh
git pull upstream master && git commit --amend -C HEAD && git push -f
```

cpcloud:
actually no need to rehash

cpcloud:
closed by #1029

PR #1041, opened 2015-04-19
Versioneer and build things

### LINKED ISSUE #1038: Can't import odo
I am simply trying to ```import odo``` and get the following traceback.  I searched for a bit but came up empty.  Any ideas?

```
<ipython-input-1-cf72f366607d> in <module>()
----> 1 import odo

/usr/lib/python2.7/site-packages/odo/__init__.py in <module>()
     38     from .backends.dynd import nd
     39 with ignoring(ImportError):
---> 40     from .backends import sql
     41 with ignoring(ImportError):
     42     from .backends import mongo

/usr/lib/python2.7/site-packages/odo/backends/sql.py in <module>()
     62 
     63 
---> 64 @discover.register(sa.sql.type_api.TypeEngine)
     65 def discover_typeengine(typ):
     66     if typ in revtypes:

AttributeError: 'module' object has no attribute 'type_api'


```

cpcloud:
closes #1038


cowlicks:
This Travis-Ci failure https://travis-ci.org/ContinuumIO/blaze/jobs/59104346#L1720

prints out code that has since been changed in datashape master. Despite us installing datashape from github. This is because odo does not recognize that the versions installed from github is newer and installs the 0.4.4 release version. Which causes this test to fail.

cowlicks:
Here is where odo overrides the installed version of datashape https://travis-ci.org/ContinuumIO/blaze/jobs/59104346#L334

I'm not sure why it is doing this...?

cpcloud:
@cowlicks this could use your review

blaze/_version.py
```diff
@@ -0,0 +1,191 @@
+
```
cowlicks: This is all generated or copy/pasted from versioneer right? So don't review this module?

cpcloud: yep

cowlicks:
This is more of a meta-question: I personally don't know much about python packaging, but the community seems to say "always choose setuptools over distutils".  So why are we switching to distutils here? Just so we can use versioneer? Are we likely to encounter more drawbacks (like `setup.py develop`) from using distutils?

blaze/compute/tests/test_sql_compute.py
```diff
@@ -1356,3 +1359,43 @@ def test_isin():
...
+    t = symbol('t', discover(data))
+    expr = by(t.ds.year, avg_amt=t.amount.mean())
+    result = str(compute(expr, data))
+
+    # FYI spark sql isn't able to parse this correctly
+    expected = """SELECT
+        EXTRACT(year FROM t.ds) as ds_year,
+        AVG(t.amount) as avg_amt
+    FROM t
+    GROUP BY EXTRACT(year FROM t.ds)
```
cowlicks: Does this query give the same result as the one for sqlalchemy < 1.0?

cpcloud: This is the one that's executed for sqlalchemy < 1.0.0

cpcloud: It's skipped if the installed version is >= 1.0.0

cpcloud: They give the same results.

cpcloud:
I think I've found a way to use setuptools with this

cpcloud:
probably shouldn't have jumped the gun on odo and distutils, since we can likely use the same strategy there

tswicegood:
> I personally don't know much about python packaging, but the community seems to say "always choose setuptools over distutils". So why are we switching to distutils here?

There is a general distrust of `setuptools` in some circles in the Python community (and we have a large portion of those that work at Continuum) and for good reason.  At previous points in its history it's not had a great track record for reliability.  That said, the current codebase is not the same codebase it was a few years ago and it's definitely where the Python community is headed.

It's worth noting that `setuptools` is a requirement of `pip` and now `pip` is automatically installed into any environment where `python` is installed in Conda, so you always have `setuptools` available.

cowlicks:
@tswicegood thanks for the explanation

@cpcloud looks good to me. I'd like to get this in so we can start using CI again. So if we want, we can change back to setuptools in another PR.

cpcloud:
@cowlicks Great thanks!

PR #1049, opened 2015-04-22
Update requirements

llllllllll:
I think that this should be odo 0.3.1 to match the other requirement file.


cpcloud:
yep that's a bug! thanks

cpcloud:
0.3.2 is coming out this week

PR #1054, opened 2015-04-22
Use flask blueprints and application context

llllllllll:
By using blueprints, users can add these routes to another flask application as long as they register the data in the application context.

### DIFF
--- a/blaze/server/server.py
+++ b/blaze/server/server.py
@@ -11,10 +11,11 @@ from contextlib import contextmanager
 import socket
 import json
 from toolz import assoc
-from blaze import into, compute
+from blaze import compute
 from blaze.expr import utils as expr_utils
 from blaze.compute import compute_up
 from datashape.predicates import iscollection, isscalar
+from odo import odo
 from ..interactive import InteractiveSymbol, coerce_scalar
 from ..utils import json_dumps
 from ..expr import Expr, symbol
@@ -281,7 +282,7 @@ def compserver():
         result = compute(expr, {leaf: dataset})
 
         if iscollection(expr.dshape):
-            result = into(list, result)
+            result = odo(result, list)
         elif isscalar(expr.dshape):
             result = coerce_scalar(result, str(expr.dshape))
     except NotImplementedError as e:
--- a/blaze/server/server.py
+++ b/blaze/server/server.py
@@ -2,7 +2,7 @@ from __future__ import absolute_import, division, print_function
 
 try:
     import flask
-    from flask import Flask, request
+    from flask import Blueprint, Flask, request
 except ImportError:
     pass
 
@@ -10,7 +10,6 @@ import blaze
 import socket
 import json
 from toolz import assoc
-from functools import partial, wraps
 from blaze import into, compute
 from blaze.expr import utils as expr_utils
 from blaze.compute import compute_up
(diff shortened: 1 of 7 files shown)


blaze/server/server.py
```diff
@@ -279,6 +279,11 @@ def compserver(dataset):
 
     try:
         result = compute(expr, {leaf: dataset})
+
+        if iscollection(expr.dshape):
+            result = into(list, result)
```
cpcloud: minor nit: for new code we're using `odo`, which is the same as `into` except the argument order is reversed, like this:

```python
into(list, result)  # old code
odo(result, list)  # like cp result list
```

blaze/server/tests/fixtures.py
```diff
@@ -0,0 +1,19 @@
+import pytest
```
cpcloud: instead of creating a `fixtures.py`, you can create a `conftest.py` and `pytest` will automatically look for fixtures there. So, instead of having a function that makes a fixture, just define the fixture in `conftest.py` and use it wherever you want (it has to be from a file in the same directory where `conftest.py` is located).

blaze/server/tests/fixtures.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def make_app_context_fixture(server):
+    """
+    Create an app_context fixture for a module.
+    """
+
+    @pytest.fixture(scope='function')
```
cpcloud: fixtures are function scope by default, so no need to specify this

blaze/server/tests/fixtures.py
cpcloud: ah i see why you have a nested function here, nevermind.

blaze/server/tests/fixtures.py
```diff
@@ -0,0 +1,19 @@
...
+    def app_context(request):
+        """
+        Fixture to mark that this test uses the flask application
+        context.
+        """
+        ctx = server.context()
+        request.addfinalizer(lambda: ctx.__exit__(None, None, None))
+        return ctx.__enter__()
```
cpcloud: can this all be done with `yield_fixture` like this:

```python
@pytest.yield_fixture
def app_context():
    with server.context() as s:
        yield s
```

blaze/server/tests/test_server.py
```diff
@@ -152,9 +157,12 @@ def dont_test_compute_with_namespace():
 
 
 @pytest.fixture
-def iris_server():
+def iris_server(request):
     iris = CSV(example('iris.csv'))
     server = Server(iris)
+    ctx = server.context()
+    request.addfinalizer(lambda: ctx.__exit__(None, None, None))
+    ctx.__enter__()
```
cpcloud: same here, i think this can probably be done with a `yield_fixture`.

llllllllll: here, the server's application context is something that needs to exist outside of the tests; when someone runs their flask app with this blueprint they need to use the correct context.

blaze/server/tests/fixtures.py
llllllllll: That is really cool, I am used to nose but this test runner is really nice

blaze/server/server.py
llllllllll: I can change this; however, I was just moving this code up into the try/except block. The `into` was the existing code.

cpcloud:
looking good!

can you give a small example of what you mean by "registering data in the application context"? would be nice to put it in the blaze server docs, but that isn't strictly necessary. if you want to just put it in this thread i'm okay with that too

will merge after that

thanks!

llllllllll:
I added docs in the server section, let me know if they are clear enough.

cpcloud:
looks good, merging on green

cpcloud:
@llllllllll thanks!

llllllllll:
Thank you

PR #1055, opened 2015-04-22
Use boundmethods instead of partials binding self

llllllllll:
Currently, the `__getattr__` of `InteractiveSymbol` returns `functools.partial` instances that bind `self` to a function. This can be sort of confusing because:

```python
In [4]: ds.apply
Out[4]: 
functools.partial(<function apply at 0x7f5eec192d08>, Data:       Engine(postgresql://localhost/bz)
DataShape:  {
  571980fb9b6b46bcab732e02ba9c854d: var * {
    asof_date: datetime,
    open: ?float64,
    high: ?float64,
    low: ?float64,
    close: ?float64,
  ...)

In [5]: type(ds.apply)
Out[5]: functools.partial
```
Also, this causes the `?` operator in IPython to return the docs for `functools.partial` instead of the function being wrapped.

With this change:

```python
In [3]: ds.apply
Out[3]: 
<bound method InteractiveSymbol.apply of Data:       Engine(postgresql://localhost/bz)
DataShape:  {
  571980fb9b6b46bcab732e02ba9c854d: var * {
    asof_date: datetime,
    open: ?float64,
    high: ?float64,
    low: ?float64,
    close: ?float64,
  ...>

In [4]: type(ds.apply)
Out[4]: method
```
This causes the `?` to resolve correctly.

### DIFF
--- a/blaze/compatibility.py
+++ b/blaze/compatibility.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import sys
+from types import MethodType
 
 PY3 = sys.version_info[0] == 3
 PY2 = sys.version_info[0] == 2
@@ -46,11 +47,15 @@ if PY2:
     unicode = builtins.unicode
     basestring = builtins.basestring
     _strtypes = (str, unicode)
+
+    def boundmethod(func, instance):
+        return MethodType(func, instance, type(instance))
 else:
     _inttypes = (int,)
     _strtypes = (str,)
     unicode = str
     basestring = str
+    boundmethod = MethodType
 
 
 import io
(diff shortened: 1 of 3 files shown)


cpcloud:
cool can you add a tiny smoke test?

llllllllll:
Sure, what would you like it to test though, just assert that it is a bound method of the instance?

cpcloud:
yep

llllllllll:
I am not familiar with the test suite yet; do you have a suggestion of where to put this test?

cpcloud:
yep, just put it in `blaze/tests/test_interactive.py`

llllllllll:
Test added

cpcloud:
woot, merging

PR #1058, opened 2015-04-23
str arithmetic

### LINKED ISSUE #1045: ENH: concat on add for string types
Something like this

```python
from blaze import  Data

In [1]: from blaze import Data

In [2]: d = Data([('a', 'b'), ('a', 'b')], fields=['a', 'b'])

In [3]: d
Out[3]: 
   a  b
0  a  b
1  a  b

In [4]: d.a + d.b
```

Would produce

```python
   
0  ab
1  ab
```

right now we get an AttributeError

```python
AttributeError                            Traceback (most recent call last)
<ipython-input-4-7be7215fd14d> in <module>()
----> 1 d.a + d.b

/Users/tom.augspurger/Envs/py3/lib/python3.4/site-packages/blaze/expr/core.py in __add__(self, other)
    155 
    156     def __add__(self, other):
--> 157         return self._add(other)
    158 
    159     def __radd__(self, other):

/Users/tom.augspurger/Envs/py3/lib/python3.4/site-packages/blaze/expr/expressions.py in __getattr__(self, key)
    163             pass
    164         try:
--> 165             result = object.__getattribute__(self, key)
    166         except AttributeError:
    167             fields = dict(zip(map(valid_identifier, self.fields),

AttributeError: 'Field' object has no attribute '_add'
```

llllllllll:
Closes #1045

I wanted to try to learn how some of the symbolic stuff worked so I figured this was an easy one to tackle.

```python
In [1]: import blaze as bz

In [2]: ds = bz.Data([('%s', 'a'), ('%s', 'b')], fields=['a', 'b'])

In [3]: ds.a + ds.b
Out[3]: 
      
0  %sa
1  %sb

In [4]: ds.a * 2
Out[4]: 
      a
0  %s%s
1  %s%s

In [5]: ds.a % ds.b
Out[5]: 
    
0  a
1  b

```

Let me know if I this needs more tests or the tests are in the wrong place.

### DIFF
--- a/blaze/compute/numpy.py
+++ b/blaze/compute/numpy.py
@@ -56,45 +56,56 @@ for i in range(2, 6):
 
 
 @dispatch(Repeat, np.ndarray)
-def compute_up(t, data, **kwargs):
+def compute_up(t, data, _char_mul=np.char.multiply, **kwargs):
     if isinstance(t.lhs, Expr):
-        return np.char.multiply(data, t.rhs)
+        return _char_mul(data, t.rhs)
     else:
-        return np.char.multiply(t.lhs, data)
+        return _char_mul(t.lhs, data)
 
 
 @compute_up.register(Repeat, np.ndarray, (np.ndarray, base))
 @compute_up.register(Repeat, base, np.ndarray)
-def compute_up_np_repeat(t, lhs, rhs, **kwargs):
-    return np.char.multiply(lhs, rhs)
+def compute_up_np_repeat(t, lhs, rhs, _char_mul=np.char.multiply, **kwargs):
+    return _char_mul(lhs, rhs)
 
 
 @dispatch(Concat, np.ndarray)
-def compute_up(t, data, **kwargs):
+def compute_up(t, data, _char_add=np.char.add, **kwargs):
     if isinstance(t.lhs, Expr):
-        return np.char.add(data, t.rhs)
+        return _char_add(data, t.rhs)
     else:
-        return np.char.add(t.lhs, data)
+        return _char_add(t.lhs, data)
 
 
 @compute_up.register(Concat, np.ndarray, (np.ndarray, base))
 @compute_up.register(Concat, base, np.ndarray)
-def compute_up_np_concat(t, lhs, rhs, **kwargs):
-    return np.char.add(lhs, rhs)
+def compute_up_np_concat(t, lhs, rhs, _char_add=np.char.add, **kwargs):
+    return _char_add(lhs, rhs)
+
+
+def _interp(arr, v, _Series=pd.Series, _charmod=np.char.mod):
+    """
+    Delegate to the most efficient string formatting technique based on
+    the length of the array.
+    """
+    if len(arr) >= 145:
+        return _Series(arr) % v
+
+    return _charmod(arr, v)
 
 
 @dispatch(Interp, np.ndarray)
 def compute_up(t, data, **kwargs):
     if isinstance(t.lhs, Expr):
-        return np.char.mod(data, t.rhs)
+        return _interp(data, t.rhs)
     else:
-        return np.char.mod(t.lhs, data)
+        return _interp(t.lhs, data)
 
 
 @compute_up.register(Interp, np.ndarray, (np.ndarray, base))
 @compute_up.register(Interp, base, np.ndarray)
 def compute_up_np_interp(t, lhs, rhs, **kwargs):
-    return np.char.mod(lhs, rhs)
+    return _interp(lhs, rhs)
 
 
 
--- a/blaze/compute/numpy.py
+++ b/blaze/compute/numpy.py
@@ -63,13 +63,9 @@ def compute_up(t, data, **kwargs):
         return np.char.multiply(t.lhs, data)
 
 
-@dispatch(Repeat, np.ndarray, (np.ndarray, base))
-def compute_up(t, lhs, rhs, **kwargs):
-    return np.char.multiply(lhs, rhs)
-
-
-@dispatch(Repeat, base, np.ndarray)
-def compute_up(t, lhs, rhs, **kwargs):
+@compute_up.register(Repeat, np.ndarray, (np.ndarray, base))
+@compute_up.register(Repeat, base, np.ndarray)
+def compute_up_np_repeat(t, lhs, rhs, **kwargs):
     return np.char.multiply(lhs, rhs)
 
 
@@ -81,13 +77,9 @@ def compute_up(t, data, **kwargs):
         return np.char.add(t.lhs, data)
 
 
-@dispatch(Concat, np.ndarray, (np.ndarray, base))
-def compute_up(t, lhs, rhs, **kwargs):
-    return np.char.add(lhs, rhs)
-
-
-@dispatch(Concat, base, np.ndarray)
-def compute_up(t, lhs, rhs, **kwargs):
+@compute_up.register(Concat, np.ndarray, (np.ndarray, base))
+@compute_up.register(Concat, base, np.ndarray)
+def compute_up_np_concat(t, lhs, rhs, **kwargs):
     return np.char.add(lhs, rhs)
 
 
@@ -99,13 +91,9 @@ def compute_up(t, data, **kwargs):
         return np.char.mod(t.lhs, data)
 
 
-@dispatch(Interp, np.ndarray, (np.ndarray, base))
-def compute_up(t, lhs, rhs, **kwargs):
-    return np.char.mod(lhs, rhs)
-
-
-@dispatch(Interp, base, np.ndarray)
-def compute_up(t, lhs, rhs, **kwargs):
+@compute_up.register(Interp, np.ndarray, (np.ndarray, base))
+@compute_up.register(Interp, base, np.ndarray)
+def compute_up_np_interp(t, lhs, rhs, **kwargs):
     return np.char.mod(lhs, rhs)
 
 
--- a/blaze/compute/numpy.py
+++ b/blaze/compute/numpy.py
@@ -10,7 +10,7 @@ from numbers import Number
 from ..expr import Reduction, Field, Projection, Broadcast, Selection, ndim
 from ..expr import Distinct, Sort, Head, Label, ReLabel, Expr, Slice, Join
 from ..expr import std, var, count, nunique, Summary, IsIn
-from ..expr import BinOp, UnaryOp, USub, Not, nelements
+from ..expr import BinOp, UnaryOp, USub, Not, nelements, Repeat, Concat, Interp
 from ..expr import UTCFromTimestamp, DateTimeTruncate
 from ..expr import Transpose, TensorDot
 from ..utils import keywords
@@ -53,6 +53,61 @@ except ImportError:
 compute_up.register(Broadcast, np.ndarray)(broadcast_ndarray)
 for i in range(2, 6):
     compute_up.register(Broadcast, *([(np.ndarray, Number)] * i))(broadcast_ndarray)
+
+
+@dispatch(Repeat, np.ndarray)
+def compute_up(t, data, **kwargs):
+    if isinstance(t.lhs, Expr):
+        return np.char.multiply(data, t.rhs)
+    else:
+        return np.char.multiply(t.lhs, data)
+
+
+@dispatch(Repeat, np.ndarray, (np.ndarray, base))
+def compute_up(t, lhs, rhs, **kwargs):
+    return np.char.multiply(lhs, rhs)
+
+
+@dispatch(Repeat, base, np.ndarray)
+def compute_up(t, lhs, rhs, **kwargs):
+    return np.char.multiply(lhs, rhs)
+
+
+@dispatch(Concat, np.ndarray)
+def compute_up(t, data, **kwargs):
+    if isinstance(t.lhs, Expr):
+        return np.char.add(data, t.rhs)
+    else:
+        return np.char.add(t.lhs, data)
+
+
+@dispatch(Concat, np.ndarray, (np.ndarray, base))
+def compute_up(t, lhs, rhs, **kwargs):
+    return np.char.add(lhs, rhs)
+
+
+@dispatch(Concat, base, np.ndarray)
+def compute_up(t, lhs, rhs, **kwargs):
+    return np.char.add(lhs, rhs)
+
+
+@dispatch(Interp, np.ndarray)
+def compute_up(t, data, **kwargs):
+    if isinstance(t.lhs, Expr):
+        return np.char.mod(data, t.rhs)
+    else:
+        return np.char.mod(t.lhs, data)
+
+
+@dispatch(Interp, np.ndarray, (np.ndarray, base))
+def compute_up(t, lhs, rhs, **kwargs):
+    return np.char.mod(lhs, rhs)
+
+
+@dispatch(Interp, base, np.ndarray)
+def compute_up(t, lhs, rhs, **kwargs):
+    return np.char.mod(lhs, rhs)
+
 
 
 @dispatch(BinOp, np.ndarray, (np.ndarray, base))
--- a/blaze/compute/pandas.py
+++ b/blaze/compute/pandas.py
@@ -79,15 +79,15 @@ def compute_up(t, s, **kwargs):
 @dispatch(Interp, Series)
 def compute_up(t, data, **kwargs):
     if isinstance(t.lhs, Expr):
-        return pd.Series(data.values % t.rhs)
+        return data % t.rhs
     else:
-        return pd.Series(t.lhs % data.values)
+        return t.lhs % data
 
 
 @compute_up.register(Interp, Series, (Series, base))
 @compute_up.register(Interp, base, Series)
 def compute_up_pd_interp(t, lhs, rhs, **kwargs):
-    return pd.Series(lhs.values % rhs.values)
+    return lhs % rhs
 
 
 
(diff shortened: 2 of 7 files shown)


blaze/expr/tests/test_arithmetic.py
```diff
@@ -72,3 +72,15 @@ def test_dir():
 def test_arith_ops_promote_dtype():
     r = w + z
     assert r.dshape == dshape('5 * 3 * float64')
+
+
+def test_str_arith():
+    assert isinstance(cs + cs, Add)
+    assert isinstance(cs * 1, Mult)
+    assert isinstance(cs % cs, Mod)
```
mrocklin: I wonder if we should have separate string operations rather than overload `Add`.  The fact that these are the same in our minds is mostly a Python thing but may not carry over well to other backends.  It may be more clear to have a string concatenation operator.  Being explicit in this way might help when building backends because we won't have to check the dtype in `Add` operations for this case.

llllllllll: I could do `Cat`, `Format`, and `Repeat` to represent these new actions, how does that sound?

cpcloud: i'd call `Format` `Interp` because `format` is a method that does string formatting in a very different way

cpcloud:
Usually when we add a new expression we add a couple of implementations using a few backends. Depending on what you're comfortable with, pandas and sql might be a nice pair to implement. after that, we can implement others piecemeal

llllllllll:
How should I go about adding the implement these; with pandas and sql, all three operators are already defined to act element-wise on the values so they would not be using these new nodes.

cpcloud:
if we go the methods rather than operators route then you can do something like this for pandas (in `compute/pandas.py`):

```python
@dispatch(Concat, pd.Series)
    return data.str.concat(expr.rhs)
```

there might be some hoop jumping required, take a look at the `BinOp` implementations in the same file

llllllllll:
Should the method names be: `cat`, `interp` and `repeat`?

cpcloud:
`cat` -> `concat`, others look good

llllllllll:
Should strings still be able to use the arithmetic operators or should you just use these new methods.

llllllllll:
Should I move the Concat and Repeat into the expr.collections and Interp into string?

cpcloud:
if there's a way that other backends such as sql and others that don't implement this kind of syntax can use it, then i'd be okay with allowing the operators

cpcloud:
> Should I move the Concat and Repeat into the expr.collections and Interp into string?

sounds good.

blaze/expr/tests/test_arithmetic.py
```diff
@@ -72,3 +72,15 @@ def test_dir():
 def test_arith_ops_promote_dtype():
     r = w + z
     assert r.dshape == dshape('5 * 3 * float64')
+
+
+def test_str_arith():
+    assert isinstance(cs + cs, Concat)
+    assert isinstance(cs * 1, Repeat)
+    assert isinstance(cs % cs, Interp)
```
cpcloud: nice. this should allow backends to use this syntax even if they don't know anything about it

blaze/expr/strings.py
```diff
@@ -54,7 +55,14 @@ def isstring(ds):
     return isinstance(getattr(measure, 'ty', measure), String)
 
 
+_add, _radd = _mkbin('add', Concat)
+_mod, _rmod = _mkbin('mod', Interp)
+_mul, _rmul = _mkbin('mul', Repeat)
```
cpcloud: nice. this should allow backends to use this syntax even if they don't know anything about it

llllllllll:
@cpcloud in your example, you have `Concat` doing elementwise string concat for dataframes; should it not call `pd.concat`?

llllllllll:
I am having some trouble figuring out how to get a series of strings to qualify for the string ops. I started with numpy though. Also, I am not super experienced with sql so I am not sure where to start for the implementation for that back end.

llllllllll:
@cpcloud, wondering if this looks okay with the numpy stuff and if you have any ideas for getting pandas to work

blaze/compute/numpy.py
```diff
@@ -55,6 +55,61 @@ def broadcast_ndarray(t, *data, **kwargs):
...
+
+
+@dispatch(Interp, np.ndarray, (np.ndarray, base))
+def compute_up(t, lhs, rhs, **kwargs):
+    return np.char.mod(lhs, rhs)
+
+
+@dispatch(Interp, base, np.ndarray)
+def compute_up(t, lhs, rhs, **kwargs):
+    return np.char.mod(lhs, rhs)
```
cpcloud: You can consolidate this function and the one above by rewriting it like this:

```python
@compute_up.register(Interp, base, np.ndarray)
@compute_up.register(Interp, np.ndarray, (np.ndarray, base))
def compute_up_np_char_mod(t, lhs, rhs, **kwargs):
    return np.char.mod(lhs, rhs)
```

Same for the above `Concat` and `Repeat` implementations as well.

cpcloud:
@llllllllll Can you add some tests in `test_numpy_compute.py` for the 3 operations that you've implemented?

blaze/compute/numpy.py
llllllllll: ah, I wasn't sure if you could do this; I was using the other models as an example. Would it make sense to add a style commit to refactor the neighbouring compute functions to do this?

cpcloud:
As for Pandas, you can do these operations with the following operations:

- `Concat`: `pd.Series.str.concat`
- `Repeat`: `pd.Series.str.repeat`
- `Interp`: use the modulo operator

llllllllll:
Sorry; I had no issue writing to the compute functions; however, the methods were not appearing because I am not sure how to tag the pandas structures with these expression types.

cpcloud:
Oh, do it like this:

```python
@dispatch(Concat, pd.Series)
    # do stuff
```

blaze/compute/numpy.py
cpcloud: that's fine. you can also just lump that change into another commit if that's more convenient

llllllllll: AttributeError: 'Dispatcher' object has no attribute '__name__'

this doesn't seem to work

llllllllll:
```python
In [2]: ds = bz.Data(pd.Series(['a', 'b', 'c']))

In [4]: 'repeat' in dir(ds)
Out[4]: False
```

I am not seeing the methods to create the expressions that the compute_up will dispatch on. I think this is because the dhape of a series with strings is just object and it fails the isstring check

blaze/compute/numpy.py
mrocklin: It won't work with the `@dispatch` decorator.  It should work if you use the `register` method on the Dispatcher itself.

https://github.com/mrocklin/multipledispatch/blob/master/multipledispatch/dispatcher.py#L74-L103

cpcloud: You can't do this

```python
@compute_up.register(...)
    # do stuff
```
You have to call it something besides `compute_up` if you use register

cpcloud: Oh, also try updating your version of `multipledispatch` by either `pip install -U git+git://github.com/mrocklin/multipledispatch` or `conda update multipledispatch`

cowlicks:
@llllllllll I think you need to add them to `schema_method_list` here: https://github.com/quantopian/blaze/blob/str-catenation/blaze/expr/arithmetic.py#L413

blaze/compute/numpy.py
cpcloud: @mrocklin Why wouldn't this work with `@dispatch`?

mrocklin: The first time we call `dispatch` on a new function we need the Dispatcher to be in the module's namespace, not the original function.  That way when someone says `from foo import myfunc` they get the Dispatcher.

mrocklin: This naming bit isn't an issue if you create the dispatcher explicitly:

    myfunc = Dispatcher('myfunc')
    
    @myfunc.register(...)
    def some_other_name(...):
        ...

blaze/expr/strings.py
```diff
@@ -54,7 +55,14 @@ def isstring(ds):
...
+
+
 schema_method_list.extend([
+    (isstring, set([_add, _radd, _mod, _rmod, _mul, _rmul,
```
llllllllll: @cowlicks I am adding them here because they act on strings, the issue is that that the dtype of a pandas series of strings is just 'object', not 'string'

blaze/expr/tests/test_arithmetic.py
```diff
@@ -72,3 +72,15 @@ def test_dir():
...
+
+    with pytest.raises(Exception):
```
cpcloud: This should probably raise a `TypeError` or something more specific than `Exception`. Can you check that here?

llllllllll: It would probably be an `AttributeError` looking up `_div` or something. It would be cool to say that this symbol does not support the '/' operator with a TypeError but as is, this is not the case. I can specialize this to AttributeError though.

cpcloud:
this is looking good, how's the pandas backend coming?

i think for sql it makes sense to implement `Concat` and possibly `Repeat`, but `Interp` may be kind of hairy as different RDBMSs implement this differently. I'd be ok with just implementing the first two.

blaze/expr/strings.py
llllllllll: @cpcloud The discovery is still blocking me on the pandas backend, I have no issue writing the compute_up functions; however, getting the methods to appear on the symbols is the issue.

cpcloud: Can you put up some failing tests so that I can take a look?

cpcloud: Or, even better: put up the code you expect to work and I'll diagnose the problem.

llllllllll: I will upload what I have sometime soon, I have been blocked on other things today, sorry.

cpcloud: cool, no rush

llllllllll: I uploaded the compute stuff I have for series

blaze/compute/pandas.py
```diff
@@ -165,6 +165,21 @@ def compute_up(expr, data, **kwargs):
...
+def compute_up(expr, data, **kwargs):
+    return expr.lhs.str.repeat(data)
+
+
+@dispatch(Concat, Series)
+def compute_up(expr, data, **kwargs):
+    return expr.lhs.str.cat(data)
+
+
+@dispatch(Interp, Series)
```
cpcloud: i'd probably implement this like this:

```python
return compute(expr, data.values)
```
because pandas doesn't implement `%` symmetrically.

blaze/compute/pandas.py
```diff
@@ -165,6 +165,21 @@ def compute_up(expr, data, **kwargs):
...
 
 
+@dispatch(Repeat, Series)
+def compute_up(expr, data, **kwargs):
+    return expr.lhs.str.repeat(data)
+
+
+@dispatch(Concat, Series)
+def compute_up(expr, data, **kwargs):
+    return expr.lhs.str.cat(data)
```
cpcloud: `expr.lhs` is an expression or a scalar value, you need to implement these operations similar to how you implemented the numpy backend version. Something like this:

```python
@dispatch(Concat, pd.Series)
def compute_up(t, data, **kwargs):
    if isinstance(t.lhs, Expr):
        return data + t.rhs
    else:
        return t.lhs + data
```
better yet, you can probably just not implement this, since `Concat` is a subclass of `Arithmetic` and the `BinOp` implement will suffice

blaze/compute/pandas.py
```diff
@@ -165,6 +165,21 @@ def compute_up(expr, data, **kwargs):
     return compute_up(expr._child.distinct().count(), data, **kwargs)
 
 
+@dispatch(Repeat, Series)
+def compute_up(expr, data, **kwargs):
+    return expr.lhs.str.repeat(data)
```
cpcloud: similar to `Concat` i think this implementation is unnecessary because of subclassing

blaze/compute/pandas.py
cpcloud: implement this `compute_up` similar to how you did it for numpy

blaze/compute/pandas.py
llllllllll: I used `*` as the `operator.mul`; however, that does not do element-wise string repeat.

cpcloud:
```python
In [1]: s = pd.Series(list('abc'))

In [2]: s
Out[2]:
0    a
1    b
2    c
dtype: object

In [3]: s * 2
Out[3]:
0    aa
1    bb
2    cc
dtype: object
```

llllllllll: Okay, I will leave these as is; however, I still do not see the methods on my `InteractiveSymbol`.

```python
In [8]: ds = bz.Data(pd.Series(('a', 'b', 'c')))

In [9]: 'concat' in dir(ds)
Out[9]: False

In [10]: ds.data + ds.data
Out[10]: 
0    aa
1    bb
2    cc
dtype: object
```

llllllllll:
Okay, I updated the pandas stuff based on the comments made; however, I think the root of the issue is that a series of strings is an ndarray of `object` and not some string type.

```python
In [9]: np.array(('a', 'b', 'c')).dtype
Out[9]: dtype('<U1')

In [10]: pd.Series(('a', 'b', 'c')).values.dtype
Out[10]: dtype('O')

In [11]: pd.Series(('a', 'b', 'c'), dtype='<U1').values.dtype
Out[11]: dtype('O')
```

This happens even if I am explicit about the dtype of the series.

blaze/compute/tests/test_numpy_compute.py
```diff
@@ -438,3 +438,24 @@ def test_nunique_recarray():
...
+def test_str_concat():
+    a = np.array(('a', 'b', 'c'))
+    s = symbol('s', discover(a))
+    expr = s.concat('a')
+    assert all(compute(expr, a) == np.char.add(a, 'a'))
+
+
+def test_str_repeat():
+    a = np.array(('a', 'b', 'c'))
+    s = symbol('s', discover(a))
```
cpcloud: when you write these tests for `Series`, you'll need to explicitly create the symbol with a `string` dtype, else the methods won't show up on the symbol.

`discover(series)` when `series` has strings will produce a datashape with `object` measure. We should probably use pandas own type inference here to check whether a `Series` is actually object vs string

cpcloud: i'll submit a patch to odo to do this

llllllllll:
Okay, I think I understand this now. I have added tests for the pandas backend. I don't really know much about sql, I might not be the best to write that compute backend.

cpcloud:
I wonder if we shouldn't just shove numpy into pandas. numpy's char module algos seem to be 10x slower than pandas

```python
In [23]: v
Out[23]:
array(['%d', '%s', '%d', ..., '%s', '%d', '%s'],
      dtype='<U2')

In [24]: s.values
Out[24]: array(['%d', '%s', '%d', ..., '%s', '%d', '%s'], dtype=object)

In [25]: %timeit np.char.mod(v, 1)
1 loops, best of 3: 3.31 s per loop

In [26]: %timeit s % 1
1 loops, best of 3: 328 ms per loop

In [27]: 3310/328
Out[27]: 10.091463414634147
``

**Comment by cpcloud** [comment]
@llllllllll what do you think about pushing numpy string algos to pandas?

**Comment by llllllllll** [comment]
I wouldn't mind investigating; however, I think that the root of the issue with the speed is that pandas treats a series of strings as dtype `object` instead of strings so it must dispatch with `PyNumber_Mod` or something.

**Comment by cpcloud** [comment]
I would intuitively think that that would be *slower* than what numpy is doing. My numbers above suggest otherwise

**Comment by llllllllll** [comment]
Ah, sorry, I misread. I will convert the numpy version to use pandas.

**Comment by llllllllll** [comment]
The cost of the numpy function grows faster than the pandas version, however, it is better for smaller arrays. Some basic testing has shown that they cross at around n=135. To factor in the cost of creating the series from the array, I think that a nice win could be to have the numpy compute function delegate to pandas for arrays of n>=145. I will work on adding that now.

File: blaze/compute/pandas.py [hunk 11]
```diff
@@ -76,6 +76,21 @@ def compute_up(t, s, **kwargs):
     return compute_up(t, s.to_frame(), **kwargs)
 
 
+@dispatch(Interp, Series)
+def compute_up(t, data, **kwargs):
+    if isinstance(t.lhs, Expr):
+        return pd.Series(data.values % t.rhs)
```
cpcloud: I think this should be

```python
if ...:
    return data % t.rhs
else:
    return t.lhs % data
````

File: blaze/compute/pandas.py [hunk 12]
```diff
@@ -76,6 +76,21 @@ def compute_up(t, s, **kwargs):
...
+    else:
+        return pd.Series(t.lhs % data.values)
+
+
+@compute_up.register(Interp, Series, (Series, base))
+@compute_up.register(Interp, base, Series)
+def compute_up_pd_interp(t, lhs, rhs, **kwargs):
+    return pd.Series(lhs.values % rhs.values)
```
cpcloud: you *should* be able to do just this:

```python
return lhs % rhs
```

cpcloud: If this isn't raising then I suspect you don't have coverage for all the dispatch cases listed here.

blaze/expr/tests/test_arithmetic.py
cpcloud: can you add a test against scalars for add and interp?

something that tests expressions like this:

```python
cs + 'a'
cs % 'a'
```

cpcloud:
@llllllllll one last comment, then i think we're good to merge

llllllllll:
updated

cpcloud:
merging

cpcloud:
@llllllllll thanks!

llllllllll:
yay

PR #1092, opened 2015-05-13
Skip spontaneously failing dask computation

cpcloud:


cpcloud:
Ok, could've just been lucky and passed, but going to merge and if it pops up again just disable dask testing altogether

PR #1097, opened 2015-05-20
Fix issue when converting datetimes and pandas Timestamps in an expression

cpcloud:
cc @mrocklin

### DIFF
--- a/blaze/compute/pyfunc.py
+++ b/blaze/compute/pyfunc.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
+import pandas as pd
 from ..expr import (Expr, Symbol, Field, Arithmetic, Math,
                     Date, Time, DateTime, Millisecond, Microsecond, broadcast,
                     sin, cos, Map, UTCFromTimestamp, DateTimeTruncate, symbol,
@@ -63,7 +64,7 @@ def _print_python(expr, leaves=None):
 
 @dispatch((datetime.datetime, datetime.date))
 def _print_python(expr, leaves=None):
-    return repr(expr), {'datetime': datetime}
+    return repr(expr), {'datetime': datetime, 'Timestamp': pd.Timestamp}
 
 @dispatch(Symbol)
 def _print_python(expr, leaves=None):
--- a/blaze/expr/arithmetic.py
+++ b/blaze/expr/arithmetic.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, division, print_function
 import operator
 from toolz import first
 import numpy as np
+import pandas as pd
 from datashape import dshape, var, DataShape
 from dateutil.parser import parse as dt_parse
 from datashape.predicates import isscalar, isboolean, isnumeric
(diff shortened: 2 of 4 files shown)


cpcloud:
@mrocklin okay with this?

blaze/compute/pyfunc.py
```diff
@@ -63,7 +64,7 @@ def _print_python(expr, leaves=None):
 
 @dispatch((datetime.datetime, datetime.date))
 def _print_python(expr, leaves=None):
-    return repr(expr), {'datetime': datetime}
+    return repr(expr), {'datetime': datetime, 'Timestamp': pd.Timestamp}
```
mrocklin: Are there cases where we intended to print python datetime objects that we actually end up printing Pandas Timestamps?  If so is there an issue?  Do we ever not assume that pandas is around?

blaze/expr/arithmetic.py
```diff
@@ -220,7 +221,7 @@ def scalar_coerce(_, val):
 
 @dispatch(ct.DateTime, _strtypes)
 def scalar_coerce(_, val):
-    return dt_parse(val)
+    return pd.Timestamp(val)
```
mrocklin: Does this work nicely with sqlalchemy?

cpcloud: In what sense?

```python
In [26]: import sqlalchemy as sa

In [27]: c = sa.Column('col', sa.types.TIMESTAMP)

In [28]: expr = c >= pd.Timestamp('now').to_pydatetime()

In [29]: print(expr.compile().render_literal_bindparam(expr.right))
```
raises `NotImplementedError`, and so does the version with the `Timestamp` instance.

mrocklin: Hrm, that's unfortunate.  I'm surprised that sqlalchemy doesn't manage python datetimes well.

cpcloud: It handles them just fine, except when rendering literal value. You can use `literal_column(dt.strftime('...'))` or it's up to the dialect to manage literal rendering. You can of course pass `datetime` around in expressions and it deals with sending those values to the database quite nicely.

blaze/compute/pyfunc.py
cpcloud: > Are there cases where we intended to print python datetime objects that we actually end up printing Pandas Timestamps?

Not sure. FWIW, `Timestamp` is a subclass of `datetime.datetime`.

> Do we ever not assume that pandas is around?

No, pandas is a hard dependency.

blaze/expr/arithmetic.py
cpcloud: As an example, I render literal datetime date and timedelta values in `kdbpy` in the way that kdbpy expects them

cpcloud:
I'll add that the thing that's nice about pandas `Timestamp`s is that they can handle the precision of other systems' timestamps, such as kdb which also has nanosecond precision timestamp types.

cpcloud:
merging

PR #1102, opened 2015-05-23
ENH: Allows the server/client to use pickle

llllllllll:
When trying to use the blaze server, you are heavily restricted in what kind of data you can work with because of the json protocol. The main thing that I needed was a clean way to send datetimes across the wire. To generalize the solution, I have just allowed (not forced) the client/server to use pickle. I realize most people would probably be afraid to unpickle data from the internet; however, blaze server is basically remote code execution by definition. The server will only respond to 'compute.pickle' if the `allow_pickle` flag is set, this is by default false.

### DIFF
--- a/blaze/server/client.py
+++ b/blaze/server/client.py
@@ -3,6 +3,10 @@ from __future__ import absolute_import, division, print_function
 import flask
 from flask import json
 import requests
+try:
+    import cPickle as pickle
+except ImportError:
+    import pickle
 
 from odo import resource
 from datashape import dshape
@@ -65,13 +69,17 @@ class Client(object):
 
     blaze.server.server.Server
     """
-    __slots__ = 'url'
+    __slots__ = 'url', 'mode', 'serial'
 
-    def __init__(self, url, **kwargs):
+    def __init__(self, url, mode='json', **kwargs):
         url = url.strip('/')
         if not url[:4] == 'http':
             url = 'http://' + url
         self.url = url
+        if mode not in ('json', 'pickle'):
+            raise ValueError('mode must be either pickle or json')
+        self.mode = mode
+        self.serial = json if mode == json else pickle
 
     @property
     def dshape(self):
@@ -100,16 +108,18 @@ def compute_down(expr, ec, **kwargs):
     from .server import to_tree
     tree = to_tree(expr)
 
-    r = requests.get('%s/compute.json' % ec.url,
-                     data=json.dumps({'expr': tree}),
-                     headers={'Content-Type': 'application/json'})
+    serial = ec.serial
+    r = requests.get('%s/compute.%s' % (ec.url, ec.mode),
+                     data=serial.dumps({'expr': tree}))
 
     if not ok(r):
         raise ValueError("Bad response: %s" % reason(r))
 
-    data = json.loads(content(r).decode('utf-8'))
-
-    return data['data']
+    if serial is json:
+        raw = content(r).decode('utf-8')
+    else:
+        raw = content(r)
+    return serial.loads(raw)['data']
 
 
 @resource.register('blaze://.+')
--- a/blaze/server/client.py
+++ b/blaze/server/client.py
@@ -1,12 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import flask
-from flask import json
 import requests
-try:
-    import cPickle as pickle
-except ImportError:
-    import pickle
 
 from odo import resource
 from datashape import dshape
@@ -14,6 +9,7 @@ from datashape import dshape
 from ..expr import Expr
 from ..dispatch import dispatch
 from .server import DEFAULT_PORT
+from .serialization import json
 
 # These are a hack for testing
 # It's convenient to use requests for live production but use
@@ -56,6 +52,11 @@ class Client(object):
     url : str
         URL of a Blaze server
 
+    serial : SerializationFormat, optional
+        The serialization format object to use. Defaults to JSON.
+        A serialization format is an object that supports:
+        name, loads, and dumps.
+
     Examples
     --------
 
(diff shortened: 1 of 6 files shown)


blaze/server/server.py
```diff
@@ -1,6 +1,10 @@
 from __future__ import absolute_import, division, print_function
 
 import socket
+try:
+    import cPickle as pickle
+except ImportError:
+    import pickle
```
mrocklin: Should we add `pickle` to compatibility?

blaze/server/server.py
```diff
@@ -310,5 +319,12 @@ def compserver():
...
 
-    return json.dumps({'datashape': str(expr.dshape),
-                       'data': result}, default=json_dumps)
+    return serial.dumps(
+        {'datashape': str(expr.dshape), 'data': result},
+        **({'default': json_dumps} if serial is json else {})
+    )
+
+
+@pickle_extension_api.route('/compute.pickle', methods=['POST', 'PUT', 'GET'])
```
mrocklin: I like the idea of specifying the serialization format by the route.

blaze/server/client.py
```diff
@@ -65,13 +69,17 @@ class Client(object):
...
+    __slots__ = 'url', 'mode', 'serial'
 
-    def __init__(self, url, **kwargs):
+    def __init__(self, url, mode='json', **kwargs):
         url = url.strip('/')
         if not url[:4] == 'http':
             url = 'http://' + url
         self.url = url
+        if mode not in ('json', 'pickle'):
+            raise ValueError('mode must be either pickle or json')
```
mrocklin: I wonder if this could be cheaply generalized to any serialization format by accepting dumps/loads functions in the Client/Server.  This would let us extend out to other formats like msgpack.

blaze/server/client.py
```diff
@@ -100,16 +108,18 @@ def compute_down(expr, ec, **kwargs):
     from .server import to_tree
     tree = to_tree(expr)
 
-    r = requests.get('%s/compute.json' % ec.url,
-                     data=json.dumps({'expr': tree}),
-                     headers={'Content-Type': 'application/json'})
+    serial = ec.serial
+    r = requests.get('%s/compute.%s' % (ec.url, ec.mode),
+                     data=serial.dumps({'expr': tree}))
```
mrocklin: Note that for pickle, if the client and server are close and if you have a decent amount of numerical data, and if you're in Python2, then you *really* want to use the following:

    dumps = partial(pickle.dumps, protocol=pickle.HIGHEST_PROTOCOL)

blaze/server/server.py
llllllllll: It is stdlib

mrocklin: I know, I've just found myself doing this exact same try-except block.  Just a thought that we could do this in `compatibility.py` and then import pickle from there.  Fine either way though.

llllllllll: sorry, I misunderstood and thought you meant some compatibility requirements file.

blaze/server/client.py
llllllllll: I just pushed a commit that generalizes this into a format where the server accepts an iterable of formats, and the client specifies which format it will use. Maybe work can be done to add a series of formats to fall back to.

blaze/server/client.py
llllllllll: Thanks, does python2 support the python3 pickle format, I haven't tried, and to remove the issue with qdb I just used json.

llllllllll:
travis just failed 2.6 because of a dict comprehension, I will replace with a map or something.

cpcloud:
you can construct a generator expression inside of a `dict` to simulate `dict` comprehensions:

```python
dict((k, v) for k, v in list_of_pairs)
```

DavidMertz:
The comprehension in the initializer seems superfluous:

```python
Python 2.6.9 |Anaconda 2.2.0 (x86_64)| (unknown, Jan 10 2014, 13:33:57)
[GCC 4.2.1 (Apple Inc. build 5577)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> list_of_pairs = [('foo',1),('bar',2)]
>>> dict(list_of_pairs)
{'foo': 1, 'bar': 2}
```

llllllllll:
I think that was just an example to point me in the correct direction, the actual code being discussed was:

```python
_get_format.cache[app] = dict(
    (f.name, f) for f in _get_option('formats', options)
)
```

cpcloud:
@llllllllll can you add a test for this?

llllllllll:
I now run the server tests with all the serialization formats

blaze/server/server.py
```diff
@@ -70,6 +85,12 @@ class Server(object):
         A dictionary mapping dataset name to any data format that blaze
         understands.
 
+    formats : ``iterable[SerializationFormat]``, optional
+        An iterable of supported serialization formats. By default, the
+        server will support JSON.
+        A serialization format is an object that supports:
+        name, loads, and dumps.
```
mrocklin: If I were doing the work I would probably skip making the new type and use a dict

```python
formats = {'json': (json.dumps, json.loads),
           'pickle': (partial(pickle.dumps, protocol=pickle.HIGHEST_PROTOCOL), pickle.loads),
           'msgpack': (msgpack.packb, msgpack.unpackb)
           }
```

But I'm quite grateful at not having to do this work and the chosen strategy currently in this PR seems sane to me; so I say go for it, particularly since this PR has been going on for a while.

cpcloud: @llllllllll if we decide to go with the `dict` solution that @mrocklin showed above at some point in the future, would that break anything you're working on?

llllllllll: No, I am not using this yet. One benefit of the namedtuple approach is that I can pass the same serialization format object to the server and the client and it will correctly generate and query the proper compute.{name} endpoint. If I passed a dict, I would need to manually keep the name in sync.

cpcloud: Cool. I'm happy as is.

cpcloud:
Merging

cpcloud:
@llllllllll Thanks!

PR #1103, opened 2015-05-26
rename binstar.org to anaconda.org

mrocklin:


mrocklin:
This should not be merged and built until tomorrow

PR #1115, opened 2015-06-04
Add directory spidering

### LINKED ISSUE #1108: Simple deploy blaze-server with loaded data
From the roadmap meeting discussion: write a simple script to deploy blaze-server with loaded data from a directory.

Goal: Given a folder directory with data (any supported blaze backend) deploy a blaze-server with the data loaded in it.

Method:
For each file in the directory, run discover(uri) or Data(uri). If data can be discovered --> add data to server. e.g. Server(uri)

cpcloud:
closes #1108

- [x] test the executable on windows py2
- [x] test on py3 on mac and windows
- [x] add yaml file format to the command line as per an internal discussion with @chdoig


blaze/utils.py
```diff
@@ -162,3 +163,25 @@ def json_dumps(dt):
...
+            if new_resources:
+                resources[basename] = new_resources
+        else:
+            with ignoring(*ignore):
+                resources[basename] = resource(filename)
+    return resources
+
+
+def spider(resource_path, ignore=(ValueError, NotImplementedError)):
+    return {
```
llllllllll: You might want to wrap this in:

```python
try:
    ....
except RuntimeError:
    raise ValueError('{dir} is too deeply nested'.format(dir=resource_path))
```

So that users are not confused by a recursion limit error.

cpcloud: an `os.walk` based solution would probably eliminate the need for this restriction entirely

mrocklin: What is the limitation here?  Is this about the case of a very very nested directory?  How likely is this?

llllllllll: It is unlikely unless there is a symlink back above the parent. also, I agree with @cpcloud about the use of walk.

cpcloud: i'll add a test case for a directory cycle

cpcloud: seems unlikely to have an actual directory structure that is >= 1000 levels deep

cpcloud: FWIW, using `os.walk` significantly complicates the solution

cpcloud: i'm probably going to leave the implementation as is (modulo the runtimeerror catching) until someone comes to us with a 1000 level deep directory

llllllllll: Okay, maybe we should add a flag for `follow_symlinks` or something? Also, it doesn't hurt to provide a nicer error than RuntimeError: Maximum recursion depth exceeded. We could even suggest altering the `follow_symlinks` flag.

cpcloud: It looks like the number of symlinks is limited by the OS. In OS X for example:

```sh
$ ag -C 5 'MAXSYMLINKS' /usr/include/
/usr/include/sys/param.h
186-
187-/*
188- * MAXPATHLEN defines the longest permissable path length after expanding
189- * symbolic links. It is used to allocate a temporary buffer from the buffer
190- * pool in which to do the name expansion, hence should be a power of two,
191: * and must be less than or equal to MAXBSIZE.  MAXSYMLINKS defines the
192- * maximum number of symbolic links that may be expanded in a path name.
193- * It should be set high enough to allow all legitimate uses, but halt
194- * infinite loops reasonably quickly.
195- */
196-#define     MAXPATHLEN      PATH_MAX
197:#define MAXSYMLINKS 32
```

cpcloud: linux is a similarly a small number.

@llllllllll @mrocklin what do you think about symlinks? always follow? not follow, but have an option? i'm thinking option is the way to go

llllllllll: I would prefer an option that defaults to True

mrocklin:
> what do you think about symlinks?

My somewhat lazy tactic would probably be just not to care until someone ran into and reported the problem.

chdoig:
Hey @cpcloud sorry for the delay reviewing this PR. 

Could we have the blaze-server runnable script accessible from the blaze top directory? There are different ways to do that in my experience, I'll be happy to write the option that you prefer, didn't want to choose one without your opinion:

Options:
- Add an entry point in your conda meta.yaml file, like what I did in topik: https://github.com/ContinuumIO/topik/blob/4f50eace339fb20db537bcc241421f793a7f6a6a/conda.recipe/meta.yaml#L12-L15

e.g.
```yaml
  entry_points:
    - blaze-server = blaze.server.server:main
```

- And entry point in your setup.py: https://github.com/ContinuumIO/topik/blob/4f50eace339fb20db537bcc241421f793a7f6a6a/setup.py#L17-L19

e.g.

```python
      entry_points= {
          'console_scripts': ['blaze-server = blaze.server.server:main']
      },
```

- In Anaconda Cluster @quasiben creates a bin directory to start the cli https://github.com/ContinuumIO/anaconda-cluster/blob/f35672ee32c09bef7aaf10ec2784f15198d4c17e/bin/acluster and then adds it as a script in the setup.py: https://github.com/ContinuumIO/anaconda-cluster/blob/f35672ee32c09bef7aaf10ec2784f15198d4c17e/setup.py#L96

e.g.

```
#!/usr/bin/env python

import sys
from blaze.server.server import main
sys.exit(main())
```

cpcloud:
i'm +1 on the `entry_points` (in conda and `setup.py`) method

cpcloud:
@chdoig mostly i'm concerned that might've gone a little hog wild with options in the `spider` function

cpcloud:
@chdoig I've add an entry point in conda and in `setup.py`

chdoig:
Cool, what do you mean by a little hog wild? Like in the arguments options? Any specific feedback you need from me?

This might not going off a tangent a little, but just to through it here: it might also be nice to have a file with Blaze URIs, name and some metadata (maybe a yaml file) or a master table / catalog:

data.file
```
name: accounts-data
uri: 'blaze/examples/data/accounts_*.csv'
----
name: iris-dataset
uri: 'sqlite:///blaze/examples/data/iris.db::iris'
```

and that you could do:

```
blaze-server --from_file data.file
```

cpcloud:
what about 

```
accounts_data: 'blaze/examples/data/accounts_*.csv'
iris_dataset: 'sqlite:///blaze/examples/data/iris.db::iris'
```
?

cpcloud:
by hog wild i was really referring to my `glob` argument which i think i'm going to remove ... it's kind of fragile and hard to reason about

chdoig:
Works, I was just thinking that you might wanted to keep some description, the datashape...

cpcloud:
hm okay what about this:

```yaml
accounts_data:
  uri: path/to/file
iris_dataset:
  uri: sqlite:///path/to/file.db
  dshape: "var * {...}"
```
We should take advantage of the fact that yaml already has a way to name things without having to explicitly say "here is the name"

chdoig:
Yes, that was silly. This looks better.

blaze/server/spider.py
```diff
@@ -0,0 +1,96 @@
...
+            resources = spider(os.path.expanduser(uri),
+                               ignore=ignore,
+                               followlinks=followlinks,
+                               hidden=hidden)
+        else:
+            resources[name] = resource(uri, dshape=info.get('dshape'))
+    return resources
+
+
+def _parse_args():
```
llllllllll: Should this accept a `host` argument to pass to the server? Also, on the line below, is there a reason for the runtime import? It makes it a lot harder to chroot before starting up the server.

cpcloud: > Should this accept a `host` argument to pass to the server?

It does, via the `--host` argument on the command line

> Also, on the line below, is there a reason for the call-time import?

`import odo` is somewhat slow because of the many `dispatch` decorated functions. I was trying to avoid importing odo until needed so that `--help` would be relatively quick, but since it causes other more insidious problems I'll move them to the top level of the module.

cpcloud: Sorry just realized you said `host` and not `port`.

cpcloud: I can add a host parameter

cpcloud:
merging this soon

cpcloud:
just tested a simple version of this on windows and it worked. merging on next pass

cpcloud:
hm, nevermind there's a strange python34 unicode error in json discovery

cpcloud:
turns out it was an untested aspect of odo

cpcloud:
odo PR on the way

cpcloud:
fixed in https://github.com/ContinuumIO/odo/pull/243

cpcloud:
merging on pass for real this time :)

PR #1120, opened 2015-06-08
sql date

llllllllll:
I got errors trying to use the `date` attribute on datetimes with a sql backend. The sql for conversion from datetime to date is not an extract, but a function call. Please let me know if there is a cleaner implementation of this.


cpcloud:
LGTM. Thanks for the fix.

llllllllll:
Thanks for the quick review

PR #1121, opened 2015-06-10
csv merge

llllllllll:
closes: https://github.com/ContinuumIO/blaze/issues/1119

This pr also allows you to change the suffixes of the left and right columns in a `Join` similar to `pandas.merge`

### DIFF
--- a/blaze/compute/core.py
+++ b/blaze/compute/core.py
@@ -446,10 +446,9 @@ def compute(expr, d, **kwargs):
     optimize_ = kwargs.get('optimize', optimize)
     pre_compute_ = kwargs.get('pre_compute', pre_compute)
     post_compute_ = kwargs.get('post_compute', post_compute)
-
     expr2, d2 = swap_resources_into_scope(expr, d)
     if pre_compute_:
-        d3 = dict([(e, pre_compute_(expr2, dat, **kwargs))
+        d3 = dict([(e, pre_compute_(e, dat, **kwargs))
                         for e, dat in d2.items()
                         if e in expr2])
     else:
--- a/blaze/compute/core.py
+++ b/blaze/compute/core.py
@@ -448,9 +448,11 @@ def compute(expr, d, **kwargs):
     post_compute_ = kwargs.get('post_compute', post_compute)
     expr2, d2 = swap_resources_into_scope(expr, d)
     if pre_compute_:
-        d3 = dict([(e, pre_compute_(e, dat, **kwargs))
-                        for e, dat in d2.items()
-                        if e in expr2])
+        d3 = dict(
+            (e, pre_compute_(e, dat, **kwargs))
+            for e, dat in d2.items()
+            if e in expr2
+        )
     else:
         d3 = d2
 
--- a/blaze/compute/core.py
+++ b/blaze/compute/core.py
@@ -167,7 +167,7 @@ def top_then_bottom_then_top_again_etc(expr, scope, **kwargs):
     optimize_ = kwargs.get('optimize', optimize)
     pre_compute_ = kwargs.get('pre_compute', pre_compute)
     if pre_compute_:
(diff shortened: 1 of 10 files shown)


llllllllll:
All tests passing locally, I will rerun

cpcloud:
hm what was the issue?

llllllllll:
I am not sure, I think travis may have just had a hiccup because the results were not loading.

cpcloud:
can you add a test for merging two csv files?

blaze/compute/core.py
```diff
@@ -446,10 +446,9 @@ def compute(expr, d, **kwargs):
     optimize_ = kwargs.get('optimize', optimize)
     pre_compute_ = kwargs.get('pre_compute', pre_compute)
     post_compute_ = kwargs.get('post_compute', post_compute)
-
     expr2, d2 = swap_resources_into_scope(expr, d)
     if pre_compute_:
-        d3 = dict([(e, pre_compute_(expr2, dat, **kwargs))
+        d3 = dict([(e, pre_compute_(e, dat, **kwargs))
```
llllllllll: This is what was causing the wrong shape to be applied to the rest of the data types, can someone make sure that this is the correct fix

cpcloud: looking now

cpcloud: it looks like your fix is correct. before your fix the code is mapping each subexpression to the same `pre_compute` result. after your fix each subexpression in `expr2` gets mapped to the correct `pre_compute`d result.

cpcloud: i suspect this was never caught because in many cases `e == expr2`

cpcloud: any chance i could get you to kill those brackets around the comprehension?

llllllllll: sure thing

llllllllll:
merge on success?

cpcloud:
hm this is also happening here: https://github.com/quantopian/blaze/blob/csv-merge/blaze/compute/core.py#L170-L172, let me poke around for a bit

llllllllll:
Other than the place I just changed, is there ever a case that passes in `e` instead of the top level expr?

cpcloud:
not sure what you mean. passes in `e` to `pre_compute_`?

llllllllll:
sorry, is there a place that passes the expr out of the scope vs the top level expr itself, basically a place that does what I changed it to do.

blaze/compute/core.py
```diff
@@ -446,12 +446,13 @@ def compute(expr, d, **kwargs):
     optimize_ = kwargs.get('optimize', optimize)
     pre_compute_ = kwargs.get('pre_compute', pre_compute)
     post_compute_ = kwargs.get('post_compute', post_compute)
-
     expr2, d2 = swap_resources_into_scope(expr, d)
     if pre_compute_:
-        d3 = dict([(e, pre_compute_(expr2, dat, **kwargs))
```
cpcloud: @mrocklin Are these the correct arguments to `pre_compute`? For expressions with multiple leaves like `Join` this does the wrong thing because it calls `into` with the dshape of the first leaf that it finds. According to the pipeline docs, `pre_compute` operates on the leaves of the expression, which is more inline with @llllllllll change here to `pre_compute` each expression in scope.

mrocklin: It looks like most cases of `pre_compute` don't care about the expression being passed in so changing the input expr to be the leaf would probably be safe.  At this point we would probably want to just kill the expression as an input though.

cpcloud: i'd like to leave the expression in the input for now and consider removing in release 0.8.2. @llllllllll go ahead and change the other `pre_compute_` call to be similar to this one

cpcloud: then merge on passing

PR #1123, opened 2015-06-10
BUG: Fix isnan with postgres

llllllllll:
sqlite does not have a `NaN` value so I don't test against it.

### DIFF
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -50,6 +50,7 @@ from ..expr import nunique, Distinct, By, Sort, Head, Label, ReLabel, Merge
 from ..expr import common_subexpression, Summary, Like, nelements
 
 from ..expr.broadcast import broadcast_collect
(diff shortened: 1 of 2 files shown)


cpcloud:
wacky.

from the postgresql docs: http://www.postgresql.org/docs/current/static/datatype-numeric.html

> In addition to ordinary numeric values, the numeric type allows the special value NaN, meaning "not-a-number". Any operation on NaN yields another NaN. When writing this value as a constant in an SQL command, you must put quotes around it, for example UPDATE table SET x = 'NaN'. On input, the string NaN is recognized in a case-insensitive manner.

cpcloud:
So apparently there *is* something bigger than infinity:

```sql
pcloud=# select float8 'NaN' > float8 'Infinity';
 ?column?
----------
 t
(1 row)
```

cpcloud:
@llllllllll thanks!

PR #1125, opened 2015-06-11
Add sql slicing and make sort use a CTE so that the semantics are similar to numpy

### LINKED ISSUE #1072: Slicing on Interactive Data with SQL does not work
Hi all,

Big fan of the project. I've started working on a 10 minutes to blaze tutorial (#423) and noticed that Slicing doesn't work for interactive objects. See [here](http://nbviewer.ipython.org/github/Will-So/Blaze-Workbook/blob/Slicing.ipynb) for an example.

I'd be happy to work on this but I still don't have a good grasp how exactly the `Data` function works. Any tips would be appreciated.

cpcloud:
closes #1072

@Will-So

This PR has all of your commits as well as a few extras to fix the order of
operations for sort on the SQL backend.


cpcloud:
cc @llllllllll this fixes the stuff we talked about on gitter regarding sort

blaze/compute/tests/test_sql_compute.py
```diff
@@ -1577,6 +1614,13 @@ def test_timedelta_arith(sql_with_dts):
     ).all()
 
 
+def test_sort_compose():
+    expr = t.name[:5].sort()
+    result = compute(expr, s)
+    expected = """with anon_1 as (select accounts.name as name from accounts limit :param_1 offset :param_2) select anon_1.name from anon_1 order by anon_1.name asc"""
```
cpcloud: note to self: format this

blaze/compute/sql.py
```diff
@@ -793,3 +806,40 @@ def post_compute(_, s, **kwargs):
...
+        if start < 0:
+            raise ValueError('start value of slice cannot be negative'
+                             ' with a SQL backend')
+
+        stop = index.stop
+        if stop is not None and stop < 0:
+            raise ValueError('stop value of slice cannot be negative with a '
+                             'SQL backend.')
+
+        if index.step is not None and index.step != 1:
```
llllllllll: This could be supported:

```sql
SELECT * FROM ( SELECT *, row_number() over () AS number FROM datatable ) foo 
    WHERE foo.number % step = 0;
```

cpcloud: windows functions are definitely at the forefront of my mind. i wonder if this could be used to implement slicing in general ... we can always change the implementation of slicing for sql.

cpcloud: going to merge this pr (after fixing my note to self)

PR #1126, opened 2015-06-11
subselect instead of cte

llllllllll:
workaround for: https://bitbucket.org/zzzeek/sqlalchemy/issue/2551/apparently-inserts-and-update-delete-can

this also gives us mysql support


cpcloud:
can you add a test that uses `odo(blaze_expr_involving_subselect, thing_to_insert_to)`

Issue #974, opened 2015-03-04
Union Operation


mrocklin:
Sometimes we want to union several tables together to form one large table.  Some things to think about

1.  Expressions with an indeterminate number of inputs are always a bit tricky.  This came up in `Broadcast`
2.  For some backends, like SQL, this is straightforward (there is a UNION sqlalchemy term), but for others, like a bunch of different CSV files or a mix of various backends things do become a bit tricky.  This might be a good opportunity for parallelism but we'll have to manage the `compute_up`/`compute_down` dance well.

notestaff:
"Union" might not be the right name for the operation, as row order matters unlike in SQL union.
VStack seems more accurate.

llllllllll:
#1128 adds this operation

PR #1128, opened 2015-06-11
concat

llllllllll:
Defines a vstack operation that can be performed on 2 tables. This pull request also implements the backend for pandas, sql, csv, and numpy

### DIFF
--- a/blaze/compute/pandas.py
+++ b/blaze/compute/pandas.py
@@ -35,7 +35,7 @@ from ..expr import (Projection, Field, Sort, Head, Broadcast, Selection,
                     Map, Apply, Merge, std, var, Like, Slice, summary,
                     ElemWise, DateTime, Millisecond, Expr, Symbol, IsIn,
                     UTCFromTimestamp, nelements, DateTimeTruncate, count,
-                    UnaryStringFunction, nunique)
+                    UnaryStringFunction, nunique, VStack)
 from ..expr import UnaryOp, BinOp, Interp
 from ..expr import symbol, common_subexpression
 from .core import compute, compute_up, base
@@ -143,6 +143,11 @@ def compute_up(t, lhs, rhs, **kwargs):
         suffixes=t.suffixes,
     )
     return result.reset_index()[t.fields]
+
+
+@dispatch(VStack, DataFrame, DataFrame)
+def compute_up(t, lhs, rhs, _vstack=np.vstack, **kwargs):
+    return DataFrame(_vstack((lhs.values, rhs.values)), columns=lhs.columns)
 
 
 @dispatch(Symbol, (DataFrameGroupBy, SeriesGroupBy))
--- a/blaze/compute/pandas.py
+++ b/blaze/compute/pandas.py
@@ -153,7 +153,7 @@ def compute_up(t, lhs, rhs, _concat=pd.concat, **kwargs):
     if not (isinstance(lhs, type(rhs)) or isinstance(rhs, type(lhs))):
         raise TypeError('lhs and rhs must be the same type')
 
-    return _concat((lhs, rhs), axis=t.axis)
+    return _concat((lhs, rhs), axis=t.axis, ignore_index=True)
 
 
 def get_scalar(result):
--- a/blaze/compute/pandas.py
+++ b/blaze/compute/pandas.py
@@ -35,7 +35,7 @@ from ..expr import (Projection, Field, Sort, Head, Broadcast, Selection,
                     Map, Apply, Merge, std, var, Like, Slice, summary,
                     ElemWise, DateTime, Millisecond, Expr, Symbol, IsIn,
                     UTCFromTimestamp, nelements, DateTimeTruncate, count,
-                    UnaryStringFunction, nunique, VStack)
+                    UnaryStringFunction, nunique, Concat)
 from ..expr import UnaryOp, BinOp, Interp
 from ..expr import symbol, common_subexpression
 from .core import compute, compute_up, base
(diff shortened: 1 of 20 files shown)


jreback:
I think you should just call this ``concat`` or ``concatenate``. I have always found the ``*stack`` methods pretty confusing, my 2c.

llllllllll:
I don't have a preference; however, I took the name from @cpcloud in #974

blaze/expr/collections.py
```diff
@@ -432,6 +432,60 @@ def join(lhs, rhs, on_left=None, on_right=None, how='inner'):
...
+    See Also
+    --------
+
+    blaze.expr.collections.Merge
+    """
+    __slots__ = '_hash', 'lhs', 'rhs'
+    __inputs__ = 'lhs', 'rhs'
+
+    @property
+    def dshape(self):
```
llllllllll: is there some magic in `datashape` to do this already?

mrocklin:
Thoughts on the `stack` operation as defined in dask.array (and now I think ported over to numpy).  This is, I think, significantly cleaner than the `*stack` functions.

http://dask.readthedocs.org/en/latest/stack.html

llllllllll:
I think that concatenate is a good name for this operation. How should I name the node to not clash with the `Concat` class for string concats?

cpcloud:
i like `stack` because it's flexible enough to cover both the table case as well as the array case. With tables the only valid value for the `axis` argument is 0 whereas with arrays you can stack along any dimension.

mrocklin:
unioning two tables sounds more like `np.concatenate` than stack.  To me the difference is that stack creates a new dimension while concatenate adds along an existing dimension.

cpcloud:
fair enough, `concat` is probably the way to go. @llllllllll i think as long as they're namespaced properly it should be fine to name this operation `Concat` as well.

llllllllll:
Do you not think that would make the compute engine code more complicated because I can string concat two tables or "vstack" concat two tables? As a reader, how would I know what is meant. Also, the fact that `bz.concat(a, b)` and `a.concat(b)` are totally different would be too confusing. I personally think that we should just change string concat to an implementation of `Add` on `string` shapes to avoid the confusion, I am not sure that the new `Concat` node was even in the new release so it can't have too much adoption yet

llllllllll:
Only test_sparksql is failing.

llllllllll:
rebased with spark fix.

blaze/compute/pandas.py
```diff
@@ -145,9 +145,15 @@ def compute_up(t, lhs, rhs, **kwargs):
...
-    return gb
+pandas_structure = DataFrame, Series, DataFrameGroupBy, SeriesGroupBy
+
+
+@dispatch(Concat, pandas_structure, pandas_structure)
+def compute_up(t, lhs, rhs, _concat=pd.concat, **kwargs):
+    if not (isinstance(lhs, type(rhs)) or isinstance(rhs, type(lhs))):
+        raise TypeError('lhs and rhs must be the same type')
+
+    return _concat((lhs, rhs), axis=t.axis)
```
llllllllll: I cannot decide if I should `reset_index(drop=True)` on this so that if you are using an integer index, it is still unique after the concat.

cpcloud: if you decided to do that then pass `ignore_index=True` to `pd.concat`

llllllllll: Ah, I didn't know about this flag.

llllllllll:
Can someone take a took at this?

blaze/expr/collections.py
```diff
@@ -437,6 +440,102 @@ def join(lhs, rhs, on_left=None, on_right=None,
...
+
+
+def concat(lhs, rhs, axis=0):
+    if axis < 0:
+        raise ValueError('axis must be greater than or equal to 0')
+
+    ldshape = lhs.dshape
+    rdshape = rhs.dshape
+    if ldshape.measure != rdshape.measure:
+        raise ValueError(
```
cpcloud: I think a `TypeError` is more appropriate here.

blaze/expr/collections.py
```diff
@@ -437,6 +440,102 @@ def join(lhs, rhs, on_left=None, on_right=None,
...
+                l=ldshape.measure, r=rdshape.measure
+            ),
+        )
+
+    lshape = ldshape.shape
+    rshape = rdshape.shape
+    for n, (a, b) in enumerate(zip_longest(lshape, rshape, fillvalue=None)):
+        if n != axis and a != b:
+            raise ValueError(
+                "Shapes are not equal along axis '{n}': {a} != {b}".format(
```
cpcloud: Any reason for the single quotes here? I found it a bit awkward to see this

```
Shapes are not equal along axis '0': 5 != 3
```

Not a big deal though.

cpcloud:
@llllllllll Can you add a test that exercises the error conditions?

blaze/compute/sql.py
```diff
@@ -111,6 +111,16 @@ def compute_up(t, s, **kwargs):
     return compute(t._scalar_expr, d, post_compute=False).label(name)
 
 
+@dispatch(Concat, (Select, Selectable), (Select, Selectable))
+def compute_up(t, lhs, rhs, **kwargs):
+    if t.axis != 0:
+        raise NotImplementedError(
+            'Cannot concat along a non-zero axis in sql',
```
cpcloud: I wonder if pointing people to `Merge` expressions is useful in this case since if they pass `axis=1`, `Merge` is likely what they want.

cpcloud: I also think a `ValueError` is better here. Whenever I see `NotImplementedError` I have hope that one day the world will improve. In this case, I don't think we'll ever implement `axis=1` for SQL.

llllllllll: I think will point people to merge here, because this seems like it might be a common mistake. I also will switch to a valueerror. I think there needs to be some notion of a `NotImplementedOnPurpose` exception in the stdlib for cases like this.

blaze/expr/collections.py
llllllllll: I will drop them. Do you think this should also be a type error since it is about the shape of the expr?

blaze/compute/sql.py
cpcloud: ha! `raise DreamCrushError()` :)

blaze/expr/collections.py
cpcloud: Yes a `TypeError` is probably the way to go here.

blaze/compute/sql.py
llllllllll: lol

cpcloud:
@llllllllll Anything else you want to add here?

llllllllll:
Nope, all set

cpcloud:
@llllllllll merge when ready

PR #1130, opened 2015-06-12
Properly cast reductions when necessary

### LINKED ISSUE #1050: sum(boolean) doesn't work on sql but it's exposed as an attribute on bool columns
sqlalchemy gives the appropriate db api error, but we should probably implicitly cast to int on reductions on boolean operations on sql database columns

cpcloud:
closes #1050

also restricts the scope of most reductions (notable exceptions are count and nelements)
to only work on sqlalchemy.elements.ColumnElements, rather than the generic ClauseElement.

### DIFF
--- a/blaze/compute/sparksql.py
+++ b/blaze/compute/sparksql.py
@@ -20,10 +20,12 @@ import sqlalchemy as sa
 from ..dispatch import dispatch
 from ..expr import Expr, symbol, Join
 from .core import compute
+from .sql import avg
 from .utils import literalquery, istable, make_sqlalchemy_table
 from ..utils import listpack
 from .spark import jgetattr
 from pyhive.sqlalchemy_hive import HiveDialect
+import pyspark
 from pyspark import SQLContext
 
 __all__ = []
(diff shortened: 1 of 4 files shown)


cpcloud:
cc @llllllllll can you try this out on a few things on your side? this touches some often hit code paths, i want to make sure nothing major is borked.

cpcloud:
ugh it looks like spark 1.2 doesn't allow casting to anything but `DECIMAL`

cpcloud:
@llllllllll i'd like to get this in, any chance you could review this tomorrow and/or try it out to make sure there are no major breakages? i've run it on some simple redshift queries (this was partly the impetus for this change) and things are working

llllllllll:
Sure thing, I will test this tomorrow morning at work

cpcloud:
awesome thx!


--
Best,
Phillip Cloud

llllllllll:
To test, I created a table with a single bool column in postgres; however, I could not cast bools to floats
```
ProgrammingError: (psycopg2.ProgrammingError) function sum(boolean) does not exist
LINE 1: SELECT sum(test.c) AS c_sum 
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
 [SQL: 'SELECT sum(test.c) AS c_sum \nFROM test']
```

llllllllll:
```sql
bz=# select cast(False as int);
 int4 
------
    0
(1 row)

bz=# select cast(False as float);
ERROR:  cannot cast type boolean to double precision
LINE 1: select cast(False as float);
               ^
bz=# select cast(False as int);
 int4 
------
    0
(1 row)

bz=# select cast(cast(False as int) as decimal);
 numeric 
---------
       0
(1 row)

```

It looks like we can actually cast to int an then to numeric though

cpcloud:
i think casting to `int` is probably okay for now

cpcloud:
closing in favor of #1137

PR #1133, opened 2015-06-13
Fix typo: s/NotIimplementedError/NotImplementedError/

maxalbert:
This isn't really a great improvement as it will fail either way, but anyway. :-P


llllllllll:
This would have caused a `NameError` instead of the intended `NotImplementedError`, thanks for the fix!

PR #1134, opened 2015-06-13
WIP: Add cross origin resource sharing by default to blaze server

### LINKED ISSUE #1131: blaze.server does not allow CORS
`blaze.server` sticks with same origin policy so it does not enable CORS requests.. This is an issue for users using bokeh `BlazeDataSources` to map to a blaze server and update data directly on the browser without `bokeh.server` because this `blaze.server` (well, Flask default) configuration will raise an error. Something like:

```
  XMLHttpRequest cannot load http://domain:9000/compute.json. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://domain:5006' is therefore not allowed access.
```
.

This issue can be fixed by setting (injecting) the CORS policies on `blaze.server`. For instance:

```Python
        server = blaze.Server({'dat': gauss})
    
        def add_crossdomain_headers(response):
            response.headers.add('Access-Control-Allow-Origin', '*')
            response.headers.add('Access-Control-Allow-Methods', 'GET,POST')
            response.headers.add('Access-Control-Allow-Headers', "Origin, X-Requested-With, Content-Type, Accept")
    
            return response
    
        server.app.after_request(add_crossdomain_headers)
        server.run(host="127.0.0.1", port=9000)

```

But a probably better solution would be to have some kind of configuration on `blaze.server` itself so users don't have to bother too much trying to find the solution. 

NOTE: Flask does expose a `crossdomain` decorator that can be used to decorate views (and that may be a better solution) but I didn't figure out a quick way of using it within `blaze.server` before the views are declared (I may be overlooking something trivial, I honestly didn't have much time to dig in details).

cpcloud:
cc @fpliger @hhuuggoo @bryevdv

closes #1131

### DIFF
--- a/blaze/server/tests/test_server.py
+++ b/blaze/server/tests/test_server.py
@@ -33,21 +33,27 @@ data = {'accounts': accounts,
           'events': events,
               'db': db}
 
-server = Server(data, all_formats)
 
-test = server.app.test_client()
+@pytest.fixture(scope='module')
+def server():
+    s = Server(data, all_formats)
+    s.app.testing = True
+    return s
 
 
-def test_datasets():
+@pytest.yield_fixture
+def test(server):
+    with server.app.test_client() as c:
+        yield c
+
+
+def test_datasets(test):
     response = test.get('/datashape')
     assert response.data.decode('utf-8') == str(discover(data))
 
 
-@pytest.mark.parametrize(
-    'serial',
-    all_formats,
-)
-def test_bad_responses(serial):
+@pytest.mark.parametrize('serial', all_formats)
+def test_bad_responses(test, serial):
     assert 'OK' not in test.post(
         '/compute/accounts.{name}'.format(name=serial.name),
         data=serial.dumps(500),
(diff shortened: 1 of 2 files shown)


blaze/server/tests/test_server.py
```diff
@@ -369,3 +344,28 @@ def test_serialization_endpoints(serial):
     result = respdata['data']
     assert result == odo(compute(expr, {t: data}), int)
     assert respdata['names'] == ['petal_length_sum']
+
+
+@pytest.mark.parametrize('serial', all_formats)
+def test_cors_compute(test, serial):
```
cpcloud: @fpliger I couldn't get this test to fail. It passes whether I have CORS enabled or not. Is there something I'm not doing properly to get it to fail when I *haven't* enabled CORS?

fpliger: @cpcloud what is the "fail" situation test and the "success" situation test?

cpcloud:
@fpliger is there an example of something that i can test this against that would trigger the failure on the browser side *without* this fix?

hhuuggoo:
you'd have to open up chrome and try to hit the a url from javascript

cpcloud:
alrightly down the rabbit hole!

cpcloud:
alrighty i smoked tested this in javascript and i'm able to send requests from another origin, merging

PR #1137, opened 2015-06-16
Coerce expression

### LINKED ISSUE #1050: sum(boolean) doesn't work on sql but it's exposed as an attribute on bool columns

cpcloud:
closes #1050


blaze/expr/tests/test_expr.py
```diff
@@ -136,3 +136,15 @@ def test_hash_to_different_values():
...
     assert (expr, '_and') in _attr_cache
     assert (expr2, '_and') in _attr_cache
+
+
+def test_coerce():
+    s = symbol('s', var * float32)
+    assert str(s.coerce('int64')) == "s.coerce(to='int64')"
+
+
+@pytest.mark.xfail(raises=AttributeError, reason='Should this be valid?')
```
llllllllll: Is the question here if we should be able to cast down from float64 to float32? I think the idea of casting records like this is pretty cool.

cpcloud: The question is: Is casting an entire set of rows on a subset of columns (possibly all the columns as well) useful? How often is it the case that you want to change the type of a set of columns? I honestly don't know. I know that I do this in one particular case when I'm shoving a DataFrame that contains strings into a bcolz ctable in Python3.4.

llllllllll: This would be nice because there are some places that I have tz-aware datetimes in dataframes so the type is object and I would like to coerce all of the columns to `datetime` instead of `object`.

llllllllll:
Can you add a series coerce?

cpcloud:
yep, just haven't gotten to it yet

blaze/expr/expressions.py
```diff
@@ -705,6 +705,21 @@ def dshape(self):
         return dshape(self._dshape)
 
 
+class Coerce(Expr):
```
llllllllll: can we add a docstring here with some examples?

blaze/expr/tests/test_expr.py
```diff
@@ -136,3 +136,15 @@ def test_hash_to_different_values():
     from blaze.expr.expressions import _attr_cache
     assert (expr, '_and') in _attr_cache
     assert (expr2, '_and') in _attr_cache
+
+
+def test_coerce():
+    s = symbol('s', var * float32)
+    assert str(s.coerce('int64')) == "s.coerce(to='int64')"
```
llllllllll: Can we check the `dshape` of the coerce node to make sure that it worked properly?

blaze/expr/expressions.py
cpcloud: Yep will do

blaze/expr/tests/test_expr.py
cpcloud: Good point

blaze/expr/tests/test_expr.py
cpcloud: In the case of tz-aware datetimes they are cast to object because pandas doesn't have a way to manage a contiguous array of tz-aware datetimes (yet).

cpcloud: Oh actually, I see what you mean ... they might be inferred as object, but you want whatever you're sending them to to see that they are datetimes

blaze/expr/expressions.py
```diff
@@ -22,7 +22,7 @@
 __all__ = ['Expr', 'ElemWise', 'Field', 'Symbol', 'discover', 'Projection',
            'projection', 'Selection', 'selection', 'Label', 'label', 'Map',
            'ReLabel', 'relabel', 'Apply', 'Slice', 'shape', 'ndim', 'label',
-           'symbol']
+           'symbol', 'Coerce', 'coerce']
```
llllllllll: Should the `coerce` function be available as a standalone function from the top level package or do you think this is most useful as a method

llllllllll:
Just had one comment on the visibility of the coerce function; but, this looks good to me

blaze/expr/expressions.py
cpcloud: I think I'll just leave it as a method for now and remove it from `__all__`

PR #1139, opened 2015-06-17
Update spark test

cpcloud:


blaze/compute/tests/test_sparksql.py
```diff
@@ -357,11 +357,12 @@ def test_strlen(ctx, db):
...
 def test_by_with_date(ctx, db, attr):
+    # TODO: investigate CSV writing precision between pandas 0.16.0 and 0.16.1
     expr = by(getattr(db.dates.ds, attr),
               mean=db.dates.amount.mean())
-    result = odo(compute(expr, ctx), set)
-    expected = odo(compute(expr, {db: {'dates': date_df}}), set)
-    assert result == expected
+    result = odo(compute(expr, ctx), pd.DataFrame).sort('mean')
+    expected = compute(expr, {db: {'dates': date_df}}).sort('mean')
+    tm.assert_frame_equal(result, expected, check_dtype=False)
```
llllllllll: does `assert_frame_equal` check floats within some epsilon?

cpcloud: it's basically this

```
return abs(desired - actual) < (0.5 * 10.0 ** -decimal)
```
with `decimal = 5`

cpcloud: Unless the two types of floats are different and we've passed in `check_dtype=False`. In that case `decimal = 3`. This is what would happen here because our inferred dshape is `?int32` for the `ds_attr` column when coming from spark (which tranlates to `float32` in pandas land) and when doing this directly on a `DataFrame` we have `float64`s already

cpcloud: I'll use the dshape from the spark expression so that we can compare apples to apples

llllllllll: Sounds good

cpcloud: hm it doesn't appear that `odo` will change the type of a `DataFrame`. I'll make an issue about it on `odo` and keep the test here as I have it

llllllllll: kk. Looks good.

PR #1140, opened 2015-06-17
Small revert and fix of broadcast_ndarray

cpcloud:
also fixes a small spark ordering issue in a test


llllllllll:
ty for the fix

PR #1145, opened 2015-06-22
find tables inside non-default schema when serving off a metadata

llllllllll:
The name of a table include the schema prefix which would disallow us from finding a table in a metadata object if it was not the default schema. This goes with: https://github.com/ContinuumIO/odo/pull/242

The `'.',join` is okay because that is what the key will look like when you have a dotted schema name: 

```python
In [2]: import sqlalchemy as sa

In [3]: ds = bz.Data(sa.MetaData('postgresql://localhost/test', schema='wow.this.sucks'))

In [4]: ds.data.tables
Out[4]: immutabledict({'wow.this.sucks.test': Table('test', MetaData(bind=Engine(postgresql://localhost/test)), Column('a', INTEGER(), table=<test>), schema='wow.this.sucks')})
```

### DIFF
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -757,6 +757,8 @@ def compute_up(expr, data, **kwargs):
 
 @toolz.memoize
 def table_of_metadata(metadata, name):
+    if metadata.schema is not None:
+        name = '.'.join((metadata.schema, name))
     if name not in metadata.tables:
         metadata.reflect(views=metadata.bind.dialect.supports_views)
     return metadata.tables[name]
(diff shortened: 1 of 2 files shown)


cpcloud:
Somewhat related: Should we discover all schemata in the database if just a database is given to resource?

For example:

```python
d = Data('postgresql://localhost/db')
d.my_schema.table1  # not the default schema
d.table1  # public schema
```

We can merge this PR before deciding on this.

llllllllll:
I think that would be a great addition; I will work on that as a separate pr. Does this pr look good though?

cpcloud:
yep :+1:

PR #1146, opened 2015-06-23
Allows the blaze server to stream results back

llllllllll:
Currently, the server always pulls the response into memory before serializing; however, this is not needed, we can send the result over in chunks.

This uses a new keyword: `chunksize` for the client. When `chunksize > 0`, the server will be instructed to chunk the response so that it is more memory efficient in some cases. This allows the client to decide which API to use.

Some questions I would like to bring up, why does the `compute.<serial_format>` endpoint return the shape and colnames when it is not needed by the client? Could we instead make an endpoint like: `/bokeh/compute` that lives in a seperate `flask.BluePrint` so that bokeh users can opt-in to these other endpoints at the cost of memory inefficiencies on the server. I think `dask` might also use the server (or there was an issue discussing this). Maybe we could have a similar `/dask/compute` endpoint. This also allows us to be more liquid with the server/client API without worrying about breaking other projects.

### DIFF
--- a/blaze/compatibility.py
+++ b/blaze/compatibility.py
@@ -60,6 +60,20 @@ if PY2:
         return MethodType(func, instance, type(instance))
 
     from itertools import izip_longest as zip_longest
+
+    from abc import ABCMeta
+    import collections
+    from types import ModuleType
+    collections_abc = ModuleType('compatibility.abc')
+    for k, v in vars(collections).items():
+        # Construct a collections.abc module with the abc's from collections.
+        if isinstance(v, ABCMeta):
+            setattr(collections_abc, k, v)
+    del ABCMeta
+    del ModuleType
+    del collections
+    del k
+    del v
 else:
     _inttypes = (int,)
     _strtypes = (str,)
@@ -68,6 +82,7 @@ else:
     boundmethod = MethodType
 
     from itertools import zip_longest
+    from collections import abc as collections_abc
 
 
 import io
--- a/blaze/compatibility.py
+++ b/blaze/compatibility.py
@@ -60,20 +60,6 @@ if PY2:
         return MethodType(func, instance, type(instance))
 
     from itertools import izip_longest as zip_longest
-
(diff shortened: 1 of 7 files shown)


blaze/compatibility.py
```diff
@@ -66,6 +80,7 @@ def boundmethod(func, instance):
     boundmethod = MethodType
 
     from itertools import zip_longest
+    from collections import abc as collections_abc
```
cpcloud: i believe abstract collections were kept in the toplevel `collections` module for back compat, were they not?

llllllllll: I didn't realize this. Should I just import them from collections at the use site?

cpcloud: yeah i'd import them at module level

blaze/server/server.py
```diff
@@ -316,20 +317,23 @@ def from_tree(expr, namespace=None):
...
+        return self._msg, self._status
+
+
+def _process_request(serial):
     try:
         payload = serial.loads(request.data)
     except ValueError:
         return ("Bad data.  Got %s " % request.data, 400)  # 400: Bad Request
 
+    chunksize = payload.get('chunksize', 0)
```
cpcloud: any reason not to use `None` instead of `0` here?

blaze/server/server.py
```diff
@@ -342,18 +346,56 @@ def compserver(serial_format):
         result = compute(expr, {leaf: dataset})
 
         if iscollection(expr.dshape):
-            result = odo(result, list)
+            result = odo(
+                result,
+                chunks(list) if chunksize > 0 else list,
```
cpcloud: if you used `None` then you can say `if chunksize else list` and the same for when you call it in the `/compute.<format>` route

llllllllll:
@cpcloud can you take a look at this now that we have changed the enpoint stuff?

llllllllll:
Closing due to inactivity

PR #1147, opened 2015-06-25
Move postgresql tests to their own file

cpcloud:
Allows testing sqlite and other sql tests while allowing us to skip postgresql
tests on machines that don't have psycopg2 installed


cpcloud:
merging on pass, this fixes some failing windows tests

PR #1148, opened 2015-06-25
Explicit int64 for windows comparison

cpcloud:
Fixes some minor test failures because of different default integer sizes


cpcloud:
merging on pass

PR #1152, opened 2015-06-29
Fix naming issues with pandas Series construction

cpcloud:
Tidies up a few areas where we were playing fast and loose with Series names.


cpcloud:
merging on pass

PR #1159, opened 2015-07-07
Adds distinct on

llllllllll:


blaze/expr/collections.py
```diff
@@ -97,8 +102,17 @@ class Distinct(Expr):
     >>> from blaze.compute.python import compute
     >>> sorted(compute(e, data))
     [('Alice', 100, 1), ('Bob', 200, 2)]
+
+    Using a subset
```
cpcloud: You have to put a space after these comments otherwise sphinx won't render them correctly.

cpcloud:
pull in master after i merge #1161, numba changed the way they `__repr__`d functions so i went ahead and wrote some proper tests instead of depending on doctest

blaze/compute/numpy.py
```diff
@@ -199,6 +199,8 @@ def compute_up(t, x, **kwargs):
 
 @dispatch(Distinct, np.ndarray)
 def compute_up(t, x, **kwargs):
+    if t.on:
+        raise ValueError('numpy backend cannot specify what to distinct on')
```
cpcloud: maybe say `numpy backend cannot specify what column to distinct on`

blaze/compute/pandas.py
```diff
@@ -192,9 +192,18 @@ def compute_up(t, s, **kwargs):
...
-    return df.drop_duplicates().reset_index(drop=True)
+    return df.drop_duplicates(
+        subset=t.on if t.on else None,
+    ).reset_index(drop=True)
+
+
+@dispatch(Distinct, Series)
+def compute_up(t, s, **kwargs):
+    if t.on:
+        raise ValueError('series distinct cannot specify what to distinct on')
```
cpcloud: Maybe have a message here that indicates that this operation doesn't make sense

llllllllll: Yeah, I will try to think of some shared language for this an the ndarray case. Any ideas?

llllllllll: Actually, I think that this is only shared when the ndarray is a vector not a matrix

cpcloud: the ndarray case makes sense for struct and recarrays but not for anything else so sharing the language doesn't make sense in every case.

i'd do something like this:

```python
if getattr(data.dtype, 'names', None) is None:
    raise ValueError('operation does not make sense')
else:
    raise NotImplementedError('makes sense, but numpy does not have an API for this operation')
```

llllllllll: For the recarray case we could just throw it into a df and then pull the underlying array back out

cpcloud: i hope that doesn't seem like a lot of extra work. i think we should try to provide error messages as soon and as accurate as possible

cpcloud: > For the recarray case we could just throw it into a df and then pull the underlying array back out

Yep, that sounds good. The only issue is converting strings back to their original fixed length dtype, but that shouldn't be too difficult

llllllllll: This is not too much, I agree that error messages are very important for the usability of a project.

blaze/expr/collections.py
llllllllll: done

llllllllll:
note: I need to get the string conversion for the pandas round trip and add tests for np and probably the other error cases before this is ready.

llllllllll:
And it's passing!

blaze/compute/pandas.py
```diff
@@ -192,9 +192,18 @@ def compute_up(t, s, **kwargs):
     return result
 
 
-@dispatch(Distinct, (DataFrame, Series))
+@dispatch(Distinct, DataFrame)
 def compute_up(t, df, **kwargs):
-    return df.drop_duplicates().reset_index(drop=True)
+    return df.drop_duplicates(
+        subset=t.on if t.on else None,
```
cpcloud: minor style comment: should these be `t.on or None`?

llllllllll: ah yes, then I previously was doing a check against `None`

blaze/compute/tests/test_spark.py
```diff
@@ -170,6 +170,14 @@ def test_distinct(rdd):
         set(['Alice', 'Bob'])
 
 
+@pytest.mark.xfail(
+    raises=NotImplementedError,
+    reason='cannot specify columns to distinct on yet',
+)
+def test_distinct_in(rdd):
```
cpcloud: should this be `test_distinct_on`?

llllllllll: oops, yup

blaze/compute/numpy.py
```diff
@@ -198,8 +198,23 @@ def compute_up(t, x, **kwargs):
...
 
 @dispatch(Distinct, np.ndarray)
-def compute_up(t, x, **kwargs):
-    return np.unique(x)
+def compute_up(t, arr, **kwargs):
+    if t.on:
+        if getattr(arr.dtype, 'names', None) is not None:
+            return pd.DataFrame.from_records(arr).drop_duplicates(
+                subset=t.on if t.on else None,
+            ).reset_index(drop=True).value.astype(arr.dtype)
```
cpcloud: i think you might not be covering this in tests since `value` isn't an attribute of `DataFrame`s

cpcloud: also, you probably want to do `to_records` here because `values` will use a lot more memory if you have different types in your `DataFrame`.

blaze/compute/numpy.py
```diff
@@ -197,9 +197,21 @@ def compute_up(t, x, **kwargs):
...
+@compute_up.register(Distinct, np.recarray)
+def recarray_distinct(t, rec, **kwargs):
+    return pd.DataFrame.from_records(rec).drop_duplicates(
+        subset=t.on or None).to_records(index=False).astype(rec.dtype)
+
+
 @dispatch(Distinct, np.ndarray)
-def compute_up(t, x, **kwargs):
-    return np.unique(x)
+def compute_up(t, arr, _recarray_distinct=recarray_distinct, **kwargs):
```
cpcloud: any reason not to just call `recarray_distinct` in the body of the function?

llllllllll: this is just a slight perf increase.

cpcloud: i don't think it's necessary. what kind perf increase are we talking? saving the cost of a globa lookup?

llllllllll: Yeah, it swaps out a global lookup for a local lookup. It's not a huge gain but it does slightly cut down the cost of calling the function.

cpcloud:
ok this looks good to me

PR #1166, opened 2015-07-09
Fix tarball

cournape:
The `0.8.1` tarball is broken, and `setup.py` cannot be run from it. This fixes it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,8 @@
 
 from __future__ import absolute_import, division, print_function
 
+import setuptools
+
 import os
 import sys
 from fnmatch import fnmatch
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,7 +4,7 @@ recursive-include docs *.rst
 include requirements.txt
 include requirements-strict.txt
 include setup.py
-include README.md
+include README.rst
 include LICENSE.txt
 include MANIFEST.in
 


setup.py
```diff
@@ -2,6 +2,8 @@
 
 from __future__ import absolute_import, division, print_function
 
+import setuptools
```
cpcloud: What does importing `setuptools` do?

mrocklin: setuptools monkey patches distutils.  Might want to check with @asmeurer to ensure that this is ok.

cournape: Note that you have to monkey patch w/ setuptools if you use `install_requires` or `entry_points`.

Without it, those options are simply ignored.

cpcloud: That doesn't seem to be true, since if you install blaze from source you get the `blaze-server` script and it behaves as expected. Are you talking about tarball builds only?

cournape: I can only see that happening if you use a environment that somehow import setuptools before running the `setup.py`. Note that distutils documentation do not mentioned either option.

If I use python 2.7.9 from ubuntu, this is what I see:

```
/usr/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'entry_points'
  warnings.warn(msg)
/usr/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'install_requires'
  warnings.warn(msg)
running config
```

Those warnings disappear after importing setuptools.

cpcloud: ok

cpcloud:
fixed by #1168

PR #1170, opened 2015-07-10
allow blaze client to disable ssl verification

llllllllll:
verifies the cert by default; however, we want to disable this

Side note: we might want to think about making ssl the default. There is no downside really and it helps people who might not have thought about the fact that they will be streaming potentially sensitive data over the interned.

note: wait for #1172 and then change the .get to .post

### DIFF
--- a/blaze/server/client.py
+++ b/blaze/server/client.py
@@ -1,11 +1,13 @@
 from __future__ import absolute_import, division, print_function
 
+import warnings
+
+from datashape import dshape
 import flask
 from flask.testing import FlaskClient
-import requests
-
 from odo import resource
-from datashape import dshape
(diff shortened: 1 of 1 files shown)


cpcloud:
> we might want to think about making ssl the default.

isn't that what you've implemented here?

cpcloud:
or do you mean use `https://` by default?

llllllllll:
> or do you mean use https:// by default?

yeah, this just makes it so that if it uses ssl with a custom cert that it doesn't check the cert. I imagine most production setups will want to serve their data with https.

cpcloud:
@llllllllll this can be merged after fixing up conflicts and passing

PR #1172, opened 2015-07-13
Only allow POST requests on compute.<format> route

cpcloud:


llllllllll:
Not sure what the conda stuff is for, but the `post` change looks good to me.

Do we need to worry about how this will affect bokeh integration?

cpcloud:
conda stuff was an accident i branched off the wrong branch

cpcloud:
cc @bryevdv @fpliger are you guys ok with this change?

bryevdv:
Looks like the `BlazeDataSource` already uses `POST`:

https://github.com/bokeh/bokeh/blob/master/bokehjs/src/coffee/source/blaze_data_source.coffee#L31

So should hopefully be ok?

fpliger:
> Looks like the BlazeDataSource already uses POST:

ah, right. This looks like a safe change.. we can pay particular attention to that and test

cpcloud:
thanks for the comments guys. merging

PR #1175, opened 2015-07-13
ENH: allows server to use basic auth

llllllllll:


cpcloud:
small test and a release note in 0.8.3 then good to go!

llllllllll:
The flask client doesn't support passing the auth keyword

llllllllll:
tests added.

PR #1176, opened 2015-07-13
Pass the serialization format as a mimetype

cpcloud:
the serialization format json, msgpack, etc is now passed as a vendor specific mime type instead of being part of the URL:

```python
requests.get('/compute',
             data=dict(expr=to_tree(expr)),
             headers={'Content-Type': 'application/vnd.blaze+json'})
```

### DIFF
--- a/blaze/server/client.py
+++ b/blaze/server/client.py
@@ -95,14 +95,28 @@ def discover(c):
     return c.dshape
 
 
+def mimetype(serial):
+    """Function to generate a blaze serialization format mimetype put into a
+    dictionary of headers for consumption by requests.
+
+    Examples
+    --------
+    >>> from blaze.server.serialization import msgpack
+    >>> mimetype(msgpack)
+    {'Content-Type': 'application/vnd.blaze+msgpack'}
+    """
+    return {'Content-Type': 'application/vnd.blaze+%s' % serial.name}
+
+
 @dispatch(Expr, Client)
 def compute_down(expr, ec, **kwargs):
     from .server import to_tree
     tree = to_tree(expr)
 
     serial = ec.serial
-    r = requests.post('%s/compute.%s' % (ec.url, serial.name),
-                      data=serial.dumps({'expr': tree}))
+    r = requests.post('%s/compute' % ec.url,
+                      data=serial.dumps({'expr': tree}),
+                      headers=mimetype(serial))
 
     if not ok(r):
         raise ValueError("Bad response: %s" % reason(r))
--- a/blaze/server/client.py
+++ b/blaze/server/client.py
@@ -37,7 +37,7 @@ def _request(method, client, url, params=None, **kwargs):
     with warnings.catch_warnings():
         warnings.simplefilter('ignore', InsecureRequestWarning)
         return method(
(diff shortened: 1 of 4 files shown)


cpcloud:
cc @llllllllll

blaze/server/client.py
```diff
@@ -130,8 +143,9 @@ def compute_down(expr, ec, **kwargs):
     serial = ec.serial
     r = post(
         ec,
-        'compute.{serial}'.format(serial=serial.name),
+        '/compute',
```
llllllllll: this will generate 2 slashes. we might want to update the `_request` function to not add the slash.

blaze/server/server.py
```diff
@@ -316,12 +317,14 @@ def from_tree(expr, namespace=None):
...
-def compserver(serial_format):
+def compserver():
+    content_type = request.headers['content-type']
+
     try:
-        serial = _get_format(serial_format)
+        serial = _get_format(*re.match(r'^application/vnd\.blaze\+(.+)$',
+                                       content_type).groups())
     except KeyError:
         return 'Unsupported serialization format', 404
```
llllllllll: This should be a `415` now

blaze/server/client.py
cpcloud: fixed

blaze/server/server.py
```diff
@@ -316,14 +317,19 @@ def from_tree(expr, namespace=None):
...
+
+@api.route('/compute', methods=['POST', 'HEAD', 'OPTIONS'])
 @crossdomain(origin='*', methods=['POST', 'HEAD', 'OPTIONS'])
-def compserver(serial_format):
+def compserver():
+    content_type = request.headers['content-type']
+
     try:
-        serial = _get_format(serial_format)
+        serial = _get_format(*mimetype_regex.match(content_type).groups())
```
llllllllll: The * makes me think that this could possibly be more than one value; could we just take [0] instead?

llllllllll:
Just one comment, but looks good

blaze/server/server.py
cpcloud: yep

blaze/server/server.py
```diff
@@ -316,14 +317,19 @@ def from_tree(expr, namespace=None):
...
+
+@api.route('/compute', methods=['POST', 'HEAD', 'OPTIONS'])
 @crossdomain(origin='*', methods=['POST', 'HEAD', 'OPTIONS'])
-def compserver(serial_format):
+def compserver():
+    content_type = request.headers['content-type']
+
     try:
-        serial = _get_format(serial_format)
+        serial = _get_format(mimetype_regex.match(content_type).groups()[0])
```
llllllllll: actually, this needs to check that match is not none before taking groups()[0] or we get 405 instead of 415

cpcloud: good catch

cpcloud: if `None` should that also be a 415?

llllllllll: Yeah, None means we don't know what the media type is; however, a keyerror means we know what it is but this particular server doesn't use that format

Issue #1184, opened 2015-07-22
blaze.interactive.InteractiveSymbol.tail(n)


scls19fr:
Hello,

I'm new to blaze.

I'm connecting to a MySQL Database table

![capture d ecran 2015-07-22 a 21 50 33](https://cloud.githubusercontent.com/assets/109167/8835591/c0346316-30bb-11e5-9919-1dfc20aa672b.png)

    import blaze as bz
    uri = 'mysql+mysqlconnector://user:password@127.0.0.1:3306/meteo::data'
    dat = bz.Data(uri)

There is

    dat.head(5)

but no

    dat.tail(5)

Moreover `dat[:5]` provide first 5 rows

but `dat[-5:]` raises `ValueError: start value of slice cannot be negative with a SQL backend`

I can get this "tail"... but that's not very convenient (compared to Pandas)

    dat[int(dat.count())-5:int(dat.count())]

That's also odd to see that

    dat[dat.count()-5:dat.count()]

return nothing.

Sorry for my newbies questions

Kind regards

llllllllll:
You can do:

```python
dat.sort(ascending=False).head(5)
```

llllllllll:
To go into more detail, the use of a negative slice was not implemented because we are using limits and offsets to implement slicing in the first place which do not support a negative slice. One idea would be to implement it with the `.count() - n` trick. Note that this is an extremely inneficient operation in sql. If you need to do a lot of slicing, I would suggest odoing your dataset into another format first. 

About the slice returning nothing, what do you mean by that?
Did it return an empty data set?
I get the results that I expect when testing against postgres (I don't have a mysql instance to test on)

```python
In [1]: raw = pd.DataFrame(np.arange(15).reshape(5, 3), columns=list('abc'))

In [2]: raw
Out[2]: 
    a   b   c
0   0   1   2
1   3   4   5
2   6   7   8
3   9  10  11
4  12  13  14

In [3]: tbl = odo(raw, resource('postgresql://localhost/test::test', dshape=discover(raw)))

In [4]: ds = bz.Data(tbl)

In [5]: ds[ds.count() - 2:]
/home/joejev/quantopian/blaze/blaze/compute/sql.py:856: UserWarning: The order of the result set from a Slice expression computed against the SQL backend is not deterministic.
  warnings.warn('The order of the result set from a Slice expression '
Out[5]: 
    a   b   c
0   9  10  11
1  12  13  14
```

[...]

PR #1187, opened 2015-07-24
ENH: adds Tail (opposite of head)

llllllllll:
implements a tail operation #1184

Not really sure what we want to do about sql.
This is NOT something you want to do on sql for perfomance reasons so idk if we should implement.


mrocklin:
No objection to this.  At one point Jeff mentioned that it would be nice to add this to the `repr`.  I don't have a concrete use case though.

llllllllll:
Good to merge then?

cpcloud:
fine by me

PR #1198, opened 2015-08-07
Fix naming of reductions when using a generated leaf from the compute sequence

cpcloud:


llllllllll:
Was the old name: `__sum`?

cpcloud:
yes

cpcloud:
while that's a valid python identifier, not all system accept this, moreover it exposes an implementation detail

llllllllll:
Gotcha, this looks good then. Want to put an entry in whatsnew to mention the fix?

cpcloud:
ah! thanks for reminding me

cpcloud:
merging on pass

PR #1199, opened 2015-08-09
MAINT: Use copydoc and cleanup some docstring additions

llllllllll:


llllllllll:
I don't understand the build output at all.

cpcloud:
@llllllllll this should be fixed

llllllllll:
cool, merging on pass

blaze/expr/arrays.py
```diff
@@ -49,11 +53,9 @@ def transpose(expr, axes=None):
         axes = tuple(axes)
     return Transpose(expr, axes)
 
-transpose.__doc__ = Transpose.__doc__
-
-
+@copydoc(Transpose)
 def T(expr):
-    return transpose(expr)
+    transpose
```
cpcloud: does this need to be `return transpose(expr)`?

llllllllll: yes it does

cpcloud:
@llllllllll looks like something strange is going on ... `dot` and `tensordot` are returning incorrect results with this PR

llllllllll:
I accidentally flipped lhs and rhs, I will fix a

PR #1201, opened 2015-08-11
WIP: Remove ScalarSelect generation

cpcloud:


cpcloud:
@llllllllll can you take a look at this when you get a chance

blaze/compute/sql.py
```diff
@@ -620,11 +644,24 @@ def compute_up(t, s, **kwargs):
...
 
-@dispatch(Label, ClauseElement)
+@dispatch(Label, ColumnElement)
 def compute_up(t, s, **kwargs):
     return s.label(t.label)
 
 
+@dispatch(Label, FromClause)
+def compute_up(t, s, **kwargs):
+    assert len(s.c) == 1, \
```
llllllllll: do we ever expect this to be hit? If so, could we make this raise a ValueError. Otherwise I am fine with leaving the assertion.

cpcloud: No, this is me expressing an assumption about the code that follows it. If it fails, then I've hit a case that I didn't handle. The common case is that I've (re)labeled the output of a column that's from a more complex statement than just selecting the column

llllllllll: I understand, just wanted to make sure that this was not a normal case.

cpcloud: Yes it would be very strange to get a label expression on something with more than a single column

cpcloud: Let me add a test that tries to fail that assertion in a simple case

llllllllll: kk, I wouldn't worry too much if it is a sufficiently rare case.

llllllllll:
This looks good, just needs a whatsnew entry

PR #1206, opened 2015-08-13
Fix incorrect generation of count on selectables

### LINKED ISSUE #1202: by function then count_values creates wrong result.
I'm using the lahman2013.sqlite file to learn blaze.   I believe the following is producing the wrong result.

```
import blaze as bz
_db = bz.Data('sqlite:///data/lahman2013.sqlite')

_sy=bz.by(_db.Teams.name, start_year=_db.Teams.yearID.min())
print(str(bz.compute(_sy, post_compute=False)))
```
produces
```
SELECT "Teams".name, min("Teams"."yearID") AS start_year 
FROM "Teams" GROUP BY "Teams".name
```
which seems correct to me.

I want to now use this result (the first year a team was in the MLB) to see how many teams entered the MLB for each year.  I believe the following should produce that..

```
print(list(_sy.start_year.count_values())
```
but it produces incorrect counts.  When I look at the sql produced I get the following:

```
print(str(bz.compute(_sy.start_year.count_values(), post_compute=False)))

SELECT anon_1.start_year, anon_1.count 
FROM (SELECT alias_1.start_year AS start_year, count(alias_2.start_year) AS count 
FROM (SELECT min("Teams"."yearID") AS start_year 
FROM "Teams" GROUP BY "Teams".name) AS alias_1, (SELECT alias_1.start_year AS start_year 
FROM (SELECT min("Teams"."yearID") AS start_year 
FROM "Teams" GROUP BY "Teams".name) AS alias_1) AS alias_2 GROUP BY alias_1.start_year) AS anon_1 ORDER BY anon_1.count DESC
```
which is incorrect.  It should produce something close to:

```
select start_year, count(*)
from
 (SELECT "Teams".name, min("Teams"."yearID") AS start_year 
FROM "Teams" GROUP BY "Teams".name)
group by start_year
```

I would love to take a stab at fixing this myself.  What I would like is for people to verify that I'm doing this correctly..

Thanks!

Billy

cpcloud:
Closes #1202

cc @earney


cpcloud:
@earney this was a bit hairy ... happy to walk you through how and why the fix works if you want

blaze/compute/tests/test_sql_compute.py
```diff
@@ -1678,6 +1678,30 @@ def test_attribute_on_filter_transform_groupby():
     assert normalize(str(result)) == normalize(expected)
 
 
+def test_baseball_nested_by():
+    data = resource('sqlite:////Users/pcloud/Downloads/lahman2013.sqlite')
```
llllllllll: This should probably be a relative path

cpcloud: good point :|

earney:
Phillip,
I'll take a look at the bug fix, and if I have any questions, I'll let you
know.
Thanks!

cpcloud:
@llllllllll any more comments here?

earney:
I think we are good.  Thanks for working on this!

llllllllll:
lgtm

PR #1207, opened 2015-08-14
Add binary operations on single column selects

cpcloud:
This is necessary when e.g., subsetting on a Merge expression


cpcloud:
merging on pass

Issue #916, opened 2015-01-09
decouple Math from UnaryOp


cpcloud:
Not all functions defined in Python's `math` module are unary, so we should not force them to be.

Currently the method resolution order of `Math` is this:

```python
In [3]: from blaze.expr.math import Math

In [4]: Math.mro()
Out[4]:
[blaze.expr.math.Math,
 blaze.expr.arithmetic.UnaryOp,
 blaze.expr.expressions.ElemWise,
 blaze.expr.expressions.Expr,
 blaze.expr.core.Node,
 object]
```

A notable example of a function that isn't unary is `math.hypot`:

```python
In [8]: math.hypot?
Type:        builtin_function_or_method
String form: <built-in function hypot>
Docstring:
hypot(x, y)

Return the Euclidean distance, sqrt(x*x + y*y).
```

earney:
cpcloud,

If you don't mind, I would like to take a stab at this issue.  is that okay with you?

Billy

cpcloud:
@earney Please do!

cpcloud:
Let me know if you need any help.

PR #1209, opened 2015-08-17
Issue #916 (decouple Math from Unary Op)

earney:
Creates a new class named Op that allows any number of operands (but for now just supports Unary and Binary Operators)

ref:  #916


blaze/expr/arithmetic.py
```diff
@@ -50,6 +50,55 @@ def name(o):
         return None
 
 
+class Op(ElemWise):
+    def __init__(self, ops):
+        if not isinstance(ops, (list, tuple,)):
```
llllllllll: This should be an ABC or something that has 2 subclasses, `UnaryOp` and `BinaryOp`. This means that we can isolate the responsibilities of each class

llllllllll:
Also, it looks like you are missing some commit from master. Do you want to rebase against master?

blaze/expr/arithmetic.py
```diff
@@ -50,6 +50,55 @@ def name(o):
...
+           rhs = parenthesize(eval_str(self._ops[1]))
+           return '%s %s %s' % (lhs, self.symbol, rhs)
+
+        if len(self._ops) == 1:
+           return '%s(%s)' % (self.symbol, eval_str(self._ops[0]))
+
+        raise BaseException('Invalid number of ops: %s' % len(self._ops))
+
+    @property
+    def _name(self):
```
cpcloud: Was this the naming policy before this PR?

blaze/expr/arithmetic.py
cpcloud: This could also be generalized to an operator with N arguments, and that would simplify some of the code below. For example,

```python
class Op(ElemWise):
    __slots__ = 'operands',
    __inputs__ = 'operands',
    def __init__(self, operands):
        self._operands = operands
```

Then, put the different `__str__` implementations in the subclass. Generally, the parent class shouldn't know about specific details of child classes.

llllllllll: That actually seems nicer than a unary/binary split. This could let us get some fun things like ifexprs in the mix maybe.

earney:
Thanks for the suggestions.  I'll look into implementing these ASAP.

earney:
Do you guys have any idea why I'm getting the below error(in travis-cl):
```
collecting 91 items / 2 skipped/home/travis/build.sh: line 41:  2883 Aborted                 (core dumped) coverage run --include='blaze/*
```

llllllllll:
There looks like there is some stack overflow or something happening. I am not totally sure, can you run the build locally?

blaze/expr/arithmetic.py
```diff
@@ -50,24 +50,48 @@ def name(o):
...
     def __init__(self, lhs, rhs):
-        self.lhs = lhs
-        self.rhs = rhs
+        Op.__init__(self, [lhs, rhs])
 
     def __str__(self):
-        lhs = parenthesize(eval_str(self.lhs))
-        rhs = parenthesize(eval_str(self.rhs))
+        lhs = parenthesize(eval_str(self._operands[0]))
+        rhs = parenthesize(eval_str(self._operands[1]))
```
cpcloud: suggest

```python
from toolz import compose
return (' %s ' % self.symbol).join(map(compose(parenthesize, eval_str), self._operands))
```

blaze/expr/arithmetic.py
```diff
@@ -50,24 +50,48 @@ def name(o):
...
+
+class BinOp(Op):
+    #__slots__ = '_hash', 'lhs', 'rhs'
+    __slots__ = '_hash'
     __inputs__ = 'lhs', 'rhs'
 
     def __init__(self, lhs, rhs):
-        self.lhs = lhs
-        self.rhs = rhs
+        Op.__init__(self, [lhs, rhs])
```
cpcloud: suggest `super(BinOp, self).__init__(self, [lhs, rhs])`

blaze/expr/arithmetic.py
```diff
@@ -50,24 +50,48 @@ def name(o):
...
+    @property
+    def rhs(self):
+        return self._operands[1]
+
+    @property
     def _name(self):
         if not isscalar(self.dshape.measure):
             return None
-        l, r = name(self.lhs), name(self.rhs)
+        l, r = name(self._operands[0]), name(self._operands[1])
```
cpcloud: you can use the `lhs` and `rhs` properties here

blaze/expr/arithmetic.py
```diff
@@ -120,14 +135,19 @@ def maxshape(shapes):
...
+    #__slots__ = '_hash', '_child',
+    __slots__ = '_hash',
 
     def __init__(self, child):
-        self._child = child
+        Op.__init__(self, [child])
 
     def __str__(self):
-        return '%s(%s)' % (self.symbol, eval_str(self._child))
+        return '%s(%s)' % (self.symbol, eval_str(self._operands[0]))
```
cpcloud: you can leave the code the same here since you define the `_child` property

blaze/expr/arithmetic.py
```diff
@@ -135,11 +155,11 @@ def symbol(self):
 
     @property
     def dshape(self):
-        return DataShape(*(shape(self._child) + (self._dtype,)))
+        return DataShape(*(shape(self._operands[0]) + (self._dtype,)))
```
cpcloud: same here, no need to change this code because you're defining `self._child` to be this

cpcloud:
@earney looks like most of your changes to use `self._operands[...]` are not necessary (this is a good thing!) since you're defining the things that were originally slots as properties

earney:
Thanks guys!  I'll make the suggested changes.

earney:
Guys,
I think I have nailed down where the stack over flow is occurring.  I'm stumped..  can you take a look?
The stack overflow is occurring at expressions.py, line 327

Any help is greatly appreciated!

blaze/expr/arithmetic.py
```diff
@@ -50,18 +50,38 @@ def name(o):
...
+        return tuple(result)
+
+class BinOp(Op):
+    __slots__ = '_hash'
     __inputs__ = 'lhs', 'rhs'
 
     def __init__(self, lhs, rhs):
-        self.lhs = lhs
-        self.rhs = rhs
+        super(BinOp, self).__init__(self, [lhs, rhs])
```
cpcloud: this should be `super(BinOp, self).__init__([lhs, rhs])`. my apologies if my original suggestion wasn't this

blaze/expr/arithmetic.py
```diff
@@ -120,16 +131,20 @@ def maxshape(shapes):
...
 
 
-class UnaryOp(ElemWise):
-    __slots__ = '_hash', '_child',
+class UnaryOp(Op):
+    __slots__ = '_hash'
 
     def __init__(self, child):
-        self._child = child
+        Op.__init__(self, [child])
```
cpcloud: write this as `super(Op, self).__init__([child])`

blaze/expr/expressions.py
```diff
@@ -306,7 +311,7 @@ class Field(ElemWise):
     >>> points['space station'].dshape
     dshape("5 * 3 * float64")
     """
-    __slots__ = '_hash', '_child', '_name'
+    #__slots__ = '_hash', '_child', '_name'
```
llllllllll: why are you commenting out the slots?

blaze/expr/arithmetic.py
```diff
@@ -186,7 +201,8 @@ class Div(Arithmetic):
 
     @property
     def _dtype(self):
-        lhs, rhs = discover(self.lhs).measure, discover(self.rhs).measure
+        lhs = discover(self._operands[0]).measure, 
+        rhs = discover(self._operands[1]).measure
```
cpcloud: revert the above two lines, they aren't necessary because you've already defined `self.lhs` and `self.rhs`

blaze/expr/arithmetic.py
```diff
@@ -196,7 +212,8 @@ class FloorDiv(Arithmetic):
...
```
cpcloud: same here revert these two changes

blaze/expr/arithmetic.py
```diff
@@ -378,7 +395,7 @@ class Not(UnaryOp):
     _dtype = ct.bool_
 
     def __str__(self):
-        return '~%s' % parenthesize(eval_str(self._child))
+        return '~%s' % parenthesize(eval_str(self._operands[0]))
```
cpcloud: again, there's no need to change anything because you've already defined `self._child`

blaze/expr/expressions.py
```diff
@@ -122,6 +122,8 @@ def schema(self):
 
     @property
     def fields(self):
+        print('def fields')
+        print(self.__class__)
```
cpcloud: please remove these print statements

blaze/expr/arithmetic.py
cpcloud: more over you're ending the first line with `,` which will make it a single element tuple, which is incorrect

blaze/expr/arithmetic.py
```diff
@@ -50,18 +50,38 @@ def name(o):
         return None
 
 
-class BinOp(ElemWise):
-    __slots__ = '_hash', 'lhs', 'rhs'
+class Op(ElemWise):
+    __slots__ = '_hash', 'operands'
+    __inputs__ = 'operands'
```
cpcloud: this needs to be a `tuple`, because it's used everywhere like this

```python
map(function, [getattr(expr, inp) for inp in expr.__inputs__])
```

in your implementation that would try to call

```
getattr(expr, 'o')
getattr(expr, 'p')
...
```
and of course expressions don't have those attributes

cpcloud: the solution is to write it like this:

```python
__inputs__ = 'operands',
```

blaze/expr/arithmetic.py
```diff
@@ -225,12 +242,12 @@ class USub(UnaryOp):
     symbol = '-'
 
     def __str__(self):
-        return '-%s' % parenthesize(eval_str(self._child))
+        return '-%s' % parenthesize(eval_str(self._operands[0]))
```
cpcloud: revert this change

blaze/expr/arithmetic.py
```diff
@@ -225,12 +242,12 @@ class USub(UnaryOp):
...
 
     @property
     def _dtype(self):
         # TODO: better inference.  -uint -> int
-        return self._child.schema
+        return self._operands[0].schema
```
cpcloud: revert this change

blaze/expr/core.py
```diff
@@ -66,7 +66,7 @@ class Node(object):
 
     blaze.expr.expressions.Expr
     """
-    __inputs__ = '_child',
+    __inputs__ = ''
```
cpcloud: revert this change

blaze/expr/arithmetic.py
```diff
@@ -50,18 +50,38 @@ def name(o):
...
+    def __init__(self, operands):
+        self._operands=operands
+
+    @property
+    def _inputs(self):
+        result = []
+        for _op in self._operands:
+            if isinstance(_op, Expr):
+               result.append(_op)
+        return tuple(result)
```
cpcloud: I think you can bypass the list -> tuple conversion and just construct the tuple directly:

```
return tuple(op for op in self._operands if isinstance(op, Expr))
```

blaze/expr/arithmetic.py
```diff
@@ -50,18 +51,34 @@ def name(o):
         return None
 
 
-class BinOp(ElemWise):
-    __slots__ = '_hash', 'lhs', 'rhs'
+class Op(ElemWise):
+    __slots__ = '_hash', 'operands'
+    __inputs__ = 'operands',
```
cpcloud: these should both be `_operands` not `operands`

cpcloud:
@earney What's the status on this?

cpcloud:
Happy to move to 0.9.0 if you're strapped for time.

earney:
Phillip,

Okay..  moving to 0.9.0 sounds okay to me.

cpcloud:
@earney what's the status on this?

cpcloud:
closing as stale

PR #1211, opened 2015-08-18
BUG/ENH: Add single field filtering

cpcloud:


llllllllll:
Is this just an optimization or does this not currently work?

cpcloud:
The first test case `t.amount[t.amount > 0]` doesn't work

cpcloud:
@llllllllll Any more comments? Would like to merge this soon

blaze/compute/sql.py
```diff
@@ -176,6 +176,12 @@ def compute_up(t, s, **kwargs):
     return getattr(t, 'op', getattr(safuncs, sym, getattr(sa.func, sym)))(s)
 
 
+@dispatch(Selection, sa.sql.ColumnElement)
+def compute_up(expr, data, scope=None, **kwargs):
+    predicate = compute(expr.predicate, data, post_compute=False)
+    return sa.select([data]).where(predicate)
```
llllllllll: should we use the `select` function from this module here?

llllllllll:
one question but lgtm

blaze/compute/sql.py
cpcloud: i'm -1 on using that function as it hides a subtle aspect of sqlalchemy that can sometimes cause weird issues if you're not very careful about nesting selects

cpcloud: I think in all cases I know the kind of thing I'm getting back (`Select`, `Table`, `ColumnElement`, etc)

cpcloud: I'll probably remove that function in an upcoming PR

llllllllll: Sounds good. I just wanted to make sure that this wasn't solving some subtle issue that has been discovered in the past. I would also be in favor of removing at it is confusing when used with `sa.select`

PR #1218, opened 2015-08-27
Join type promotion

### LINKED ISSUE #1193: Failing joins on field that is nullable in one table but not in the other
Given following tables (postgresql db):
```
example> \d table1
+----------+---------+-------------+
| Column   | Type    | Modifiers   |
|----------+---------+-------------|
| id       | integer | not null    |
| field    | integer |             |
+----------+---------+-------------+
Indexes:
    "table1_pkey" PRIMARY KEY, btree (id)

example> \d table2
+----------+---------+-------------+
| Column   | Type    | Modifiers   |
|----------+---------+-------------|
| id       | integer | not null    |
| field    | integer | not null    |
+----------+---------+-------------+
Indexes:
    "table2_pkey" PRIMARY KEY, btree (id)
```

Following code fails:

```
join(table1, table2, 'field')
```

With this error message:

```
TypeError                                 Traceback (most recent call last)
<ipython-input-14-2a191a07eec9> in <module>()
----> 1 join(database.table1, database.table2, 'field')

/home/user/anaconda/envs/env/lib/python2.7/site-packages/blaze/expr/collections.pyc in join(lhs, rhs, on_left, on_right, how, suffixes)
    435                          (lhs, rhs))
    436     if types_of_fields(on_left, lhs) != types_of_fields(on_right, rhs):
--> 437         raise TypeError("Schema's of joining columns do not match")
    438     _on_left = tuple(on_left) if isinstance(on_left, list) else on_left
    439     _on_right = (tuple(on_right) if isinstance(on_right, list)

TypeError: Schema's of joining columns do not match
```

Types returned by types_of_fields function:
```
types_of_fields('field', database.table1) #  ?int32
types_of_fields('field', database.table2) #  ctype("int32")  
```

llllllllll:
closes #1193

__edit__: this pr also does generic type promotion of joined fields.

needs https://github.com/blaze/datashape/pull/172

This also fixes an issue that could occur when joining on a single column name passed as a string when other column names were substrings of the join column.

### DIFF
--- a/blaze/compute/tests/test_python_compute.py
+++ b/blaze/compute/tests/test_python_compute.py
@@ -358,16 +358,28 @@ def test_graph_double_join():
     t_arc = symbol('t_arc', 'var * {a: int32, b: int32}')
     t_wanted = symbol('t_wanted', 'var * {name: string}')
 
+    # >>> compute(join(t_idx, t_arc, 'b'), {t_idx: idx, t_arc: arc})
+    # [[1, A, 3],
+    #  [1, A, 2],
+    #  [1, A, 5],
+    #  [3, C, 1],
+    #  [3, C, 2],
+    #  [3, C, 4],
+    #  [3, C, 5],
+    #  [6, F, 1],
+    #  [6, F, 2],
+    #  [6, F, 4]]
+
     j = join(join(t_idx, t_arc, 'b'), t_wanted, 'name')[['name', 'b', 'a']]
 
     result = compute(j, {t_idx: idx, t_arc: arc, t_wanted: wanted})
     result = sorted(map(tuple, result))
-    expected = sorted([('A', 3, 1),
-                       ('A', 2, 1),
-                       ('A', 5, 1),
-                       ('F', 1, 6),
-                       ('F', 2, 6),
-                       ('F', 4, 6)])
+    expected = sorted([('A', 1, 3),
+                       ('A', 1, 2),
+                       ('A', 1, 5),
+                       ('F', 6, 1),
+                       ('F', 6, 2),
+                       ('F', 6, 4)])
 
     assert result == expected
 
--- a/blaze/compute/tests/test_python_compute.py
+++ b/blaze/compute/tests/test_python_compute.py
@@ -363,11 +363,11 @@ def test_graph_double_join():
     result = compute(j, {t_idx: idx, t_arc: arc, t_wanted: wanted})
     result = sorted(map(tuple, result))
     expected = sorted([('A', 3, 1),
-                    ('A', 2, 1),
-                    ('A', 5, 1),
-                    ('F', 1, 6),
-                    ('F', 2, 6),
-                    ('F', 4, 6)])
+                       ('A', 2, 1),
+                       ('A', 5, 1),
+                       ('F', 1, 6),
+                       ('F', 2, 6),
+                       ('F', 4, 6)])
 
     assert result == expected
 
--- a/blaze/expr/collections.py
+++ b/blaze/expr/collections.py
@@ -452,13 +452,13 @@ class Join(Expr):
 
         right_types = keymap(
             dict(zip(on_right, on_left)).get,
-            dict(self.rhs.schema[0].parameters[0]),
+            self.rhs.dshape.measure.dict,
         )
         joined = (
             (name, promote(dt, right_types[name], promote_option=False))
             for n, (name, dt) in enumerate(filter(
                 compose(op.contains(on_left), first),
-                self.lhs.schema[0].parameters[0]
+                self.lhs.dshape.measure.fields,
             ))
         )
 
--- a/blaze/expr/collections.py
+++ b/blaze/expr/collections.py
@@ -398,17 +398,17 @@ class Join(Expr):
 
     @property
     def on_left(self):
-        if isinstance(self._on_left, tuple):
-            return list(self._on_left)
-        else:
-            return self._on_left
+        on_left = self._on_left
+        if isinstance(on_left, tuple):
+            return list(on_left)
+        return on_left
 
     @property
     def on_right(self):
-        if isinstance(self._on_right, tuple):
-            return list(self._on_right)
-        else:
-            return self._on_right
+        on_right = self._on_right
+        if isinstance(on_right, tuple):
+            return list(on_right)
+        return on_right
 
     @property
     def schema(self):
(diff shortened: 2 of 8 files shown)


blaze/expr/collections.py
```diff
@@ -435,22 +435,31 @@ def schema(self):
         """
         option = lambda dt: dt if isinstance(dt, Option) else Option(dt)
 
-        joined = [[name, dt] for name, dt in self.lhs.schema[0].parameters[0]
-                  if name in self.on_left]
+        on_left = self.on_left
+        on_right = self.on_right
+
+        right_params = self.rhs.schema[0].parameters[0]
```
cpcloud: you can do this with `self.rhs.measure.fields`, which IMO is a bit more clear

blaze/expr/collections.py
```diff
@@ -435,22 +435,31 @@ def schema(self):
...
+        joined = [
+            [name, extract_option(dt)
+             if isinstance(dt, Option) and
+             not isinstance(right_params[n], Option) else dt]
```
cpcloud: should this be `right_params[n][1]`?

blaze/expr/collections.py
```diff
@@ -435,22 +435,31 @@ def schema(self):
...
+            for n, (name, dt) in enumerate(self.lhs.schema[0].parameters[0])
```
cpcloud: same here, I think this should be `self.lhs.measure.fields`

blaze/expr/collections.py
llllllllll: yeah, I am changing this up a bit to use: https://github.com/blaze/datashape/pull/172

blaze/expr/collections.py
llllllllll: there doesn't always appear to be a measure attribute on this type.

cpcloud: hm weird, is that the cause of the failure?

llllllllll: not sure, I just installed pyspark to run this locally; however, I am running into new issues when trying to extend this to handle the join of int32 and int64.

llllllllll: I can ping you when I get something working.

cpcloud: Ok. I'd install spark 1.3 for now, because I haven't worked up a PR to add support for 1.4 yet.

llllllllll:
The build is going to fail until conda gets updated with the datashape pr. Getting 1 error locally where the order of columns are flipped. `test_graph_double_join` in the python compute tests. Not sure what is causing this. If you have a chance, some extra eyes would be apreciated @cpcloud.

llllllllll:
Should be ready to go now.

blaze/compute/tests/test_python_compute.py
```diff
@@ -358,16 +358,28 @@ def test_graph_double_join():
...
-                    ('A', 5, 1),
-                    ('F', 1, 6),
-                    ('F', 2, 6),
-                    ('F', 4, 6)])
+    expected = sorted([('A', 1, 3),
+                       ('A', 1, 2),
+                       ('A', 1, 5),
+                       ('F', 6, 1),
+                       ('F', 6, 2),
+                       ('F', 6, 4)])
```
cpcloud: was this just wrong before?

cpcloud: ah i see your comment

blaze/expr/collections.py
```diff
@@ -435,35 +442,57 @@ def schema(self):
...
+        if not isinstance(on_right, list):
+            on_right = on_right,
 
-        right = [[name, dt] for name, dt in
-                 zip(self.rhs.fields, types_of_fields(
-                     self.rhs.fields, self.rhs))
-                 if name not in self.on_right]
+        right_types = keymap(
+            dict(zip(on_right, on_left)).get,
+            dict(self.rhs.schema[0].parameters[0]),
```
cpcloud: was it not possible to use `self.rhs.measure` here?

llllllllll: no, something didn't have a measure attribute, not sure what the types were here

cpcloud: just tried this, you can do `self.rhs.dshape.measure.dict` and remove the `dict()` call around it

blaze/expr/collections.py
```diff
@@ -435,35 +442,57 @@ def schema(self):
...
+        )
+        joined = (
+            (name, promote(dt, right_types[name], promote_option=False))
+            for n, (name, dt) in enumerate(filter(
+                compose(op.contains(on_left), first),
+                self.lhs.schema[0].parameters[0]
```
cpcloud: here you can do `self.lhs.dshape.measure.fields`

cpcloud: these are longer names, but IMO much more readable

blaze/compute/tests/test_python_compute.py
```diff
@@ -358,16 +358,28 @@ def test_graph_double_join():
...
+    # [[1, A, 3],
+    #  [1, A, 2],
+    #  [1, A, 5],
+    #  [3, C, 1],
+    #  [3, C, 2],
+    #  [3, C, 4],
+    #  [3, C, 5],
+    #  [6, F, 1],
+    #  [6, F, 2],
+    #  [6, F, 4]]
```
cpcloud: why is the joined key all the way over to the left? is that just how the join dshape pops out?

llllllllll: when we construct the pairs we emit `joined + left + right`

llllllllll: Also, I the shape of the measure in the join puts the joined keys first

blaze/compute/tests/test_python_compute.py
cpcloud: where is the line that fixes the column ordering here?

llllllllll: I actually did not intend on fixing this, it sort of fell out of the other changes. That is why I walked through the join manually to assure myself I still had the correct answer. I _think_ what fixed this were the checks on collections.py:446 and 450. This is because we were doing an inclusion check on a field name against a string and 'a' is in 'name' but should not have been selected.

cpcloud: Ah makes sense

blaze/expr/collections.py
```diff
@@ -511,10 +540,28 @@ def join(lhs, rhs, on_left=None, on_right=None,
...
-    if types_of_fields(on_left, lhs) != types_of_fields(on_right, rhs):
-        raise TypeError("Schema's of joining columns do not match")
+        raise ValueError(
+            "Can not Join.  No shared columns between %s and %s" % (lhs, rhs),
+        )
+    left_types = listpack(types_of_fields(on_left, lhs))
+    right_types = listpack(types_of_fields(on_right, rhs))
+    if len(left_types) != len(right_types):
+        raise ValueError(
+            'Length of on_left=%d not equal to lenght of on_right=%d' % (
```
cpcloud: small typo `lenght` should be `length`

PR #1224, opened 2015-08-28
allow sql functions in expressions

llllllllll:


blaze/compute/sql.py
```diff
@@ -871,7 +871,12 @@ def compute_up(expr, data, **kwargs):
 @dispatch(DateTime, (ClauseElement, sa.sql.elements.ColumnElement))
 def compute_up(expr, data, **kwargs):
     if expr.attr == 'date':
-        return sa.func.date(data).label(expr._name)
+        col, = get_inner_columns(data)
+        d = sa.func.date(col).label(expr._name)
+        # Manually reassign the bind, cannot pass to function:
+        # https://bitbucket.org/zzzeek/sqlalchemy/issues/3519
+        d.bind = data.bind
```
cpcloud: is this going to be a problem with computing on any function, i.e., not just calling `date` on a function?

cpcloud: if so, then i think we should do this in `post_compute`

llllllllll: In the `post_compute`, how would we get the original bind? I think the issue is with the label, maybe I should scan for usages of label

cpcloud: Pulling down now to play with this

llllllllll: thanks

cpcloud: I don't think you're testing the issue here. The test passes with current master.

llllllllll: You're right, I actually need to execute this. I will move to the pg tests.

llllllllll:
Pushed a test that fails on master. This issue propagetes to any function base though. I can now do now().date but I cannot do now().date + timedelta or something. We need some solution for this in the general case. For now, until we hear about: https://bitbucket.org/zzzeek/sqlalchemy/issues/3519 I can just make a helper function called label that does the bind copying for us.

blaze/compute/sql.py
```diff
@@ -79,25 +80,77 @@ def inner_columns(s):
...
+    Returns
+    -------
+    dispatcher : callable
+        A decorator that will both wrap the function in ``ensure_types``
+        and the normal ``blaze.dispatch.dispatch``.
+    """
+    return toolz.compose(dispatch(*args), ensure_bind)
+
+
+@dispatch_and_bind(Projection, Selectable)
```
cpcloud: any reason not to do this as two separate decorators like this:

```python
@bound
@dispatch(Expr, Data)
    pass
```

llllllllll: I was just trying to prevent people from forgetting to use this new decorator in the future. `binop_sql` uses the two decorator aproach. If you think this is just more confusing, I can change it.

llllllllll:
@cpcloud I seperated the decorators, good to merge?

blaze/compute/tests/test_postgresql_compute.py
```diff
@@ -176,3 +179,8 @@ def test_distinct_on(sql):
     FROM {tbl}) AS anon_1 ORDER BY anon_1."A" ASC
     """.format(tbl=sql.name))
     assert odo(computation, tuple) == (('a', 1), ('b', 2))
+
+
+def test_function_in_expr():
+    s = symbol('s', 'int32')
+    assert odo(compute(s + 1, sa.func.max(1, bind=resource(db_url))), int) == 2
```
cpcloud: I'm hesitant to merge this because it seems like we're working around an odo issue in blaze. For example, the following should be okay for blaze:

```python
s = symbol('s', 'int32')
sqla = compute(s + 1, sa.func.max(1))
odo(sqla, int)  # this fails, but that's because there's no connection
```

Should we instead do something like `odo(sqla, int, bind=engine)`?

llllllllll: This is already the case with tables and selects. I think having the bind as an argument to odo is a good idea; however, if the bind is part of the data, it should be used.

llllllllll:
closing in favor of: https://github.com/blaze/odo/pull/305

PR #1231, opened 2015-09-02
Update rosetta-pandas.rst

scls19fr:
According https://github.com/pydata/pandas/issues/10654
read from SQL database is now easier with Pandas 0.17 than it was previously

Instead of

```
import sqlalchemy as sa
engine = sa.create_engine('sqlite://db.db')
df = pd.read_sql('select * from t',
                 con=engine)
```

we can do

```
df = pd.read_sql('select * from t',
    con='sqlite://db.db')
```

### DIFF
--- a/docs/source/rosetta-pandas.rst
+++ b/docs/source/rosetta-pandas.rst
@@ -107,8 +107,6 @@ Blaze can simplify and make more readable some common IO tasks that one would wa
 +-----------------+-----------------------------------------------------------------+---------------------------------------------------+
 | Read from       | .. code-block:: python                                          | .. code-block:: python                            |
 | SQL database    |                                                                 |                                                   |
-|                 |    import sqlalchemy as sa                                      |    df = Data('sqlite://db.db::t')                 |
-|                 |    engine = sa.create_engine('sqlite://db.db')                  |                                                   |
-|                 |    df = pd.read_sql('select * from t',                          |                                                   |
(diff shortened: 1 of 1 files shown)


cpcloud:
@scls19fr This is nice thanks for doing this! One concern I have is that not everyone is going to be on the latest pandas. Can you keep the original rosetta for this example and add the new way below it with a tiny note saying something like "only available in pandas 0.17"?

scls19fr:
There is no version information about `engine` use with Pandas (and that's quite new too) ! (a raw database connection was needed until recently)

cpcloud:
> There is no version information about engine use with Pandas

You say above in your comment that in version 0.17.0 you can use engine in this new way. I'm not sure I understand what you mean by "there's no version information".

scls19fr:
I mean that what was written was also wrong for old Pandas version (I don't know
exactly what version of Pandas was able to accept engine and don't need a raw database connection)

cpcloud:
> I mean that what was written was also wrong for old Pandas version:

In pandas 0.16.2

```python
In [2]: pd.read_sql('select * from t', con=sa.create_engine('postgresql://localhost'))
Out[2]:
   a
0  1
1  2
2  3
```

We don't need to specify every single possible way of doing this, including all past ways. All I'm asking is to say "In version 0.17.0 `read_sql` accepts a string for the `con` parameter"

PR #1233, opened 2015-09-04
Add test for join type promotion in postgresql

cpcloud:


llllllllll:
Did you run into the issue where using url as a fixture caused both to get the same url?

cpcloud:
yes i think i'm misunderstanding how the fixture is being executed

llllllllll:
I looked into this, I think there is no scope that means "each time this is called" so each fixture is cached per test

cpcloud:
ok is there a pytest doc for this

llllllllll:
This doesn't really explicitly state it, but this is what I was looking at: https://pytest.org/latest/fixture.html#fixtures-as-function-arguments

cpcloud:
ok thanks. i'm going to change the url stuff around a bit here before merging

PR #1238, opened 2015-09-09
Fixed dead links that went to parent path

alexmojaki:


cpcloud:
@alexmojaki Thanks for the fix

Issue #1245, opened 2015-09-13
memoize dshape 


llllllllll:
I had a pretty large expression which I wanted to construct a predicate to select from.
It took 0.5 seconds to just construct the actual expression (without any computation), and running under cProfile showed that the time was dominated by access of the dshape/schema attributes. I think that we should memoize these properties because the expressions are immutable. If the expressions are weakreferencable, we could even slip an `@memoize(cache=WeakKeyDictionary())` between the getter and the property decorator

cpcloud:
@llllllllll can this be closed?

cpcloud:
closing as this was fixed already

PR #1257, opened 2015-09-25
Memoize `Join.schema` to improve expression performance.

kwmsmith:
@cpcloud You were right, `memoizing` `Join.schema` was the ticket.  Performance is much better now.

### DIFF
--- a/blaze/expr/collections.py
+++ b/blaze/expr/collections.py
@@ -10,7 +10,7 @@ from datashape import (
 from datashape.predicates import isscalar, iscollection, isrecord
 from toolz import (
     isdistinct, frequencies, concat as tconcat, unique, get, first, compose,
-    keymap,
+    keymap, memoize
 )
 import toolz.curried.operator as op
 from odo.utils import copydoc
@@ -418,6 +418,7 @@ class Join(Expr):
         return on_right
 
     @property
+    @memoize
     def schema(self):
         """
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ install:
   - conda update conda
 
   # Install dependencies
-  - conda create -n blaze python=$TRAVIS_PYTHON_VERSION pytest numpy sqlalchemy pandas h5py pip flask requests pytables cython bcolz xlrd coverage psutil networkx numba
+  - conda create -n blaze python=$TRAVIS_PYTHON_VERSION pytest numpy sqlalchemy pandas h5py pip flask requests pytables cython bcolz xlrd coverage psutil networkx numba pyyaml
   - if [ -n "$PANDAS_VERSION" ]; then conda remove -n blaze pandas; fi
   - source activate blaze
 


blaze/expr/collections.py
```diff
@@ -418,6 +418,7 @@ def on_right(self):
         return on_right
 
     @property
+    @memoize
     def schema(self):
```
llllllllll: Instead of memoizing a property, could we just make schema a descriptor that swaps itself out on the instance?

llllllllll:
related:  #1245

blaze/expr/collections.py
cpcloud: > could we just make schema a descriptor that swaps itself out on the instance

@llllllllll can you give an example of this? i don't quite understand what you mean here.

llllllllll:
```python
class Join:
    ...
    class schema:
        def __get__(self, instance, owner):
            if instance is None:
                return self
            ret = <compute schema>
            instance.schema = schema
            return ret
    schema = schema()
```
becase we don't have a `__set__` or `__delete__` on `schema`, it is a non-data descriptor and once we have assigned a value to the instance, that will be looked up first. The class does not need to be defined inline of course, we could generalize this into some `lazyval` class.

llllllllll: Actually, probably more clear is to put a `__weakref__` in the `__slots__` and then just use a `WeakKeyDictionary` for the cache of the `memoize`. This is what I have done elsewhere to prevent this leak.

cpcloud: Sounds good to me.

cpcloud:
@kwmsmith i'm going to rip out 082d96a and merge it directly into master

blaze/expr/collections.py
kwmsmith: @llllllllll sorry for the delay getting back to you -- yes, agreed on the descriptor for `schema`.  Thanks for the example code, my attempt at this consistently ran into a recursion limit exception.  I'll put in your version and test it out.

kwmsmith:
@cpcloud +1.

blaze/expr/collections.py
kwmsmith: @llllllllll adding `__weakref__` to `Join.__slots__` led to this cryptic error:

```
blaze/__init__.py:19: in <module>
    from .expr import (Symbol, TableSymbol, symbol, ndim, shape)
blaze/expr/__init__.py:11: in <module>
    from .collections import *
blaze/expr/collections.py:359: in <module>
    class Join(Expr):
E   TypeError: Error when calling the metaclass bases
E       __weakref__ slot disallowed: either we already got one, or __itemsize__ != 0
```

I'm also curious about the `__slots__` usage in general in `collections.py`, see #1268.

llllllllll: I think that you are correct about the slots not being used. The object was already weakreferencable from the super class's `__dict__`. We can probably just remove the slots or add slots to the super class.

kwmsmith:
Closing -- see #1269.

blaze/expr/collections.py
cpcloud: > adding __weakref__ to Join.__slots__ led to this cryptic error:

Why not just add `__slots__ = ()` to `Node` and `Expr`? That'll prevent that error from occurring.

cpcloud:
@kwmsmith when you close a PR can you remove the branch on upstream so that we don't have a ton of random branches floating around upstream? thx!

cpcloud:
I'd do it myself, but I'm not sure if you want to keep it around on your local machine and I don't want to prevent you from doing that.

kwmsmith:
> @kwmsmith when you close a PR can you remove the branch on upstream so that we don't have a ton of random branches floating around upstream? thx!

Yes, will do.

blaze/expr/collections.py
kwmsmith: @cpcloud I tried adding `__slots__ = ()` to `Node` and `Expr` along with weak referenceable `Join.__slots__` and `@memoize(cache=WeakKeyDictionary())` but I get recursion errors.  I've seen this with other attempts, and makes me want to refactor a lot of the machinery in the `Node` and `Expr` hierarchy.

The reason for the recursion errors is that `Expr.__getattr__` looks up `self.dshape`, the `Join.dshape` property depends on `Join.schema`, which hits `Expr.__getattr__` again.

So to use weakrefs with memoization here, it requires a change in attribute lookup and `Expr.__getattr__`.

My preference is to do something simpler, like in #1269.

PR #1259, opened 2015-09-27
Add binary math functions and refactor math ops

cpcloud:


cpcloud:
@llllllllll this could use your review

note that i've added a `greatest` and `least` function (inspired by postgres) that are the two argument forms of max and min respectively

llllllllll:
I will be leaving for ny at 5:00 tonight, I will review this on the train.

blaze/compute/numpy.py
```diff
@@ -122,12 +150,12 @@ def compute_up(t, x, **kwargs):
 
 @dispatch(Not, np.ndarray)
 def compute_up(t, x, **kwargs):
-    return ~x
+    return np.logical_not(x)
```
llllllllll: What is the difference between these two? Same with negative

blaze/compute/numpy.py
```diff
@@ -96,6 +97,33 @@ def compute_up_np_interp(t, lhs, rhs, **kwargs):
...
+
+@compute_up.register(least, np.ndarray, (np.ndarray, base))
+@compute_up.register(least, base, np.ndarray)
+def compute_up_least(expr, lhs, rhs, **kwargs):
+    return np.minimum(lhs, rhs)
+
+
+@dispatch(greatest, np.ndarray)
+def compute_up(expr, data, **kwargs):
+    if isinstance(expr.lhs, Expr):
```
llllllllll: We use this pattern a lot, is there some way we can generalize this?

llllllllll:
Greatest behaves strangely for python compared to numpy:

```python
In [69]: bz.greatest(np.array([1, 2]), np.array([2, 1]))
Out[69]: array([2, 2])

In [71]: bz.greatest([1, 2], [2, 1])
Out[71]: [2, 1]
```

I think that the numpy result is the correct version. `builtin.(max|min)` are just sorting the lists themselves.

blaze/compute/sql.py
```diff
@@ -154,11 +155,85 @@ def compute_up(t, data, **kwargs):
...
+
+
+@compute_up.register(BinaryMath, base, Select)
+def binary_math_sql_select(t, lhs, rhs, **kwargs):
+    left, right = lhs, first(rhs.inner_columns)
+    result = getattr(sa.func, type(t).__name__)(left, right)
+    return reconstruct_select([result], rhs)
+
+
+@compute_up.register(BinaryMath, Select, Select)
```
llllllllll: I think we also want to dispatch on `Table`, maybe `FromClause`?

```python
In [25]: ds = bz.Data('postgresql://localhost/yui::test')

In [26]: es = bz.Data('postgresql://localhost/yui::test2')

In [27]: repr(bz.greatest(ds, es))
NotImplementedError                       Traceback (most recent call last)
  ...
NotImplementedError: Don't know how to compute:
expr: greatest(_3, _4).head(11)
data: {_4: Table('test2', MetaData(bind=Engine(postgresql://localhost/yui)), Column('a', INTEGER(), table=<test2>), Column('b', INTEGER(), table=<test2>), schema=None), _3: Table('test', MetaData(bind=Engine(postgresql://localhost/yui)), Column('a', INTEGER(), table=<test>), Column('b', INTEGER(), table=<test>), schema=None)}

In [28]: ds.dshape
Out[28]: dshape("var * {a: ?int32, b: ?int32}")

In [29]: es.dshape
Out[29]: dshape("var * {a: ?int32, b: ?int32}")
```

blaze/compute/tests/test_numpy_compute.py
```diff
@@ -558,3 +563,45 @@ def test_concat_mat():
         compute(concat(s, t, axis=1), {s: s_data, t: t_data}) ==
         np.concatenate((s_data, t_data), axis=1)
     ).all()
+
+
+def test_least():
```
llllllllll: should we have some test cases that are mixed, like:
```python
s_data = np.array([1, 2])
t_data = np.array([[2, 1])
```

blaze/compute/tests/test_postgresql_compute.py
```diff
@@ -212,3 +228,74 @@ def test_join_type_promotion(sqla, sqlb):
     result = set(map(tuple, compute(expr, {t: sqla, s: sqlb}).execute().fetchall()))
     expected = set([(1, 'a', 'a'), (1, None, 'a')])
     assert result == expected
+
+
+def test_dist(nyc):
```
llllllllll: Are these just to validate that some published examples continue to work?

blaze/expr/math.py
```diff
@@ -11,67 +11,166 @@
...
-class acos(RealMath): pass
-class acosh(RealMath): pass
-class asin(RealMath): pass
-class asinh(RealMath): pass
-class atan(RealMath): pass
-class atanh(RealMath): pass
 
-class radians(RealMath): pass
-class degrees(RealMath): pass
+class abs(UnaryMath):
```
llllllllll: Could we programmtically generate these?

```python
for name in unary_type_names:
    locals()[name] = type(name, (UnaryMath,), {})
```
same for the binary names.
This makes it easier to scan through quickly imho.

blaze/expr/functions.py
```diff
@@ -77,6 +80,27 @@ def sqrt(o):
...
 
+binary_types = {
+    builtins: [(object, object)],
+    np: [((np.ndarray, np.number), (np.ndarray, np.number))],
+    pymath: [(Number, Number)],
+    blazemath: [(Expr, (Expr, base)), (base, Expr)]
+}
+
+
+binary_mappings = {
```
llllllllll: Can we rename this `fallback_binary_mapping` or something to imply that this is not the default place to look for the binary function mappings.

blaze/expr/expressions.py
```diff
@@ -713,7 +713,7 @@ def apply(expr, func, dshape, splittable=False):
     return Apply(expr, func, datashape.dshape(dshape), splittable)
 
 
-class Coerce(Expr):
+class Coerce(ElemWise):
```
llllllllll: I like the practicality of this; however, I wonder if it would make more sense to make the syntax for this something like:

```python
arr = bz.Data([1, 2, 3])
asfloat = arr.coerce('var * float64')
```

This might even let us do some fun things with reshaping like:

```python
arr = bz.Data([1, 2, 3, 4])
asmat = arr.coerce('2 * 2 * int32')
```

Just a thought, not sure if this is practical.

blaze/compute/numpy.py
cpcloud: nothing, it just reads more clearly IMO

blaze/compute/tests/test_numpy_compute.py
cpcloud: yep

blaze/compute/tests/test_postgresql_compute.py
cpcloud: kind of, they are tests that are using real data as opposed to hand-constructed data

blaze/expr/expressions.py
cpcloud: This would be a good issue to raise separtely

blaze/compute/sql.py
cpcloud: This doesn't make sense for tables as you can't add them together ... the reason I've dispatched on `Select` is because `Field` expressions on transform operations generate single column select statements

cpcloud: It's not possible to construct a `Table` where it makes sense to define arithmetic on it

blaze/compute/tests/test_postgresql_compute.py
cpcloud: This data also presents a real use case for some of the less-often-seen math functions like `atan2`

cpcloud: I'm going to reduce the size of it as it's slowing down tests unacceptably

cpcloud: actually nevermind the slowdown is from constructing the epxression

blaze/compute/numpy.py
llllllllll: Does our `Not` only work with bools? This gives different results for dtypes other than `bool`

```python
In [2]: ~np.array([1, 2, 3])
Out[2]: array([-2, -3, -4])

In [3]: np.logical_not(np.array([1, 2, 3]))
Out[3]: array([False, False, False], dtype=bool)
```

blaze/compute/sql.py
llllllllll: Okay, just wanted to make sure.

blaze/compute/numpy.py
cpcloud: Right, it was technically incorrect then. Blaze's `Not` operation is only defined for boolean datashaped things

cpcloud: This wasn't showing up because you can't construct an expression that would allow you to call `~` on that expression unless it had a `bool` measure

blaze/compute/numpy.py
cpcloud: Maybe. I'll raise a separate issue

cpcloud: Hm in this case these definitions may not be necessary.

cpcloud:
@llllllllll Merging on pass. Any last comments?

llllllllll:
nope, looks good

PR #1260, opened 2015-09-28
Add test for all string CSV using has_header=True

cpcloud:


cpcloud:
This is just adding a test, merging on pass.

PR #1262, opened 2015-10-05
BUG: don't unpack options in schema_concat

llllllllll:
This was causing some strange dshape stuff in transform:

```python
deltas.dshape dshape("var * {sid: ?int64, asof_date: datetime, timestamp: datetime, value: float64}")
deltas.value.dshape dshape("var * ?float64")
```

In the full expr, `value` is not an option type; however, when we pull the column out, it becomes optional. The issue is that `value` in the base expression is dropping the option

### DIFF
--- a/blaze/expr/collections.py
+++ b/blaze/expr/collections.py
@@ -256,21 +256,17 @@ def schema_concat(exprs):
 
     In the case of Units, the name is taken from expr.name
     """
-    names, values = [], []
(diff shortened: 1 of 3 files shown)


cpcloud:
This was definitely a bug, thanks for fixing!

llllllllll:
Merging on pass

cpcloud:
Roger roger

PR #1266, opened 2015-10-07
Adds a Shift expression

cpcloud:


blaze/expr/collections.py
```diff
@@ -708,8 +710,45 @@ def isin(expr, keys):
...
+    """
+    __slots__ = '_hash', '_child', 'n'
+
+    @property
+    def schema(self):
+        measure = self._child.schema.measure
+
+        # if we are not shifting or we are already an Option type then return
+        # the child's schema
+        if not self.n or isinstance(measure, Option):
```
llllllllll: should a shift(expr, 0) just return expr?

blaze/expr/collections.py
```diff
@@ -708,8 +710,45 @@ def isin(expr, keys):
...
+
+    def __str__(self):
+        return '%s.%s(n=%d)' % (
+            self._child, type(self).__name__.lower(), self.n
+        )
+
+
+@copydoc(Shift)
+def shift(expr, n):
+    if not isinstance(n, (numbers.Integral, np.integer)):
```
llllllllll: is there a reason we cannot allow expressions for n. like, `a.shift(b.count())`?

blaze/expr/collections.py
```diff
@@ -708,8 +710,49 @@ def isin(expr, keys):
     return IsIn(expr, frozenset(keys))
 
 
+class Shift(Expr):
+    """ Shift a column backward or forward by N elements
+
+    Parameters
+    ----------
+    expr : Expr
+        The expression to shift. This expression's dshape should be columnar
```
llllllllll: Should we typecheck in the `shift` constructor?

cpcloud: no, this done by the dshape methods/property checking in `__getattr__`

blaze/expr/collections.py
cpcloud: No, but I'd like to do that in a separate PR

blaze/expr/collections.py
llllllllll: Should we not expose the `shift` toplevel function also?

blaze/expr/collections.py
llllllllll: kk

blaze/expr/collections.py
cpcloud: I'm going to leave it as is for now. If this becomes a bottleneck in the future then it's very easy to change

PR #1269, opened 2015-10-10
Cached `Join` schema

kwmsmith:
Builds on #1257.

Add `_schema` to `Join.__slots__`, fix logic in schema property.

Ensures that the schema is cached per-instance.  Fixes leak that `@memoize` use would introduce.

This implementation works better than creating a custom descriptor or using a weakreferenced `@memoize`, given the complex interaction with other Blaze magic.

This implementation matches the pattern in other `Expr` subclasses, like `blaze.expr.expressions.Apply` and `blaze.expr.expressions.Map`.


kwmsmith:
Apologies for the new PR -- I rebased on master without thinking.

cpcloud:
@kwmsmith merge master back into your branch here as well

kwmsmith:
@cpcloud Thanks, that fixed it.

In the meantime, after trying the different descriptor and `__slots__`-based solutions, I went back to `@memoize`.  The other attempts kept hitting recursion depth exceeded exceptions due to complex property access.

My preference at this point, if you agree, is to merge this PR and make an issue to improve and simplify the overall design of properties for `Node`, `Expr`, `Join`, and related classes.

Thoughts?

llllllllll:
this is just going to leak if we have memoize on a property without using a weakkeydict

kwmsmith:
@llllllllll @cpcloud comments on the latest here?

blaze/expr/collections.py
```diff
@@ -397,7 +398,7 @@ class Join(Expr):
     blaze.expr.collections.Merge
     """
     __slots__ = (
-        '_hash', 'lhs', 'rhs', '_on_left', '_on_right', 'how', 'suffixes',
+        '_hash', 'lhs', 'rhs', '_on_left', '_on_right', 'how', 'suffixes'
```
llllllllll: is `__weakref__` in the superclass's slots? otherwise I am not sure how this is weakreferencable

llllllllll: or a `__dict__` maybe?

PR #1272, opened 2015-10-12
Fix scalar_coerce for nan->nat

llllllllll:
This bug also prevented me from taking the min of an empty series of `datetime64` in pandas when using the blaze server because the server was trying to scaler_coerce the `nan` into a `datetime.datetime`.

### DIFF
--- a/blaze/interactive.py
+++ b/blaze/interactive.py
@@ -14,7 +14,7 @@ import numpy as np
 from odo import resource, odo
 from odo.utils import ignoring, copydoc
 from odo.compatibility import unicode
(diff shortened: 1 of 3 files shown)


jreback:
you should conditionally check results based on the pandas version

cpcloud:
> you should conditionally check results based on the pandas version

what's the reason for doing this?

llllllllll:
Should we be checking the output because of the reduction fix that was just merged? I still think that it is reasonable to coerce an explicitly passed `nan` just like `pd.Timestamp` does.

PR #1273, opened 2015-10-12
BUG: fix len for interactive exprs that wrap sql

llllllllll:
`compute` for the sqlbackend would return a `select`. the `len` machinery would then try to coerce this into an `int` and fail with a type error.

### DIFF
--- a/blaze/compute/tests/test_postgresql_compute.py
+++ b/blaze/compute/tests/test_postgresql_compute.py
@@ -14,7 +14,7 @@ import pandas.util.testing as tm
 
 from odo import odo, resource, drop, discover
 from blaze import symbol, compute, concat, join, sin, cos, radians, atan2
-from blaze import sqrt, transform
+from blaze import sqrt, transform, Data
 from blaze.utils import example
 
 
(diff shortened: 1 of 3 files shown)


blaze/compute/tests/test_postgresql_compute.py
```diff
@@ -212,3 +212,8 @@ def test_join_type_promotion(sqla, sqlb):
     result = set(map(tuple, compute(expr, {t: sqla, s: sqlb}).execute().fetchall()))
     expected = set([(1, 'a', 'a'), (1, None, 'a')])
     assert result == expected
+
+
+def test_interactive_len(sql):
+    t = Data(sql)
+    assert len(t) == odo(t.count(), int)
```
cpcloud: You can actually just call `int(t.count())`, but up to you

llllllllll: oh, I didn't realize `__int__` was the same. I like the clarity of `odo` here; however, I don't mind promoting this pattern.

cpcloud:
@llllllllll Merge when ready

llllllllll:
I only needed to resolve the whatsnew but I rebased so I will let this pass before merging.

cpcloud:
cool

Issue #1268, opened 2015-10-10
Join.__slots__ : meaningless, according to Python documentation?


kwmsmith:
From [the fine manual](https://docs.python.org/3.4/reference/datamodel.html#notes-on-using-slots):

"When inheriting from a class without `__slots__`, the `__dict__` attribute of that class will always be accessible, so a `__slots__` definition in the subclass is meaningless."

Looking at `Join`--which defines `__slots__`--it is a subclass of `Expr` and `Node`, neither of which define `__slots__`.  From my reading, this makes `Join.__slots__` "meaningless."

What is the intent of the usage of `__slots__` in the classes in `blaze.expr.collections`?  Space savings?  An easy way to restrict attributes?

cpcloud:
> "When inheriting from a class without `__slots__`, the `__dict__` attribute of that class will always be accessible, so a `__slots__` definition in the subclass is meaningless."

Unless you're [using `__slots__` in the superclass](https://github.com/blaze/blaze/blob/master/blaze/expr/core.py#L71-L86)

> What is the intent of the usage of `__slots__` in the classes in `blaze.expr.collections`? Space savings? An easy way to restrict attributes?

Both. Another reason is simplification of initialization. Notice that many blaze expressions have no explicit constructor defined: these instances are constructed via the `Node` constructor, which uses slots.

mrocklin:
This decision came out of experience with sympy which had non-trivial logic within constructors.  This ended up making a lot of transformations of the expression tree difficult because you were always running into complexity in the constructors.  

With blaze.expr the goal was to have expression classes be restricted and dumb.  Slots were used mostly as an explicit list of fields.

kwmsmith:
> Unless you're using __slots__ in the superclass

I saw this, but is it doing what's expected?  The docs say that "the default can be overridden by defining __slots__ in a class definition."  Nowhere do `Node` or `Expr` define `__slots__` in their definitions, so any subclasses would have a "meaningless" `__slots__`, AFAICT.  It seems like *using* a subclass' `__slots__` isn't sufficient, it has to be defined at the class level for the entire hierarchy.

kwmsmith:
> Another reason is simplification of initialization. Notice that many blaze expressions have no explicit constructor defined: these instances are constructed via the Node constructor, which uses slots.

This makes sense.

cpcloud:
I saw this, but is it doing what's expected?

```python
In [1]: class Foo(object):
   ...:     __slots__ = ()
[...]

PR #1274, opened 2015-10-12
Add empty __slots__ to Expr and Node classes.

kwmsmith:
See issue #1268.

Consistently uses `__slots__` in class hierarchies, even when empty.


cpcloud:
Hm it'd be great to have a test. Maybe something as simple as

```
assert Expr.__slots__ == ()
assert Node.__slots__ == ()
```

cpcloud:
When you're ready to merge please also add a tiny release note blurb to `docs/source/whatsnew/0.5.0.txt`

kwmsmith:
I agree a test would be good, but it feels strange to test an internal detail like "does this class have an empty `__slots__`."  It feels like testing the API for a private method.  I don't know of any better way to test for this, though, so I'll add the asserts you mention.

cpcloud:
Yes that test is definitely a bit awkward. If there's a way to test it using the user facing API, by all means

kwmsmith:
> When you're ready to merge please also add a tiny release note blurb to docs/source/whatsnew/0.5.0.txt

Should that be `0.8.4.txt`?  The latest tag I see is `0.8.3`.

cpcloud:
Sorry I meant 0.9.0. I've moved the version to a new minor version since we're dropping support for things.

cpcloud:
@kwmsmith merge master back into your branch it should be passing now

Issue #1230, opened 2015-09-01
selection fails but predicate works


llllllllll:
```python
In [1]: dt
Out[1]: datetime.datetime(2015, 9, 1, 17, 30, 53, 819315)

In [2]: ds = bz.Data(dt)

In [3]: es = bz.Data(pd.DataFrame([[pd.Timestamp.now(), 1], [pd.Timestamp.now(), 2]], columns=['time', 'value']))

In [4]: es.time.date > ds.date - datetime.timedelta(days=1)
Out[4]: 
       
0  True
1  True

In [5]: es[es.time.date > ds.date - datetime.timedelta(days=1)]
Out[5]: ---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/home/yui/.virtualenvs/databazaar/lib/python3.4/site-packages/IPython/core/formatters.py in __call__(self, obj)
    695                 type_pprinters=self.type_printers,
    696                 deferred_pprinters=self.deferred_printers)
--> 697             printer.pretty(obj)
    698             printer.flush()
    699             return stream.getvalue()

/home/yui/.virtualenvs/databazaar/lib/python3.4/site-packages/IPython/lib/pretty.py in pretty(self, obj)
    381                             if callable(meth):
    382                                 return meth(obj, self, cycle)
--> 383             return _default_pprint(obj, self, cycle)
    384         finally:
    385             self.end_group()

/home/yui/.virtualenvs/databazaar/lib/python3.4/site-packages/IPython/lib/pretty.py in _default_pprint(obj, p, cycle)
    501     if _safe_getattr(klass, '__repr__', None) not in _baseclass_reprs:
    502         # A user-provided repr. Find newlines and replace them with p.break_()
--> 503         _repr_pprint(obj, p, cycle)
    504         return
    505     p.begin_group(1, '<')

/home/yui/.virtualenvs/databazaar/lib/python3.4/site-packages/IPython/lib/pretty.py in _repr_pprint(obj, p, cycle)
    683     """A pprint that just redirects to the normal repr function."""
    684     # Find newlines and replace them with p.break_()
--> 685     output = repr(obj)
    686     for idx,output_line in enumerate(output.splitlines()):
    687         if idx:

/home/yui/quantopian/blaze/blaze/interactive.py in expr_repr(expr, n)
    259     if (ndim(expr) == 1 and (istabular(expr.dshape) or
    260                              isscalar(expr.dshape.measure))):
--> 261         return repr_tables(expr, 10)
[...]

PR #1275, opened 2015-10-14
Fix selections that have inner inputs.

llllllllll:
Things like `a[a.a == b.a]` would fail because it wouldn't compute the `b.a.`

### DIFF
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -154,9 +154,24 @@ def compute_up(t, data, **kwargs):
         return t.op(t.lhs, column)
 
 
-@compute_up.register(BinOp, (ColumnElement, base), ColumnElement)
-@compute_up.register(BinOp, ColumnElement, base)
+@compute_up.register(
+    BinOp, (Select, ColumnElement, base), (Select, ColumnElement),
+)
+@compute_up.register(BinOp, (Select, ColumnElement), base)
 def binop_sql(t, lhs, rhs, **kwargs):
+    if isinstance(lhs, Select):
+        assert len(lhs.c) == 1, (
+            'Select cannot have more than a single column when doing'
+            ' arithmetic, got %s' % lhs
+        )
+        lhs = first(lhs.inner_columns)
+    if isinstance(rhs, Select):
+        assert len(rhs.c) == 1, (
+            'Select cannot have more than a single column when doing'
+            ' arithmetic, got %s' % rhs
+        )
+        rhs = first(rhs.inner_columns)
+
     return t.op(lhs, rhs)
 
 
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -280,7 +280,7 @@ def compute_up(expr, data, scope=None, **kwargs):
     return compute_up(expr, data, predicate, scope=scope, **kwargs)
 
 
-@dispatch(Selection, sa.sql.ColumnElement, ClauseElement)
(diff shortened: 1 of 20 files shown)


blaze/compute/sql.py
```diff
@@ -262,19 +262,41 @@ def compute_up(t, s, **kwargs):
 @dispatch(Selection, sa.sql.ColumnElement)
 def compute_up(expr, data, scope=None, **kwargs):
     predicate = compute(expr.predicate, data, post_compute=False)
-    return sa.select([data]).where(predicate)
+    return compute_up(expr, data, predicate, scope=scope, **kwargs)
+
+
+@dispatch(Selection, sa.sql.ColumnElement, ClauseElement)
```
cpcloud: Should `ClauseElement` be `sa.sql.expression.BinaryExpression`?

cpcloud: `ClauseElement` is very general and would let basically any sqlalchemy expression in.

blaze/expr/broadcast.py
```diff
@@ -125,15 +127,20 @@ def broadcast_collect(expr, Broadcastable=Broadcastable,
...
+
     # Recurse down
-    children = [broadcast_collect(i, Broadcastable, WantToBroadcast)
-                for i in expr._inputs]
-    return expr._subs(dict(zip(expr._inputs, children)))
+    children = (
+        broadcast_collect(i, broadcastable, want_to_broadcast, no_recurse)
+        for i in expr._inputs
+    )
+    return expr._subs({e: c for e, c in zip(expr._inputs, children)})
```
cpcloud: You couldn't resist :D

blaze/compute/sql.py
llllllllll: Is `BinaryExpression` too restrictive here? Are there any valid predicate functions that don't fall into this? Is there any `ClauseElement` that we would not treat as a valid predicate?

blaze/expr/broadcast.py
llllllllll: yeah, I saw my chance to take advantage of dropping 2.6 and I went for it.

llllllllll:
looks like I broke everything. I will clean the other backends up and ping you when its passing.

llllllllll:
related #1230

I need to get back to this. I haven't had much time to work on blaze recently ;_;

llllllllll:
Hoping the tests pass. This should be ready to review so I am removing the wip tag.

llllllllll:
@cpcloud any thoughts on this? Currently using this for our deploy so it would be good to get this merged.

cpcloud:
I'll review this today n

blaze/compute/sql.py
cpcloud: Yes, `Selectable` is a subclass of `ClauseElement` and we don't support non-column boolean expressions.

llllllllll: I tried making this `BinaryExpression` but it looks like there are some cases where we are getting a `BooleanClauseList`. The first common parrent is `ColumnElement`, does that seem restrictive enough?

cpcloud: Yeah, `ColumnElement` is fine

llllllllll:
Not sure if the tests are still failing with a 500 from binstar, any idea what caused (or is still causing) this?

cpcloud:
+1 here

llllllllll:
> +1

good to merge?

PR #1280, opened 2015-10-24
Use aliases instead of common table expressions in the SQL backend

### LINKED ISSUE #1278: "WITH alias as (...) SELECT" statements are generated for Mysql while mysql doesn't support them
Example
```python
db = blaze.Data('mysql://...')
tb = db['my_table'][['field1', 'time', 'column']]

sel = tb[tb.field1 == 661]

sel.time.min()
```

Generated SQL:

```sql
WITH alias_25 AS 
(SELECT my_table.time AS time 
FROM my_table
WHERE my_table.field1 = %s)
 SELECT min(alias_25.time) AS time_min 
FROM alias_25
```

Error:

    ProgrammingError: (_mysql_exceptions.ProgrammingError) (1064, "You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'alias_25 AS \n(SELECT my_table.time AS time \nFROM my_table' at line 1") [SQL: u'WITH alias_25 AS \n(SELECT my_table.time AS time \nFROM my_table \nWHERE my_table.field1 = %s)\n SELECT min(alias_25.time) AS time_min \nFROM alias_25'] [parameters: (661,)]

cpcloud:
Fixes #1278

### DIFF
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -457,7 +457,7 @@ def reconstruct_select(columns, original, **kwargs):
 def compute_up(expr, data, **kwargs):
     if expr.axis != (0,):
         raise ValueError('axis not equal to 0 not defined for SQL reductions')
-    data = data.cte(name=next(aliases))
+    data = data.alias(name=next(aliases))
     cols = list(inner_columns(data))
     d = dict((expr._child[c], cols[i])
              for i, c in enumerate(expr._child.fields))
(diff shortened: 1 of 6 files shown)


cpcloud:
@llllllllll do you depend on CTEs anywhere?

llllllllll:
Nope, we are not using the intermediate results of the sql backend currently, just pure blaze.

cpcloud:
sweet merging

PR #1289, opened 2015-11-12
ENH: Adds 'tail' to sql backend

llllllllll:
Tail can be implemented efficiently in sql by flipping the sort order, taking a limit, and then resorting the results.


cpcloud:
@llllllllll can you fix the merge conflicts, then good to go

PR #1308, opened 2015-11-19
Fix empty string comparisons on option types

cpcloud:


cpcloud:
@llllllllll @kwmsmith this PR breaks the following test:

```
============================================================ FAILURES ============================================================
_______________________________________________________ test_scalar_coerce _______________________________________________________
blaze/expr/tests/test_scalar.py:258: in test_scalar_coerce
    assert scalar_coerce('?date', '') == None
E   assert datetime.date(2015, 11, 19) == None
E    +  where datetime.date(2015, 11, 19) = scalar_coerce('?date', '')
```

I think that a comparison such as `t.date_column == ''` should raise a `TypeError`. Thoughts?

cpcloud:
Right now, `t.date_column == ''` would be equivalent to `t.date_column == None`

llllllllll:
I think this is correct behavior. I was actually suprised that comparisons with strings coerced to datetimes at all. I am onboard with changing the test and deprecating that behavior

PR #1317, opened 2015-11-20
Add some of the lowhanging fruit from dask.dataframe

cowlicks:


blaze/compute/pandas.py
```diff
@@ -36,6 +36,13 @@
 from datashape.predicates import isscalar
 
 from odo import into
+try:
+    import dask.dataframe as dd
+    DaskDataFrame = dd.DataFrame
+    DaskSeries = dd.Series
+except ImportError:
+    DaskDataFrame = pd.DataFrame
```
llllllllll: This might cause ambiguity with compute_up in the pandas backend.

blaze/compute/pandas.py
```diff
@@ -523,11 +533,14 @@ def compute_up(t, df, scope=None, **kwargs):
     return pd.concat(children, axis=1)
 
 
-@dispatch(Summary, DataFrame)
+@dispatch(Summary, (DataFrame, DaskDataFrame))
 def compute_up(expr, data, **kwargs):
     values = [compute(val, {expr._child: data}) for val in expr.values]
     if expr.keepdims:
-        return DataFrame([values], columns=expr.fields)
+        if isinstance(data, Dataframe):
```
llllllllll: this could be `type(data)(...)`

cpcloud: this is also a typo `Dataframe` should be `DataFrame` ... i think `type(self)(...)` would solve the problem

cowlicks: Thanks, that is much better.

cpcloud:
looks like dask dataframes are missing a shape attribute ... does that even make sense for that object?

mrocklin:
No, dask.dataframes do not know their own length, except through computation.

cpcloud:
@cowlicks You'll probably have to implement a new dispatched version of those expressions.

cpcloud:
@cowlicks You could probably do `nelements` with `series.isnull().sum() + series.count()` and `count` should probably be `series.count()`

cowlicks:
@mrocklin dask.dataframes do have a len method, which silently does computation because it is usually cheap.

@cpcloud  hows this? https://github.com/blaze/blaze/commit/238c963041ece51df5211b0d417fb354864c5f21

mrocklin:
The cost of `len(df)` depends strongly on the graph that creates the dataframe.  If it is based off of an in-memory pandas dataframe then yes, it is cheap.  However, if it is based off of something more expensive, like a CSV file, then it can be quite expensive to compute.

cowlicks:
@mrocklin I'll leave it out then.

cpcloud:
I'd be okay with leaving out `nelements` but we can't leave out count

cpcloud:
@cowlicks On second thought, I still think `nelements` should work. I don't think we should fail because it may take a long time to compute. A user is going to need to have *some* knowledge of the thing they are computing on to be able to predict performance. A warning would suffice I think

cowlicks:
@cpcloud any idea why [this call to count](https://github.com/blaze/blaze/pull/1317/files#diff-4119ff549f68024d9c2bfca5ddf22e78R112) does not dispatch to [here](https://github.com/blaze/blaze/pull/1317/files#diff-b3b902a31bfadd3c970b54cb4d4c9350R612)?

cpcloud:
because the test is operating on a `Field` and that implementation is for `count` on an entire `DataFrame`

cowlicks:
This is not comprehensive, but includes a lot of easy stuff from dask dataframe. I believe the tests cover all the added `compute_up` methods.

cpcloud:
@cowlicks can you add a release note in `docs/source/whatsnew/0.9.0.txt`? then good to merge

cowlicks:
Added release notes. Merging soon unless there are further comments.

PR #1321, opened 2015-11-23
Pandas: sort(columns=) is deprecated, use sort_values(by=)

pratapvardhan:
`Pandas 0.17.0` gives a warning for the use of `sort`

```FutureWarning: sort(columns=....) is deprecated, use sort_values(by=.....)```

Probably move to sort_values, as suggested?

### DIFF
--- a/blaze/compatibility.py
+++ b/blaze/compatibility.py
@@ -2,11 +2,8 @@ from __future__ import absolute_import, division, print_function
 
 import sys
 from types import MethodType
-from distutils.version import LooseVersion
 
 import pandas.util.testing as tm
-
-import pandas as pd
 
 PY3 = sys.version_info[0] == 3
 PY2 = sys.version_info[0] == 2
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by cpcloud** [comment]
We want to maintain compatibility with a few versions of pandas so we can't change to `sort_values` just yet.

**Comment by cpcloud** [comment]
If you want to wrap up sort in an internal function that would be okay:

```python
def sort(df, *args, **kwargs):
    try:
        return df.sort_values(*args, **kwargs)
    except AttributeError:
        return df.sort(*args, **kwargs)
```

jreback:
@cpcloud might be safer to do actual version detection (as you need to *maybe* handle the ``by`` kw)

e.g.
```
if LooseVersion(pd.__version__) < '0.17.0':
     return df.sort(*args, **kwargs)
return df.sort_values(...)
```

llllllllll:
ideally we would only do this check once at module scope and then conditionally define some `sort` function.

cpcloud:
@llllllllll that seems like a good idea @pratapvardhan up for implementing @llllllllll's suggestion?

pratapvardhan:
@cpcloud , @llllllllll does the new internal `pandas_sort` function seem right?

blaze/compatibility.py
```diff
@@ -133,3 +136,17 @@ def assert_series_equal(left, right, check_names=True, **kwargs):
...
+    return df.sort(*args, **kwargs)
+
+def _pandas_sort_new(df, *args, **kwargs):
+    """Use pd.sort_values() for newer versions"""
+    return df.sort_values(*args, **kwargs)
+
+if LooseVersion(pd.__version__) < '0.17.0':
+    pandas_sort = _pandas_sort_old
+else:
+    pandas_sort = _pandas_sort_new
```
cpcloud: you could probably do this with less code:

```python
pdsort = methodcaller(
    'sort' if LooseVersion(pd.__version__) < '0.17.0' else 'sort_values'
)
```

cpcloud: You should define this in `compute/pandas.py` since it's only relevant there. `compatibility` is for language compatibility

pratapvardhan:
@cpcloud -- thanks for the tip, pushed the updated commit.

cpcloud:
@pratapvardhan nice! thanks for the update ... merging!

pratapvardhan:
Thanks @cpcloud =)

PR #1327, opened 2015-11-27
Fix date/datetime subtraction dshape

### LINKED ISSUE #1322: datetime subtraction has incorrect dshape, fails to compute on scalars
I'd expect subtraction of `datetime` type symbols to be of type `timedelta`.
```
In [27]: d = bz.symbol('d', ' var * datetime')

In [28]: (d - d).dshape
Out[28]: dshape("var * datetime")
```

The computation correctly produces a `timedelta` when computed against numpy arrays:
```
In [35]: bz.compute(d - d, array([np.datetime64('2014-01-01')]))
Out[35]: array([0], dtype='timedelta64[D]')
```

However, symbols of scalar `datetime` type fail with `NotImplementedError`:

```
In [55]: bz.compute(d - d, datetime.utcnow())
...
NotImplementedError: Don't know how to compute:
expr: d - d
data: {d: datetime.datetime(2015, 11, 24, 0, 47, 31, 468095)}
```
Computing on `np.datetime64` fails similarly.

cpcloud:
Fixes #1322

### DIFF
--- a/blaze/expr/arithmetic.py
+++ b/blaze/expr/arithmetic.py
@@ -189,6 +189,8 @@ class Sub(Arithmetic):
                 return ct.TimeDelta('D')
             else:
                 return ct.timedelta_
+        elif isinstance(lhs, ct.Time) or isinstance(rhs, ct.Time):
+            raise TypeError('Difference of two Time types not implemented')
         else:
             return promote(lhs, rhs)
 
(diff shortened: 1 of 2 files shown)


llllllllll:
This looks good for the problem of datetimes. I think the we really need to build the type validation system. imho, if you have a constructed node it should mean that the types have already been checked. It makes a lot of the code easier if we can use that assumption.

cpcloud:
Yeah, it's probably not worth keeping this as it will just create more technical debt.

Issue #919, opened 2015-01-12
Comprehensive test suite with py.test


mrocklin:
While I'm hesitant to buy in to `py.test` in per-module unit-tests I do think that it would be worth experimenting on a comprehensive test suite.  The current solution in https://github.com/ContinuumIO/blaze/blob/master/blaze/compute/tests/test_comprehensive.py could be improved.

kwmsmith:
@ltransom you've had some thoughts along these lines.  Thoughts on this approach?

cpcloud:
I have a PR up that refactor the existing comprehensive test suite into a py test friendly version using parametrized fixtures

PR #1328, opened 2015-11-27
Large refactor of test comprehensive

cpcloud:
The benefit here is that after this PR you can each individual expression and
which backend it's being run against.


kwmsmith:
related to #919.

kwmsmith:
Closing for now; may revisit at some point in the future.

PR #1329, opened 2015-12-01
Allow resources to be added dynamically to blaze server.

### LINKED ISSUE #1177: Allow dynamic registering of new data source
1. Allowing different data source to be associated to different url path. 
For example, given a set of data:

```yaml
iris:
  source: sqlite:///../examples/data/iris.db
accounts:
  source: ../examples/data/accounts.json.gz
```
access each part separately via:
http://localhost:6363/iris,  
and
http://localhost:6363/accounts

 2. Further, allowing dynamic registering of new data source after the original construction of server, the server may start with empty data source, and grow as different data source become available. 

For example, adding a new piece

```yaml
cities:
  source: ../examples/data/cities.csv
```
to the server above on the run, and addressing the newly added via: http://localhost:6363/cities

cowlicks:
I'm still having some trouble with this, just getting it out there for now.

closes #1177

### DIFF
--- a/blaze/server/server.py
+++ b/blaze/server/server.py
@@ -4,6 +4,7 @@ import socket
 import functools
 import re
 from warnings import warn
+import collections
 
 import flask
 from flask import Blueprint, Flask, request, Response
@@ -446,6 +447,17 @@ def addserver():
         return ("Bad data.  Got %s " % request.data, 400)  # 400: Bad Request
 
 
-    _get_data.cache[flask.current_app].update({(k, resource(v))
-                                               for k, v in payload.items()})
+    data = _get_data.cache[flask.current_app]
+
+    data_not_mm_msg = ("Cannot update blaze server data since its current data"
+                       " is a %s and not a mutable mapping (dictionary like).")
+    if not isinstance(data, collections.MutableMapping):
+        return (data_not_mm_msg % type(data), 400)
+
+    payload_not_mm_msg = ("Cannot update blaze server with a %s payload, since"
+                          " it is not a mutable mapping (dictionary like).")
+    if not isinstance(payload, collections.MutableMapping):
+        return (payload_not_mm_msg % type(payload), 400)
+
+    data.update({(k, resource(v)) for k, v in payload.items()})
     return 'OK'
--- a/blaze/server/server.py
+++ b/blaze/server/server.py
@@ -436,6 +436,12 @@ def compserver(payload, serial):
 @authorization
 @check_request
 def addserver(payload, serial):
+    """Add a data resource to the server.
+
+    The reuest should contain serialized MutableMapping (dictionary) like
+    object, and the server should already be hosting a MutableMapping
+    resource.
+    """
     data = _get_data.cache[flask.current_app]
 
     data_not_mm_msg = ("Cannot update blaze server data since its current data"
--- a/blaze/server/server.py
+++ b/blaze/server/server.py
@@ -26,7 +26,7 @@ from datashape.predicates import iscollection, isscalar
 from odo import odo
 
 import blaze
-from blaze import compute
+from blaze import compute, resource
 from blaze.expr import utils as expr_utils
 from blaze.compute import compute_up
 
@@ -150,8 +150,6 @@ class Server(object):
     >>> server = Server({'accounts': df})
     >>> server.run() # doctest: +SKIP
     """
-    __slots__ = 'app', 'data', 'port'
-
     def __init__(self, data=None, formats=None, authorization=None):
         app = self.app = Flask('blaze.server.server')
         if data is None:
@@ -423,3 +421,31 @@ def compserver():
         'data': result,
         'names': expr.fields
     })
+
+@api.route('/add', methods=['POST', 'HEAD', 'OPTIONS'])
+@crossdomain(origin='*', methods=['POST', 'HEAD', 'OPTIONS'])
+@authorization
+def addserver():
+    content_type = request.headers['content-type']
+    matched = mimetype_regex.match(content_type)
+
+    if matched is None:
+        return 'Unsupported serialization format %s' % content_type, 415
+
+    try:
+        serial = _get_format(matched.groups()[0])
+    except KeyError:
+        return (
+            "Unsupported serialization format '%s'" % matched.groups()[0],
+            415,
+        )
+
+    try:
+        payload = serial.loads(request.data)
+    except ValueError:
+        return ("Bad data.  Got %s " % request.data, 400)  # 400: Bad Request
+
+
+    _get_data.cache[flask.current_app].update({(k, resource(v))
+                                               for k, v in payload.items()})
+    return 'OK'
--- a/blaze/server/server.py
+++ b/blaze/server/server.py
@@ -20,7 +20,7 @@ except ImportError:
             return wrapped
         return wrapper
 
-from toolz import assoc
+from toolz import assoc, valmap
 
 from datashape import Mono, discover
 from datashape.predicates import iscollection, isscalar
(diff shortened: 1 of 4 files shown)


cowlicks:
API question:

A new data source is added to the severer by passing it a dictionary like `{'data_name': data_resource}`.

And recall that a blaze server is either instantiated with no data, a single data source, or a dict of data sources.

In the single data case, if I want to add `{'foo': bar}`, I have no name for my original data. And moving the original data source into a dictionary would change the datashape of the server. What should I name it?

Options:
* All data on the blaze server must be in a dictionary, so we always have names for separate datasets. This breaks backward compatibility.
* When adding a new datasource to a new blaze server, move the original data into a dictionary with some arbitrary name that can be learned by asking the server for its datashape. 


I think option 1 is sensible.

cowlicks:
cc @llllllllll @cpcloud

llllllllll:
The server always serves a single resource, the dictionary case is not special it is just that the single resource is itself a record. I could see only supporting the update if the resource is a mutable mapping.

cowlicks:
Anything else y'all'd like to see?

blaze/server/server.py
```diff
@@ -406,3 +406,42 @@ def compserver():
...
+    except ValueError:
+        return ("Bad data.  Got %s " % request.data, 400)  # 400: Bad Request
+
+
+    data = _get_data.cache[flask.current_app]
+
+    data_not_mm_msg = ("Cannot update blaze server data since its current data"
+                       " is a %s and not a mutable mapping (dictionary like).")
+    if not isinstance(data, collections.MutableMapping):
+        return (data_not_mm_msg % type(data), 400)
```
cowlicks: Is this the right error code?

blaze/server/server.py
```diff
@@ -406,3 +406,42 @@ def compserver():
...
+
+    try:
+        payload = serial.loads(request.data)
+    except ValueError:
+        return ("Bad data.  Got %s " % request.data, 400)  # 400: Bad Request
+
+
+    data = _get_data.cache[flask.current_app]
+
+    data_not_mm_msg = ("Cannot update blaze server data since its current data"
```
cowlicks: How are these messages?

blaze/server/server.py
```diff
@@ -406,3 +406,42 @@ def compserver():
...
+                       " is a %s and not a mutable mapping (dictionary like).")
+    if not isinstance(data, collections.MutableMapping):
+        return (data_not_mm_msg % type(data), 400)
+
+    payload_not_mm_msg = ("Cannot update blaze server with a %s payload, since"
+                          " it is not a mutable mapping (dictionary like).")
+    if not isinstance(payload, collections.MutableMapping):
+        return (payload_not_mm_msg % type(payload), 400)
+
+    data.update({(k, resource(v)) for k, v in payload.items()})
```
cowlicks: This should be a dict comprehension not a set comprehension.

llllllllll: This could also be `toolz.valmap(resource, payload)`. Passing an iterable of tuples to update is the same as passing a dict here though.

blaze/server/server.py
```diff
@@ -406,3 +406,42 @@ def compserver():
...
+    try:
+        serial = _get_format(matched.groups()[0])
+    except KeyError:
+        return (
+            "Unsupported serialization format '%s'" % matched.groups()[0],
+            415,
+        )
+
+    try:
+        payload = serial.loads(request.data)
```
llllllllll: since we are just passing k:v pairs, do you think it would make sense to encode this as query params, is there a type of resource that can have a comma in it?

cowlicks: There is a discussion of this in #1177 

I'm still thinking about this.

cowlicks:
Closes #1177

blaze/server/server.py
llllllllll: maybe a `422 Unprocessable Entity` would be better here since the syntax of the message is okay

cowlicks:
In #1177  @pzwang mentions accessing data on the server URL's like http://localhost:6363/cities

Does this work? It is not working for me locally in a web browser.

llllllllll:
There is no data endpoint like that in the default blaze blueprint.

blaze/server/server.py
cowlicks: Thoughts on this @cpcloud ? I prefer doing this in a POST since it is simpler. But if encoding it as a query is better we should do that.

llllllllll: One positive of the post is the ability to send over more complicated structures in the future.

blaze/server/server.py
```diff
@@ -113,6 +114,32 @@ def authorized(*args, **kwargs):
     return authorized
 
 
+def check_request(f):
```
cowlicks: I moved this stuff to a decorator since it was copied in the `compserver` and `addserver` functions.

cpcloud:
@cowlicks LGTM, can you add a release note under `Improved Backends`? thx

cpcloud:
@llllllllll any more feedback?

cowlicks:
Merging soon

docs/source/whatsnew/0.9.0.txt
```diff
@@ -25,6 +25,7 @@ Improved Backends
 
 * Adds support for :func:`~blaze.expr.collections.tail` in the sql backend
   (:issue:`1289`).
+* Blaze Serever now supports dynamically adding datasets (:issue:`1329`).
```
cpcloud: typo here `Serever` -> `Server`

cpcloud:
@cowlicks can you add a doc blurb in the blaze server docs showing a minimal working example?

cowlicks:
@cpcloud the current example code with requests in the docs is not working for me. I'm working on it.

docs/source/server.rst
```diff
@@ -324,6 +324,22 @@ These queries deconstruct the Blaze expression as nested JSON.  The ``":leaf"``
 string is a special case pointing to the base data.  Constructing these queries
 can be difficult to do by hand, fortunately Blaze can help you to build them.
 
+Adding Data to the Server
```
cowlicks: @cpcloud doc blurb

docs/source/server.rst
```diff
@@ -309,7 +309,7 @@ server::
    }
 
    $ curl \
-       -H "Content-Type: application/json" \
+       -H "Content-Type: application/vnd.blaze+json" \
```
cowlicks: @llllllllll I changed all of these.

blaze/server/tests/test_server.py
```diff
@@ -504,3 +513,94 @@ def test_isin(test, serial):
...
+        data=blob,
+    )
+    assert 'OK' in response1.status
+
+    # check for expected server datashape
+    response2 = test.get('/datashape')
+    data2 = data.copy()
+    data2.update({'iris': resource(iris_path)})
+    expected2 = str(discover(data2))
+    assert response2.data.decode('utf-8') == expected2
```
llllllllll: can we check the dshape before adding and then compare it to the dshape after adding the new resource?

blaze/server/tests/test_server.py
```diff
@@ -504,3 +513,94 @@ def test_isin(test, serial):
...
+def test_add_data_to_server(test, serial):
+    # add data
+    iris_path = example('iris.csv')
+    blob = serial.dumps({'iris': iris_path})
+    response1 = test.post(
+        '/add',
+        headers=mimetype(serial),
+        data=blob,
+    )
+    assert 'OK' in response1.status
```
llllllllll: we can do: `assert response1.status_code == 200`

blaze/server/tests/test_server.py
```diff
@@ -504,3 +513,94 @@ def test_isin(test, serial):
...
+def test_cant_add_data_to_server(iris_server, serial):
+    # try adding more data to server
+    iris_path = example('iris.csv')
+    blob = serial.dumps({'iris': iris_path})
+    response1 = iris_server.post(
+        '/add',
+        headers=mimetype(serial),
+        data=blob,
+    )
+    assert '422' in response1.status
```
llllllllll: we can also check the `status_code` here

blaze/server/tests/test_server.py
cowlicks: Yeah. This caught a bug with the tests actually. I'm mutating the global server, since this gets run with multiple formats we are updating the server with data it already has.

blaze/server/tests/test_server.py
cowlicks: :+1:

PR #1334, opened 2015-12-02
WIP: Window API

cpcloud:
This is a work in progress. There are probably things missing.

In the next two days I plan to add support for Windows on other
expressions besides base `sa.sql.elements.ColumnElement`s.

For example, a window function on a subquery such as a `GROUP BY` doesn't work ATM.

Feedback in any form would be most helpful.

cc @llllllllll

### DIFF
--- a/blaze/expr/reductions.py
+++ b/blaze/expr/reductions.py
@@ -14,6 +14,38 @@ from .strings import isstring
 from .expressions import dshape_method_list, method_properties
 
 
+class Window(Expr):
+    __slots__ = (
+        '_hash',
+        '_child', 'group_by', 'sort_by', 'preceding', 'following',
+        '_asdshape'
+    )
+
+    __inputs__ = '_child', 'group_by', 'sort_by'
+
+    def __str__(self):
+        slots = 'group_by', 'sort_by', 'preceding', 'following'
+        return '%s(%s, %s)' % (
+            type(self).__name__,
+            self._child,
+            ', '.join('%s=%r' % (slot, getattr(self, slot)) for slot in slots)
+        )
+
+    @property
+    def _inputs(self):
+        return list(filter(
+            lambda x: x is not None,
+            [getattr(self, i) for i in self.__inputs__]
+        ))
+
+    def _dshape(self):
+        return self._asdshape
+
+    @property
+    def _name(self):
+        return self._child._name
+
+
 class Reduction(Expr):
 
     """ A column-wise reduction
@@ -98,6 +130,19 @@ class Reduction(Expr):
             return '%s(%s, %s)' % (name, self._child, ', '.join(kwargs))
         else:
             return '%s(%s)' % (name, self._child)
+
+    def over(self,
+             group_by=None, sort_by=None, preceding=None, following=None):
+        return Window(
+            self,
+            group_by=group_by,
+            sort_by=sort_by,
+            preceding=preceding,
+            following=following,
+            _asdshape=DataShape(
+                *(self._child._child.shape + (self._child.schema,))
+            ),
+        )
 
 
 class any(Reduction):
--- a/blaze/expr/reductions.py
+++ b/blaze/expr/reductions.py
@@ -1,9 +1,12 @@
 from __future__ import absolute_import, division, print_function
 
+import numbers
+
 import datashape
 from datashape import Record, DataShape, dshape, TimeDelta
 from datashape import coretypes as ct
 from datashape.predicates import iscollection, isboolean, isnumeric, isdatelike
+import numpy as np
 from numpy import inf
 from odo.utils import copydoc
 import toolz
@@ -17,7 +20,7 @@ from .expressions import dshape_method_list, method_properties
 class Window(Expr):
     __slots__ = (
         '_hash',
-        '_child', 'group_by', 'sort_by', 'preceding', 'following',
+        '_child', 'kind', 'group_by', 'sort_by', 'preceding', 'following',
         '_asdshape'
     )
 
@@ -33,10 +36,10 @@ class Window(Expr):
 
     @property
     def _inputs(self):
-        return list(filter(
-            lambda x: x is not None,
-            [getattr(self, i) for i in self.__inputs__]
-        ))
+        return [
+            i for i in [getattr(self, i) for i in self.__inputs__]
+            if i is not None and isinstance(i, Expr)
+        ]
 
     def _dshape(self):
         return self._asdshape
@@ -131,17 +134,39 @@ class Reduction(Expr):
         else:
             return '%s(%s)' % (name, self._child)
 
-    def over(self,
+    def over(self, kind='ROWS',
              group_by=None, sort_by=None, preceding=None, following=None):
+        kind = kind.upper()
+        if kind != 'ROWS' and kind != 'RANGE':
+            raise ValueError(
+                '"kind" must be either "ROWS" or "RANGE" (case insensitive)'
+            )
+        if preceding is not None and not isinstance(
+            preceding, (numbers.Integral, np.integer)
+        ):
+            raise TypeError(
+                '"preceding" must be an integer, got %s of type %r' % (
+                    preceding, type(preceding).__name__
+                )
+            )
+        if following is not None and not isinstance(
+            following, (numbers.Integral, np.integer)
+        ):
+            raise TypeError(
+                '"following" must be an integer, got %s of type %r' % (
+                    following, type(following).__name__
+                )
+            )
         return Window(
             self,
+            kind=kind,
             group_by=group_by,
             sort_by=sort_by,
             preceding=preceding,
             following=following,
             _asdshape=DataShape(
                 *(self._child._child.shape + (self._child.schema,))
-            ),
+            )
         )
 
 
--- a/blaze/expr/tests/test_reductions.py
+++ b/blaze/expr/tests/test_reductions.py
@@ -100,17 +100,18 @@ def test_reduction_naming_with_generated_leaves():
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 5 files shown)


cpcloud:
cc @kwmsmith and @ltransom 

this could use some review especially on the API

ltransom:
@cpcloud and @kwmsmith I reviewed the API and changes.  They are consistent with what we discussed.

+1 on the merge from me.

kwmsmith:
Looks great.

The `group_by` kwarg is new, if we discussed that two weeks ago, I don't recall the conversation.

Given that we have a `by()` / `group_by()` expression, and will likely be improving its API, does it make sense to have it be the go-to source for all grouping, rather than having multiple places in the API where group-bys occur?  

IOW instead of this:

```python
expr = tdate.amount.sum().over(preceding=2, group_by=tdate.name)
```

what about this?

```python
expr = tdate.amount.sum().over(preceding=2).group_by(tdate.name)
```

Same goes for sorting:

```python
expr = tdate.amount.sum().over(preceding=2).sort(by=tdate.occurred_on)
```

blaze/expr/reductions.py
```diff
@@ -14,6 +17,38 @@
...
+    def __str__(self):
+        slots = 'group_by', 'sort_by', 'preceding', 'following'
+        return '%s(%s, %s)' % (
+            type(self).__name__,
+            self._child,
+            ', '.join('%s=%r' % (slot, getattr(self, slot)) for slot in slots)
+        )
+
+    @property
+    def _inputs(self):
```
llllllllll: should we redfine `_inputs` on the superclass to use this? I imagine the case of an input optionally being a literal or expression is probably common. How does `BinOp` handle this?

blaze/expr/reductions.py
```diff
@@ -14,6 +17,38 @@
...
+        return [
+            i for i in [getattr(self, i) for i in self.__inputs__]
+            if i is not None and isinstance(i, Expr)
```
llllllllll: if `i` is `None` then `isinstance(i, Expr` is False.

cpcloud: right ... duh

blaze/expr/reductions.py
cpcloud: Actually I stole this from `BinOp` :) so yes, that seems reasonable

blaze/expr/tests/test_reductions.py
```diff
@@ -97,3 +97,33 @@ def test_max_min_on_datetime_and_timedelta(reduc, measure):
...
+        list('ab'),
+        [None, 0, 3],
+        [None, 0, 3]
+    ))
+)
+def test_basic_over(reduction, group_by, sort_by, preceding, following):
+    t = symbol('t', 'var * {a: int64, b: ?float64}')
+    expr = getattr(t.b, reduction)().over(
+        group_by=group_by,
+        sort_by=sort_by,
```
llllllllll: Should we infer the `sort_by` from the child instead?

blaze/expr/reductions.py
```diff
@@ -99,6 +134,41 @@ def __str__(self):
...
+            )
+        return Window(
+            self,
+            kind=kind,
+            group_by=group_by,
+            sort_by=sort_by,
+            preceding=preceding,
+            following=following,
+            _asdshape=DataShape(
+                *(self._child._child.shape + (self._child.schema,))
```
llllllllll: should we move this logic to the `_dshape` function of `Window`?

blaze/expr/tests/test_reductions.py
cpcloud: We could do that, but it has a slightly different meaning when combined with `group_by`. `sort_by` sorts within a group when `group_by` is provided

llllllllll: ah, I see. This makes sense then. When we change the `groupby` api to be more pandas like, do you forsee changing the over syntax to by based on that object?

blaze/expr/reductions.py
cpcloud: no, because when we make new leaves in `compute` from the results of computed expressions `self._child` will be a leaf by the time we get to the `compute_up` for `Window` and therefor `self._child._child` won't exist

blaze/expr/tests/test_reductions.py
cpcloud: similar to what @kwmsmith said we could allow something like the following:

```python
df.a.sum().over(preceding=2).sort(df.c).group_by(df.b)
df.a.sum().over(preceding=2).group_by(df.b).sort(df.c)
df.a.sum().over(preceding=2).sort(df.c)
df.a.sum().over(preceding=2).group_by(df.b)
```

We *could* do this by having some hidden classes like `WindowGroupBy`, `WindowSortBy` whose methods simply create a new `Window` instance when you call their `sort` or `group_by` methods

blaze/expr/reductions.py
llllllllll: This makes sense, thanks for clearing this up. Can we put a comment either here or in the `_dshape` method so  we don't forget this later.

blaze/expr/tests/test_reductions.py
llllllllll: I could see suporting both options. This would allow you to pass all that information in at the construction of the `over` call _or_ chain new behavior as you go. I Like the way the current `over` call lets you express all the relevant inforation quickly. Right now the only fear I would have with the method chaining would be that the `sort` top level function creates `Sort` objects, not `WindowSort` objects. Currently `a.b()` is always the same as `b(a)` so I am not thrilled about changing that.

llllllllll: Actually, we could always dispatch the sort function

blaze/expr/reductions.py
```diff
@@ -99,6 +134,41 @@ def __str__(self):
         else:
             return '%s(%s)' % (name, self._child)
 
+    def over(self, kind='ROWS',
```
llllllllll: Could we put a docstring here for users.

cpcloud: Yep will do

blaze/expr/tests/test_reductions.py
cpcloud: cc @kwmsmith we continued the points you raised here.

cpcloud:
@llllllllll @kwmsmith I've added `sort_by` and `group_by` methods on `Window` expressions which return a new `Window` expression filling in the relevant slot

blaze/expr/reductions.py
```diff
@@ -14,6 +17,51 @@
...
+        )
+
+    def _dshape(self):
+        return self._asdshape
+
+    @property
+    def _name(self):
+        return self._child._name
+
+    def sort_by(self, key):
```
llllllllll: Should we call this `sort` to be consistent?

Also, should we validate that the sort key is in the fields?

cpcloud: I'm not really a huge fan of `sort` (I always think mutation), though it is a somewhat standard API

cpcloud: We should validate, which means we'll have to keep the _child's _child around

cpcloud: Or at least the dshape

llllllllll: I agree, mostly because of `list.sort`. I am just concerned with spelling this two different ways. Maybe we can deprecate `sort` on old expressions expressions and add `sort_by` for 0.9.0 and only implement `sort_by` for windows. This doesn't break any existing code and lets us move forward.

cpcloud: We could also change everything to sort_by, though this doesn't look very good on single columns of things

cpcloud: Oh I see that's what you're saying, right?

llllllllll: ah, I hadn't considered the single column case. I don't have super strong feelings about this but I normally prefer standardizing the names of operations.

cpcloud: i really wish there was keyword only args in py2 then we could do this:

```python
def sort(self, *, by=None):
    pass
```

llllllllll: what would you pass for scalar types?

cpcloud: Not sure ... maybe `None`?

cpcloud: The problem is that arrays aren't typically named things so saying "sort by something" is really awkward

cpcloud: what do you think about `.sorted()`?

llllllllll: I think sorted is nice, the builtin `sorted` is functional so it gives the right idea.

llllllllll: Although, the verb tense is different. For example, we normally say `Add` or `Reduce` not `Added` and `Reduced`.

kwmsmith: > I'm not really a huge fan of sort (I always think mutation), though it is a somewhat standard API

NumPy's `sort()` is pure, so there's precedence there, at least.

There's always Python's non-mutating `sorted()`, but that's probably worse than `sort()`.  If we consistently repeat to users, "blaze works with expressions; all operations are expressions that compute values", then I think that should drive it home.

cpcloud: Yeah, numpy has it both ways though, `np.sort` is pure but `ndarray.sort()` isn't. I'm probably going to stay with `sort` for now

cpcloud: I still think the `sort` API looks weird in this context. I think this is a similar problem to `join` where array-land and database-land prioritize things differently at the expense of other things. array-land can be under generalized (in this case) and database-land can be redundant (having to say `a` twice to sort a single column `select a from table order by a`)

cpcloud:
I may push this to 0.9.1

postelrich:

PR #1335, opened 2015-12-03
kill the attr cache

llllllllll:
The attr cache held a reference to every blaze expression that ever had
an attribute looked up on it. This cache prevented them from ever
getting garbage collected.

Moving the attr_cache to the objects' `__dict__` slot allows for the
object to become isolated and the cycle detector can kill the object.

### DIFF
--- a/blaze/expr/expressions.py
+++ b/blaze/expr/expressions.py
@@ -48,9 +48,6 @@ __all__ = [
 ]
 
 
-_attr_cache = dict()
-
-
 def isvalid_identifier(s):
     """Check whether a string is a valid Python identifier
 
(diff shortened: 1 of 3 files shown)


blaze/expr/expressions.py
```diff
@@ -104,8 +101,7 @@ class Expr(Node):
     contains shared logic and syntax.  It in turn inherits from ``Node`` which
     holds all tree traversal logic
     """
-
-    __slots__ = '_hash', '__weakref__'
+    __slots__ = '_hash', '__weakref__', '__dict__'
```
cpcloud: With this change, do we lose the benefit of a lower memory footprint from using `__slots__`? It's probably not a huge performance benefit for us right now, I'm just curious.

cpcloud:
Needs a release note, and generally +1 on this change.

blaze/expr/expressions.py
llllllllll: We will now allocate one extra dictionary per expression. We will still preallocate the space for all of the other slots so there is still some use for the `__slots__`.

llllllllll:
Added the entry in the whatsnew, merging on pass.

PR #1342, opened 2015-12-04
ENH: allow clients to send extra compute and odo kwargs

llllllllll:
`blaze.compute` accepts keyword args that can change how computations are
performed. This change forwards two extra dictionaries to the server for
use in computation and odo conversion. These are:

`compute_kwargs` - `**`unpacked into the compute call on the server
`odo_kwargs` - `**`unpacked into the odo call on the server

### DIFF
--- a/blaze/interactive.py
+++ b/blaze/interactive.py
@@ -232,7 +232,7 @@ def coerce_to(typ, x, odo_kwargs=None):
     try:
         return typ(x)
     except TypeError:
-        return odo(x, typ, **odo_kwargs or {})
+        return odo(x, typ, **(odo_kwargs or {}))
 
 
 def coerce_scalar(result, dshape, odo_kwargs=None):
--- a/blaze/interactive.py
+++ b/blaze/interactive.py
@@ -1,10 +1,9 @@
 from __future__ import absolute_import, division, print_function
 
 from collections import Iterator
-from functools import reduce
-
 import decimal
 import datetime
+from functools import reduce, partial
 import itertools
 import operator
 import warnings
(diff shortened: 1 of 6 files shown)


llllllllll:
@cpcloud @kwmsmith any thoughts on this?

blaze/interactive.py
```diff
@@ -229,26 +228,27 @@ def short_dshape(ds, nlines=5):
     return s
 
 
-def coerce_to(typ, x):
+def coerce_to(typ, x, odo_kwargs=None):
     try:
         return typ(x)
     except TypeError:
-        return odo(x, typ)
+        return odo(x, typ, **odo_kwargs or {})
```
kwmsmith: I presume the precedence here is equivalent to `**(odo_kwargs or {})`, correct?  What about putting explicit parens around the disjunction to make it more clear -- the spacing makes it look strange at first read.

docs/source/whatsnew/0.9.0.txt
```diff
@@ -26,6 +26,16 @@ Improved Backends
 * Adds support for :func:`~blaze.expr.collections.tail` in the sql backend
   (:issue:`1289`).
 * Blaze Server now supports dynamically adding datasets (:issue:`1329`).
+* Two new keyword only arguments are added to :func:`~blaze.compute` for use
+  when computing against a :class:`~blaze.Client` object:
+
+  1. ``compute_kwargs``: This is a dictionary to send to the server to use as
+     keyword arguments when calling ``compute`` on the server.
+  2. ``odo_kwargs``: This is a dictionary to send to the server to use aas
```
kwmsmith: aas -> as

llllllllll:
Updated to address the comments, anything else before merging?

blaze/server/tests/test_server.py
```diff
@@ -621,3 +658,77 @@ def test_bad_add_payload(empty_server, serial):
...
+        headers=mimetype(serial),
+        data=serial.dumps(bad_query),
+    )
+    assert result.status_code == 500
+    assert b'return_df must be passed' in result.data
+
+    good_query = {
+        'expr': to_tree(expr),
+        'compute_kwargs': {
+            'return_df': odo(DumbResource.df, list),
```
kwmsmith: There's a subtlety here -- this is the same kv pair as for the `odo_kwargs` above which I didn't see at first; since it ends up hitting the `compute_down()` dispatch on line 62, it ultimately calls `dumb_to_df` and everything works out.

kwmsmith:
@llllllllll LGTM, thanks.

blaze/server/tests/test_server.py
llllllllll: Correct. In the `odo_kwargs` test, `expr` is just a field which does not call `compute_down`. This changes the `expr` to have a sort on it causing it hit the `compute_down` and that is where the resource is converted into a dataframe.

PR #1343, opened 2015-12-04
Fix failing windows tests and add tests for numpy ops without numba

cpcloud:

PR #1344, opened 2015-12-04
Fix test_numpy_compute.py::test_binary_math for atan2 and hypot

cowlicks:


cpcloud:
I've fixed this here: https://github.com/blaze/blaze/pull/1343

and i found a few bugs along the way that i also fixed

PR #1347, opened 2015-12-04
Only create a ThreadPool if necessary

### LINKED ISSUE #1331: importing blaze starts `cpu_count() + 3` threads
Not exactly a bug, but maybe an improvement request -

In debugging some context switching in our app, I was surprised to see a number of extra python Threads running.  It turned out they are started by the import of blaze, because [`blaze/__init__.py`] (https://github.com/blaze/blaze/blob/737953bfd8c1bfd819eb41f1d9ad17d969081ba9/blaze/__init__.py#L18) imports `blaze/utils.py`, which [instantiates a `multiprocessing.ThreadPool`] (https://github.com/blaze/blaze/blob/737953bfd8c1bfd819eb41f1d9ad17d969081ba9/blaze/utils.py#L27) at module scope.  The pool starts 3 helper threads plus 1 thread per cpu.

I *think* this pool is only used by [h5py] (https://github.com/blaze/blaze/blob/e0474f3131ea50c24113b6133178cee98fa4b727/blaze/compute/h5py.py#L127). Is its creation something that could be made lazy, such that apps don't need to run the extra threads just to import blaze, especially if they don't compute with h5py?

cpcloud:
Closes #1331

### DIFF
--- a/blaze/compute/h5py.py
+++ b/blaze/compute/h5py.py
@@ -6,6 +6,10 @@ from multipledispatch import MDNotImplementedError
 from datashape import DataShape, to_numpy
 from toolz import curry
 
+import psutil
+
+from multiprocessing.pool import ThreadPool
+
 from ..partition import partitions
 from ..expr import Reduction, Field, symbol
 from ..expr import Expr, Slice, ElemWise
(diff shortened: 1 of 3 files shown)


llllllllll:
:+1:

cpcloud:
merging on pass

PR #1348, opened 2015-12-05
ENH: use object identity for isidentical

llllllllll:
This changes `isidentical` into an object identity check. When profiling blaze, I found that this function is called a _lot_. This actually was in the top 5 tottime calls even with a very small percall count because of how often it was called.

This change makes it so that all expression objects are weakly cached by their arguments. This means that any two objects that are identical will have the same object id. This will save memory by reducing copies of the same object and improve performance because `is` checks are extremely fast.


llllllllll:
Test failures are doctests. I will hold off on reworking the docs until the code has been looked at. I want to make sure that this feature is wanted before continuing with it. I saw a 4x speedup over master in one particular test suite but that is not the best source.

cpcloud:
Generally +1 here, will give a more in-depth review after 0.9.0 is released

llllllllll:
ping, @cpcloud @koverholt I just rebased this against master, Looking for some feedback before rewriting the docs that mention `isidentical`

koverholt:
Perhaps @kwmsmith can help with this?

llllllllll:
Oops, I meant to ping Kurt. Sorry about that

kwmsmith:
Thanks for the ping -- looking over this now.

llllllllll:
ping, tests are passing including some changes to the docs

kwmsmith:
This is good stuff -- I see how the caching of the Node subclass' instance's class and arguments inside `Node.__new__` ensures things like:

```python
>>> (a + b) is (a + b)
True
```
works, which I expected to fail at first before digging in to the code.  This behavior is slightly magical, though; I'd prefer we recommend `==` rather than `is`.

One issue arises with `InteractiveSymbol`s that happen to have the same `name`:

```python
In [47]: s0 = bz.Data(1, name='s')

In [48]: s1 = bz.Data(2, name='s')

In [49]: s0
Out[49]: 1

In [50]: s1
Out[50]: 2

In [51]: s0 + s1
Out[51]: 4
```

Actually, it looks like this bug is also on `master` -- I'll make a separate issue out of this.

My main quibble is recommending that users use `is` rather than `==`.  The fact that we're able to use identity checks here due to the caching of all of the instance's data strikes me as an implementation detail, due to the way that we're caching inside `Node.__new__`.

I see that `Node.__eq__` first checks if `self is other`: other than the slight performance hit of calling `__eq__`, are there fundamental reasons why users should do `e1 is e2` rather than `e1 == e2`?

I much prefer recommending `==`: it's idiomatic, it's not surprising to users, and it is somewhat future proof if we have to change how equality checks happen in the future.  The only downside I can see is the slight performance hit of method resolution and call overhead, but we're still way ahead of `isidentical()`.

llllllllll:
I think that it is very important to show `is` vs `==`. In blaze, `==` would mean construct an `Eq` node where `is` would mean, "are these the same expression.

blaze/expr/tests/test_optimize.py
```diff
@@ -10,76 +10,76 @@ def test_lean_on_Symbol():
...
-    assert expr2.isidentical(t[['x', 'z']].relabel(x='X'))
-    assert fields == set(['x', 'z'])
+    assert expr2 is t[['x', 'z']].relabel(x='X')
+    assert fields == {'x', 'z'}
 
 
 def test_merge_with_table():
     expr = lean_projection(merge(t, a=t.x+1))
-    assert expr.isidentical(expr)
+    assert expr is expr  # wut?
```
kwmsmith: How about fixing this up while we're here:

```python
m = merge(t, a=t.x+1)
assert lean_projection(m) is m
```

kwmsmith:
> I think that it is very important to show is vs ==. In blaze, == would mean construct an Eq node where is would mean, "are these the same expression. 

Of course, you're right -- the tests I was running didn't take that branch, just the `self is other` one...

I see now that because we can't use `==` for this, the only other builtin operator is `is`, which likely motivated the original `isidentical` function / method.

I still have questions & concerns:

* This is a fundamental change that welds expression identity with expression equality (not `Eq` equality).  It obviously affects dozens of modules, and is user facing.  It gives me pause because there's no way to hook into `is`, so using `is` for both identity and equality is fairly constraining.  It'll be hard to go back if we ever need to modify the behavior here.

* I'd prefer the `Node` instance caching you've implemented to be an internal-only thing that's solely for performance, similar to Python's interning of small strings and small integers.  But because `is` would be the only way to compare expressions, we're locked in to the behavior of this caching as well.  If, in the future, we ever want to change or tweak the caching implementation (or if it starts acting up like other weak ref dictionary stuff has and we need to do major surgery), we're highly constrained due to the need to support `is`.

* There's a less invasive middle way that still improves performance: if we keep `isidentical` around but reimplement it as you have it here--a simple `return a is b` and remove the warning--we solve a lot of these concerns.  We're still using `is` internally, but it isn't user facing, and it leaves a lot of the codebase unmodified.  This way doesn't rule out moving to `is` down the line, and it allows us to test over time whether these changes have any subtle bugs lurking.

* I think I understand the reason you're removing `__slots__` and replacing it with `_arguments`, as well as the reason to remove `_hash`, but could you explain it in case there are subtleties?  Is it because the space savings of `__slots__` is no longer worth it, given that we're instantiating fewer expressions now?

* Similarly, I see you're replacing most (all?) occurrences of `__init__` with `__new__` in `Expr` subclasses -- please walk me through the reasons why.

llllllllll:
I am not really sure why one would ever want two identical expressions to be separate objects. Just conceptually it is really nice to know that if two expressions are the same, they will actually be the same object in memory. Is there a motivating example for making these separate?

I guess I would be okay with removing the warning in `isidentical` and advertising that as the main entry point; however, because I think that equality and identity should be the same concept `is` is a more natural verb for this.

I am removing the `__slots__` because we already have switched to including a `__dict__` in the slots to use the object's dict as the attrcache for methods that are cached from the schema and datashape method lists. This means that we are not actually getting any space saving out of the slots. I think that `__slots__` was probably a premature optimization or a way to help ensure that only a select few attributes were mutable on the objects but we have since strayed from this. We also were not properly inheriting our parents' slots because we were using the slots to define the inputs the the expression which was reflected in the old `Node.__init__`. The new code does not attempt to make `__slots__` fit our use case and instead defines a field that simply describes the names of the arguments to the function. This is just like our `__inputs__` (which I would like to rename since we should not be using the `__*__` namespace). There is even the relationship of `_arguments -> args` just like `__inputs__` -> `_inputs` where the first are the names of the fields and the second are the values associated with those names. `_hash` being a slot was a holdover from when we didn't have a `__dict__` on `Node` objects and it needed to be in the slots for hashing. There was a lot of code that ignored the first slot because it should have been `_hash` so it hung around. This is just cleaning up that debt.

We are swapping all of the `__init__` functions out with `__new__` functions because of the way our object identity caching works. When I call the constructor for any expression I may get an object that was previously allocated and initialized. Any pre setup I want should happen before the allocation has happened. We also cannot use `__init__` because `__init__` is called any time your `__new__` returns an instance of `cls`. For example, using the following class:

```python
class C:
    instance = None

    def __new__(cls):
        if cls.instance is None:
            cls.instance = super().__new__(cls)
        return cls.instance

    def __init__(self):
        print('calling __init__')
```
```python
>>> c = C()
calling __init__
>>> d = C()
calling __init__
>>> c is d
True
```

This shows that even though we are returning the cached object, we will still call the `__init__` func. This would potentially reset the object so we instead call an `_init` function.

Hopefully this clears some of this up.

kwmsmith:
I think I found a bug, but it's not 100% clear to me.  It has to do with hash collisions interacting with the `Node` instance cache.

Say for the sake of argument we have two symbols `r` and `s`, both `int`s, and say also that `hash(r) == hash(s)`.  With the instance cache, the `WeakKeyDictionary()` will have a hash collision, which puts us in an edge case scenario.

Since it's really hard to ensure hash collisions in Python (especially with Python 3), I instrumented `blaze.expr.expressions.Symbol` to have a dumb constant `__hash__` function to trigger the collision:

```python
class Symbol(Expr):
    # ...
    __hash__ = lambda self: 42
```

With this in place, badness happens:

```python
In [1]: import blaze as bz

In [2]: s = bz.symbol('s', 'int')

In [3]: r = bz.symbol('r', 'int')

In [4]: e = s + r

In [6]: f = r + s
ValueError                                Traceback (most recent call last)
<ipython-input-6-2e7f690f5d9e> in <module>()
----> 1 f = r + s

/Users/ksmith/work/blaze/blaze-repo/blaze/expr/core.py in __add__(self, other)
    228
    229     def __add__(self, other):
--> 230         return self._add(other)
    231
    232     def __radd__(self, other):

/Users/ksmith/work/blaze/blaze-repo/blaze/expr/arithmetic.py in _bin(self, other)
    288
    289     def _bin(self, other):
--> 290         result = cons(self, scalar_coerce(self.dshape, other))
    291         result.dshape  # Check that shapes and dtypes match up
    292         return result

/Users/ksmith/work/blaze/blaze-repo/blaze/expr/core.py in __new__(cls, *args, **kwargs)
     82         static_id = cls._static_identity(*args, **kwargs)
     83         try:
---> 84             return cls.__expr_instance_cache[static_id]
     85         except KeyError:
     86             cls.__expr_instance_cache[static_id] = self = super(

/Users/ksmith/anaconda/envs/3.4/lib/python3.4/weakref.py in __getitem__(self, key)
    129
    130     def __getitem__(self, key):
--> 131         o = self.data[key]()
    132         if o is None:
    133             raise KeyError(key)

/Users/ksmith/work/blaze/blaze-repo/blaze/interactive.py in <lambda>(x)
    339 Expr.__float__ = lambda x: convert_base(float, x)
    340 Expr.__complex__ = lambda x: convert_base(complex, x)
--> 341 Expr.__bool__ = lambda x: convert_base(bool, x)
    342 Expr.__nonzero__ = lambda x: convert_base(bool, x)
    343 Expr.__iter__ = into(Iterator)

/Users/ksmith/work/blaze/blaze-repo/blaze/interactive.py in convert_base(typ, x)
    329
    330 def convert_base(typ, x):
--> 331     x = compute(x)
    332     try:
    333         return typ(x)

/Users/ksmith/anaconda/envs/3.4/lib/python3.4/site-packages/multipledispatch/dispatcher.py in __call__(self, *args, **kwargs)
    162             self._cache[types] = func
    163         try:
--> 164             return func(*args, **kwargs)
    165
    166         except MDNotImplementedError:

/Users/ksmith/work/blaze/blaze-repo/blaze/interactive.py in compute(expr, **kwargs)
    169     resources = expr._resources()
    170     if not resources:
--> 171         raise ValueError("No data resources found")
    172     else:
    173         return compute(expr, resources, **kwargs)

ValueError: No data resources found
```

This code works on `master`, so there's definitely something going on when there's a hash collision.

kwmsmith:
Thinking about the bug: `dict.__getitem__(key)` first gets the key in the dictionary at the location `hash(key)` and then compares that with the `key` using `is`.  If there's a hash collision, they aren't identical in memory, so it has to then fallback to an `==` test (`PyObject_RichCompareBool()`).  This triggers a call to `Node.__eq__` and ultimately we hit the code path above.

kwmsmith:
> I am not really sure why one would ever want two identical expressions to be separate objects. Just conceptually it is really nice to know that if two expressions are the same, they will actually be the same object in memory. Is there a motivating example for making these separate?

I completely understand where you're coming from here; languages like Clojure take this perspective on identity and equality, since all core datatypes and containers are immutable, and this is very powerful.  In Clojure, an objects' identity and its value are one and the same.  But even Clojure doesn't take it as far as we're doing here: Clojure allows you to create two separate objects that have the same exact value, so they compare equal, but they're still distinct objects.

As you know in Python, small strings and small integers (and small floats? I'm not sure) are interned for performance.  But it would be a mistake to use `is` instead of `==` when testing equality between strings or between ints, since the interning behavior is an optimization and not a guarantee.

My reasons for not going all the way and making `is` part of Blaze expression's public API are several:

* Python makes no assumption about the relation between object identity and value, so the entire language is geared to treat them separately.  Unifying them as we're doing here goes against the grain somewhat; because of that, I really want to keep it an internal thing that we may change in the future if we have to.  I'm concerned that we may see other non deterministic weak key caching bugs show up again, so I don't want to be tied to this implementation.

* In typical Python, pretty much the only time it's correct to use `is` is when testing for identity with `None`.  I've maybe seen `is` used *correctly* to compare object identities a handful of other times in real world code.  Python users instinctually avoid using `is` outside of these cases.  If we were to bring `is` in as part of Blaze's public API for expressions, it would move us away from the common expectation.  I'm advocating for the principle of least surprise here, basically.

* Combining identity and equality like we're doing here complicates the code: we have to manage a class-level cache of all node instances; that cache complicates all Node subclasses, since we can't implement `__init__` on them; if we want to do anything special we have to use `__new__`, etc.  Again, because we're changing equality semantics from the Python default, we pay the price in code complications.  I want to keep those complications as internal as possible.

* We're basically implementing the flyweight pattern for blaze expressions (as much as I don't like design pattern formalisms, it is useful to provide a common terminology).  Normally flyweight is a memory saving technique for immutable objects that's applied in limited scenarios; we're doing it for *every* blaze expression; again, I want to be able to change this in the future if we have to.

* Even though blaze expressions are immutable, this implementation does introduce a mutable instance cache inside `Node`.  It's not global in the sense that anything can access it arbitrarily, but it is global in the sense that it is tied to the `Node` class, and it's mutating a lot, so there is some complexity tradeoff there, too.

To summarize my long-winded response :)

* +1 on this implementation provided we can fix the hash collision bug, and provided we keep `isidentical()` the public API.

Thanks again for your work here, I think this is a big improvement.

llllllllll:
I am confused by this here:

> an objects' identity and its value are one and the same. ... Clojure allows you to create two separate objects that have the same exact value, so they compare equal, but they're still distinct objects.

If the identity and the value are the same, then how can you have two objects with different identity and the same value?

About the hash collision issue:

> hash(r) == hash(s). With the instance cache, the WeakKeyDictionary() will have a hash collision, which puts us in an edge case scenario.

We are not using a `WeakKeyDictionary` here, instead we are using a `WeakValueDictionary`. This is keyed on a tuple and the values are the `Node` objects. When the dict is the only reference to this node it will be removed; it is frozen during iteration and commited after so that iteration is not dangerous.

We cache all the symbols by their `_static_identity`. By default, and for symbols this is just `(cls,) + tuple(_resolve_args(cls, *args, **kwargs).values())` where `_resolve_args` turns the `*args` and `**kwargs` into an ordered dict that represents the arguments of the expression.

for example:

```python
>>> class MyNode(Node):
...     _arguments = 'a', 'b', 'c'
...
>>> _resolve_args(MyNode, 1, 2, c=3)
OrderedDict([('a', 1), ('b', 2), ('c', 3)])
```

The key for an expression becomes a tuple of the type and all the argname, argvalue pairs. For this to collide is extremely unlikely. I am not sure how we can ever stop hash collisions like this when we have overridden `__eq__`. This is not a new issue to blaze though, we use the `hash` of nodes for many api's including the `scope`s used in the core compute code so I am not sure that this has ever been an issue.

One thing that might ease some concerns is that this is that we have had a lot of success with this exact node caching pattern in the zipline pipeline compute engine (https://github.com/quantopian/zipline/blob/master/zipline/pipeline/term.py#L28). This has been running in production for a while and has not caused any problems, though we did not override `__eq__`. 

I have rebased against master and added a couple commits. I removed the deprecation warning on `isidentical`. do you think we should continue to expose the free function? It doesn't come through the top level blaze api so I think it is fine to just leave this as a method.

Thank you for all of the feedback and seriously considering the tradeoffs with this patch!

kwmsmith:
> If the identity and the value are the same, then how can you have two objects with different identity and the same value?

You're right -- I misspoke here.  I meant that when working with immutable values, distinct objects that have the same value can be thought of as having the same *logical* identity.

> We are not using a WeakKeyDictionary here, instead we are using a WeakValueDictionary. This is keyed on a tuple and the values are the Node objects. When the dict is the only reference to this node it will be removed; it is frozen during iteration and commited after so that iteration is not dangerous.

Thanks for taking the time to explain this (and other parts of this PR).  It's an education for me around `weakref`s in Python :)  I still have a lingering concern in the back of my mind relating to weak ref memoization that we pulled out in another issue because of strange errors we were getting.  If those start cropping up here, we'll have to take a look at this code more carefully and consider other options.

> The key for an expression becomes a tuple of the type and all the argname, argvalue pairs. For this to collide is extremely unlikely. I am not sure how we can ever stop hash collisions like this when we have overridden __eq__. This is not a new issue to blaze though, we use the hash of nodes for many api's including the scopes used in the core compute code so I am not sure that this has ever been an issue.

From an abstract correctness point of view, this is concerning, although I realize at the same time that it's unlikely :)  With Python 3.3's introduction of hash randomization, any collision-related issues that may come up here will be non deterministic unless we set an appropriate env variable.  If we get collisions in the wild it will be basically impossible to reproduce.  I wonder if this is related at all to the other seemingly random weak ref memoization issues we were having elsewhere...

Thanks for pointing out the other places we assume no hash collisions in Blaze, that's good to know.

I'll make a separate issue out of this -- I'd like to brainstorm to see if there's a way we can fix this issue completely.  It'll be hard since we use `__eq__` for other purposes, but I'd still like to see if there are options here.

> I have rebased against master and added a couple commits. I removed the deprecation warning on isidentical.

I see that you're still using `is` for comparisions rather than `.isidentical()`.  To keep `is` usage isolated, I'd prefer to revert these comparisons to use `isidentical()` again.  We're still getting a performance boost over the old implementation of `isidentical()`.

> do you think we should continue to expose the free function? It doesn't come through the top level blaze api so I think it is fine to just leave this as a method.

+1.  I'd rather keep this PR as an internal change only and not change the API, that way we can include it in the 0.9.1 point release :)  We can revisit whether `isidentical` is a method or free function later (and consider other names for it, perhaps `equiv` or `equivalent`?).

> Thank you for all of the feedback and seriously considering the tradeoffs with this patch!

Thanks for putting in the work on it, and for your patience while I made the time to review it.

llllllllll:
I will go through and revert the isidentical->is change; however, I need to finish up some other stuff before coming back to this.

kwmsmith:
> I will go through and revert the isidentical->is change; however, I need to finish up some other stuff before coming back to this.

@llllllllll what's the latest on this one?  Since this PR touches a lot of code, I'd like to bring it to a close sooner rather than later.

llllllllll:
I need to port all the is checks I made to `isidentical`. I just haven't had the time to do that. I still have some concerns about that though because all my profiling of blaze has shown this function to have an absurd call count / tottime even with a near 0 percall count. do you want to discuss this on gitter quick for some more immediate feedback?

llllllllll:
@kwmsmith I actually rebased and got this branch in shape! This is because we were seeing  unreal slowdowns in our production application with blaze comparing objects. We had a moderatly complex blaze query which spend ~17 seconds to build the sql which ran in about ~0.3 seconds. With this branch the time is down below 1 second.

llllllllll:
@kwmsmith Can we merge this soon?

llllllllll:
@kwmsmith just rebased and tests are passing. I would like to merge this today or tomorrow

kwmsmith:
@llllllllll I see you've merged this in; for some reason this thread was buried in my email so I didn't have a chance to comment.

You couldn't have known this, but we're targeting a release next week for an internal product; I should have mentioned this earlier.

This merge is causing problems in other PRs, like the `str-namespace` branch.  Can you take a brief look and help us understand why?

https://travis-ci.org/blaze/blaze/jobs/135960332#L3080

llllllllll:
Sorry, I took the silence to mean that this was okay to merge. I commented on the str branch. We just need to translate the `__slots__` to `_arguments`

PR #1351, opened 2015-12-07
Fix reduction types that could come from decimal children

cpcloud:


cpcloud:
@llllllllll any comments here?

llllllllll:
should all reductions be optional aware? Also, where is the code that updates sum?

cpcloud:
sorry, yeah `sum` is update via the recent datashape maxtype fix

cpcloud:
yeah all reductions should be option aware

cpcloud:
actually, something like count shouldn't be

cpcloud:
so not all

llllllllll:
Ah, good point. Looks good.

cpcloud:
merging on pass

PR #1353, opened 2015-12-11
Fix issue with "by" expression and "count" reduction.

cowlicks:
This fixes an issue with the `by` expression and pandas.

The pandas `compute_by` takes some columns to group on `g`, a summary of reductions for after the groupby `s`, and data `df`.

To execute the `by`, first blaze constructs another dataframe with additional columns that will be reduced with the summary reductions. For example if one reduction is `df.a.sum` the column `df.a` is concated to `df`.  But if one of the reductions is `df.count`, what column do we add? Before we tried adding the whole dataframe which failed. This PR adds a column of zero values to fix the issue.

The below test failed before.
```python
tbig = symbol('tbig',
              'var * {name: string, sex: string[1], amount: int, id: int}')

dfbig = DataFrame([['Alice', 'F', 100, 1],
                   ['Alice', 'F', 100, 3],
                   ['Drew', 'F', 100, 4],
                   ['Drew', 'M', 100, 5],
                   ['Drew', 'M', 200, 5]],
                  columns=['name', 'sex', 'amount', 'id'])

def test_by_with_reduction_on_df():
    expr = by(tbig.name, id_sum=tbig.id.sum(), count=tbig.count())
    compute(expr, dfbig)
```

### DIFF
--- a/blaze/compute/pandas.py
+++ b/blaze/compute/pandas.py
@@ -20,6 +20,7 @@ import fnmatch
 import itertools
 from distutils.version import LooseVersion
 import warnings
+from collections import defaultdict
 
 import numpy as np
 
@@ -30,6 +31,7 @@ from pandas import DataFrame, Series
 from pandas.core.groupby import DataFrameGroupBy, SeriesGroupBy
 
 from toolz import merge as merge_dicts
+from toolz import groupby
 from toolz.curried import pipe, filter, map, concat
 
 import datashape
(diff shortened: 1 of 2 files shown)


kwmsmith:
@cowlicks looks like all the python 2 tests are failing with the same failure.  It doesn't look to be related to your PR, but can you investigate?

cowlicks:
@kwmsmith it is related, I'm investigating

cowlicks:
@kwmsmith fixed

Issue #1336, opened 2015-12-03
Blaze not loading HTTP data


ltransom:
Blaze will not compute on data associated with a RESTful service, although it will properly query the data to get the data shape.

In the attached script, Blaze will hit the REST endpoint and display the data shape, but the compute fails.

```python
from blaze import Data, compute

bdf = Data("http://real-chart.finance.yahoo.com/table.csv?s=%5EGSPC&a=00&b=1&c=2014&d=11&e=31&f=2014&g=d&ignore=.csv")

# will print the datashape for the endpoint
print(bdf.dshape)

# will produce and error
print(compute(bdf))
```

kwmsmith:
Closed by #1354.

PR #1354, opened 2015-12-11
Support http resources and URLs of CSVs.

kwmsmith:
Fixes issue #1336.

Further improvements may include expanding to support other formats
besides CSV.  URLs of CSVs is the major usecase here.

@llllllllll @cowlicks you mind taking a look?

### DIFF
--- a/blaze/compute/tests/test_url_csv_compute.py
+++ b/blaze/compute/tests/test_url_csv_compute.py
@@ -1,0 +1,39 @@
+import pytest
+
+import os
+
+from blaze import Data, compute
+from blaze.utils import raises
+from odo import URL, CSV
+
+import pandas as pd
+import pandas.util.testing as tm
+
+from functools import partial
+
+try:
+    from urllib2 import urlopen
+    from urllib2 import HTTPError, URLError
+except ImportError:
+    from urllib.request import urlopen
+    from urllib.error import HTTPError, URLError
+
+pytestmark = pytest.mark.skipif(raises(URLError,
+                                       partial(urlopen, "http://google.com")),
+                                reason='unable to connect to google.com')
+
+iris_url = ('https://raw.githubusercontent.com/'
+            'blaze/blaze/master/blaze/examples/data/iris.csv')
+
+@pytest.fixture
+def iris_local():
+    thisdir = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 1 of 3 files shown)


blaze/compute/tests/test_url_csv_compute.py
```diff
@@ -0,0 +1,39 @@
...
+    from urllib2 import HTTPError, URLError
+except ImportError:
+    from urllib.request import urlopen
+    from urllib.error import HTTPError, URLError
+
+pytestmark = pytest.mark.skipif(raises(URLError,
+                                       partial(urlopen, "http://google.com")),
+                                reason='unable to connect to google.com')
+
+iris_url = ('https://raw.githubusercontent.com/'
```
llllllllll: Could we have the test spin up a local webserver and then serve this test data? I would prefer to be able to run the tests without internet connection.

cowlicks: +1

cowlicks: I would also be fine with something like `pytest.skipif(no internet)`. [This](http://stackoverflow.com/a/3764660/1609380) looks like a fine way to check for an internet connection.

llllllllll: I would be okay with that too; however, I would prefer to have it local so that it is both faster and capable of being tested without internet.

kwmsmith: > I would also be fine with something like pytest.skipif(no internet). This looks like a fine way to check for an internet connection.

That's exactly what lines 21--23 are doing, which was stolen directly from Odo's `URL` unittests.

> Could we have the test spin up a local webserver and then serve this test data? I would prefer to be able to run the tests without internet connection.

Yes, I'd prefer that too.  And while we're at it, we should refactor Odo's URL tests to use the same logic.

I don't have the time to do this refactoring this week.  In the interests of time and priorites, I made a separate issue for this improvement (#1360).

kwmsmith:
@llllllllll @cowlicks -- I'd like to merge this as-is soon, and do the follow-on local server improvement as a separate PR.  Any further comments?

cowlicks:
I'd like to get this in for 0.9.0 which I'm working on today, so I'm going to merge.

Note that it does skip the test if there is no internet connection.

PR #1364, opened 2015-12-17
Release fixes

cowlicks:


blaze/compute/tests/test_numpy_compute.py
```diff
@@ -18,6 +16,10 @@
 import blaze
 from odo import into
 from datashape import discover, to_numpy, dshape
+try:
+    import numba
+except ImportError:
+    numba = None
```
llllllllll: won't this raise an attribute error later?

cpcloud: I think the only reason these fail is because I'm using a `numba.TypingError` in a `raises` argument somewhere below. That test could be moved to another file and then you wouldn't need to import numba here. Generally these kinds of imports should be avoided as they hide errors

cowlicks: You are correct. But it turns out we didn't actually need numba here anymore. We were checking for a specific binop (fmod) to raise the `numba.TypingError` but it isn't tested there anymore.

Issue #1371, opened 2015-12-28
Don't know how to compute `like()` on SQL select


kwmsmith:
Minimal example:

```python
from odo import odo
import blaze as bz
import pandas as pd

db0 = odo(pd.DataFrame({'a': ['a', 'b', 'c']}), 'sqlite:///db.db::df0')
db1 = odo(pd.DataFrame({'b': ['b', 'c', 'd']}), 'sqlite:///db.db::df1')
dd0 = bz.Data(db0)
dd1 = bz.Data(db1)
j = bz.join(dd0, dd1, on_left='a', on_right='b')
bz.compute(j[j.a.like('*b')])
```

Running the above, I get:

```python
In [1]: %run bug_compute_like_selectable.py
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
/Users/ksmith/work/blaze/blaze-repo/bugfix_compute_like_selectable.py in <module>()
     10 dd1 = bz.Data(db1)
     11 j = bz.join(dd0, dd1, on_left='a', on_right='b')
---> 12 bz.compute(j[j.a.like('*b')])

/Users/ksmith/anaconda/envs/tblaze/lib/python2.7/site-packages/multipledispatch/dispatcher.pyc in __call__(self, *args, **kwargs)
    162             self._cache[types] = func
    163         try:
--> 164             return func(*args, **kwargs)
    165
    166         except MDNotImplementedError:

/Users/ksmith/work/blaze/blaze-repo/blaze/interactive.pyc in compute(expr, **kwargs)
    169         raise ValueError("No data resources found")
    170     else:
--> 171         return compute(expr, resources, **kwargs)
    172
    173

/Users/ksmith/anaconda/envs/tblaze/lib/python2.7/site-packages/multipledispatch/dispatcher.pyc in __call__(self, *args, **kwargs)
    162             self._cache[types] = func
    163         try:
--> 164             return func(*args, **kwargs)
    165
    166         except MDNotImplementedError:

/Users/ksmith/work/blaze/blaze-repo/blaze/compute/core.pyc in compute(expr, d, **kwargs)
    402         d4 = d3
    403
--> 404     result = top_then_bottom_then_top_again_etc(expr3, d4, **kwargs)
    405     if post_compute_:
    406         result = post_compute_(expr3, result, scope=d4)

[...]

PR #1373, opened 2015-12-31
FIX: Fix issue with `Like` selections on SQL selects.

kwmsmith:
`Like` only worked on `ColumnElement`s in sql.py.  This led to
`NotImplementedError`s when calling `like()` on a column of a SQL join for example,
which this commit fixes.

@llllllllll @cpcloud can you take a look?

ping @sandhujasmine

### DIFF
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -999,6 +999,13 @@ def compute_up(t, s, **kwargs):
     )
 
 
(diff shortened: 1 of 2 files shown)


kwmsmith:
Closes #1371

kwmsmith:
PR tested and verified by @sandhujasmine.  Merging.

PR #1374, opened 2015-12-31
FIX: `is` test in `Expr.__eq__`.

kwmsmith:
Shortcuts some expensive `isidentical()` checks for when comparing an
object with itself.

ping @jcrist


kwmsmith:
Closing; #1348 is a superset of this PR.

PR #1377, opened 2016-01-06
Allow extra per-dataset arguments in blaze server yaml file.

kwmsmith:
Extra arguments are passed through to underlying `resource()` call.

Required for some custom backends.

Adds `imports` option that will trigger imports of listed packages before `resource()` call.

Example `blaze-server` yaml file with a single dataset `ds1` and a directory of datasets `dir`:

```
ds1:
    source: "foo://bar/baz"
    dshape: "var * {...}"
    extra_arg: 1
    extra_arg2: "spam!"
    imports: ["foo", "custom_module"]
ds2:
    source: "./dir"
    dshape: "..."
```

Note that this is strictly expanding functionality and will not affect existing Blaze server behavior with YAML spec files.

### DIFF
--- a/blaze/server/spider.py
+++ b/blaze/server/spider.py
@@ -103,7 +103,8 @@ def from_yaml(path, ignore=(ValueError, NotImplementedError), followlinks=True,
         if 'source' not in info:
             raise ValueError('source key not found for data source named %r' %
                              name)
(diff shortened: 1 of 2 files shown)


ltransom:
LGTM.

PR #1378, opened 2016-01-11
Use flask-cors package for CORS; don't use Bokeh anymore.

kwmsmith:
With Bokeh 0.11, the CORS handling was moved.  This removes the Bokeh
dependency for this part of Blaze and uses instead a small MIT-licensed
flask extension.

### DIFF
--- a/blaze/server/server.py
+++ b/blaze/server/server.py
@@ -8,17 +8,7 @@ import collections
 
 import flask
 from flask import Blueprint, Flask, request, Response
-
-try:
-    from bokeh.server.crossdomain import crossdomain
-except ImportError:
-    def crossdomain(*args, **kwargs):
-        def wrapper(f):
-            @functools.wraps(f)
-            def wrapped(*a, **k):
-                return f(*a, **k)
-            return wrapped
-        return wrapper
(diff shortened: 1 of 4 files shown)


kwmsmith:
@ltransom @johneo Can you please review?

blaze/server/server.py
```diff
@@ -219,7 +209,7 @@ def run(self, port=DEFAULT_PORT, retry=False, **kwargs):
 
 
 @api.route('/datashape', methods=['GET'])
-@crossdomain(origin='*', methods=['GET'])
+@cross_origin(origins='*', methods=['GET'])
```
johneo: We need to consider if we want to let any and all browser content to load from the Blaze server, as the `*` essentially defeats the cross-origin content loading protection. Might want to eventually move toward making this configurable or default to localhost:6363?

kwmsmith: I'm in agreement with you; this PR is scoped to just reproduce the current functionality with the `flask-cors` extension.  I'll make a separate feature request for making this configurable.

johneo:
Looks good!

PR #1379, opened 2016-01-11
Remove DyND for now.

kwmsmith:
Due to DyND instability and on the recommendation of DyND developers, we
are removing the DyND dependency from Blaze until DyND reaches its 1.0
release.


kwmsmith:
ping @izaid

izaid:
+1 I think the DyND in Blaze is incredibly old anyway.

llllllllll:
What are the blockers on this?

kwmsmith:
We're waiting on the 0.9.1 release; I don't want to remove a backend as part of a bugfix release, so this is waiting for 0.10.

kwmsmith:
OK, merging based on feedback.

PR #1381, opened 2016-01-11
BUG: fix compute_up(Time, ColumnElement)

llllllllll:
just like `date`, `extract(time from x)` does not work, you can just cast the timestamp to a `time`.

### DIFF
--- a/blaze/expr/datetime.py
+++ b/blaze/expr/datetime.py
@@ -1,9 +1,10 @@
 from __future__ import absolute_import, division, print_function
 
-from .expressions import ElemWise, schema_method_list, method_properties
-
 import datashape
 from datashape import dshape, isdatelike, isnumeric
+
+from .expressions import ElemWise, schema_method_list, method_properties
+from blaze.utils import instance
 
 
 __all__ = ['DateTime', 'Date', 'date', 'Year', 'year', 'Month', 'month', 'Day',
@@ -26,9 +27,10 @@ class DateTime(ElemWise):
     def _name(self):
         return '%s_%s' % (self._child._name, self.attr)
 
-    @property
-    def attr(self):
-        return type(self).__name__.lower()
+    @instance
+    class attr(object):
+        def __get__(self, instance, owner):
+            return owner.__name__.lower()
 
 
 class Date(DateTime):
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -54,11 +54,46 @@ from .core import compute_up, compute, base
 
 
 from ..expr import (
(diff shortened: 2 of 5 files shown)


llllllllll:
Adding the `wip` tag. I would like to talk to someone about standardizing the meaning of these fields. Pandas and sql give very different answers

llllllllll:
Does anyone want to discuss this? ping @cpcloud @kwmsmith

blaze/expr/datetime.py
```diff
@@ -26,9 +27,10 @@ def _schema(self):
     def _name(self):
         return '%s_%s' % (self._child._name, self.attr)
 
-    @property
-    def attr(self):
-        return type(self).__name__.lower()
+    @instance
+    class attr(object):
+        def __get__(self, instance, owner):
+            return owner.__name__.lower()
```
kwmsmith: Why this descriptor implementation rather than the original `@property`?  In what cases is `owner` different from `type(self)` in the original code?

llllllllll: I wanted to be able to access this off of the class and instances

kwmsmith:
@llllllllll I'm working through the PR backlog -- could you update this PR and move the whatsnew entry to the 0.10 file?

llllllllll:
I am still not sure what the correct behavior for the datetime accessors should be in sql. Should we try to model the sql behavior or the pandas behavior. I think we can get the pandas behavior from the sql behavior but going the other way around is harder

kwmsmith:
> Should we try to model the sql behavior or the pandas behavior.

+1 Pandas behavior.

Since Pandas will be the `'core'` return type unless the user specifies otherwise, and since we're targeting users who are familiar with Pandas and who aren't SQL pros, Pandas seems the natural choice when there's ambiguity.

llllllllll:
@kwmsmith I also prefer to favor pandas over sql, do you think that is appropriate to use as a general guideline for future problems like this?

kwmsmith:
> do you think that is appropriate to use as a general guideline for future problems like this?

+ :100: :)

kwmsmith:
ping @llllllllll -- can you update the whatsnew?  And do you have other questions, or is this ready to go?

llllllllll:
I have not changed this to foloow the pandas semantics yet, I haven't had time to come back to this and it is low priority right now. Also, since we are going with pandas-semantics should we drop the `milliseconds` datetime node, this is not a field in pandas.

kwmsmith:
Based on the lower priority, I'm moving this to post 0.10 release.

llllllllll:
sounds good

postelrich:

PR #1382, opened 2016-01-12
timedelta improvements

llllllllll:
depends on: https://github.com/blaze/odo/pull/394 (merged) and https://github.com/blaze/datashape/pull/207 (merged)

This adds:

* `datetime - datetime :: timedelta`
* `datetime - timedelta :: datetime`
* `datetime + timedelta :: datetime`
* support for those cases where lhs, rhs, or both are optional
* reductions over `timedelta` are of type `timedelta` not `float`
* adds `std` and `var` to timedelta in pandas and sql. This requires coercing to int to do the reduction then coercing back because neither pandas or postgres support this nativly.
* corrected datetime subtraction coercing in postgres. Done at expression time with `optimize`

When we are ready to update the type checking for operators these tests will help a lot. I hope that soon we can have some type of datashape based registration / type checking like:

```python
@typecheck('(Integral, Integral) -> Integral')
@typecheck('(datetime, timedelta) -> timedelta')
class add(Arithmetic):
    pass
```

### DIFF
--- a/blaze/compute/pandas.py
+++ b/blaze/compute/pandas.py
@@ -16,6 +16,7 @@ Name: name, dtype: object
 """
 from __future__ import absolute_import, division, print_function
 
+from datetime import timedelta
 import fnmatch
 import itertools
 from distutils.version import LooseVersion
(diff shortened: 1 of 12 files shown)


llllllllll:
@kwmsmith Any comments on this?

kwmsmith:
@llllllllll sorry for the delay, getting through other backlogs.

Overall LGTM, thanks for fixing.

I'll be putting more time in on other PR and issue review to catch up.

PR #1385, opened 2016-01-12
Rename blaze.server.spider.spider to data_spider.

### LINKED ISSUE #1384: Issue running blaze-server in 0.9.0 release
```
$ python --version
Python 3.4.4 :: Continuum Analytics, Inc.

$ python -c 'import blaze; print(blaze.__version__)'
0.9.0

$ blaze-server ./foo.yaml
Traceback (most recent call last):
  File "/Users/ksmith/anaconda/envs/test-blaze-0.9.0/bin/blaze-server", line 6, in <module>
    sys.exit(blaze.server.spider._main())
AttributeError: 'function' object has no attribute '_main'
```

The best I can tell, this issue is related to the `import blaze.server.spider` line, and an ambiguity in this `import`.  There's both a `blaze/server/spider.py` module, and inside `blaze/server/__init__.py` is a `from .spider import spider`, which imports the spider *function* in the `blaze.server` namespace.

The strange thing is that this particular logic is unchanged since Blaze 0.8.3, and `blaze-server` worked fine then.

In any case, dealiasing the `spider` name in `blaze.server` is a good idea anyway to rule out this issue once and for all.  Fix forthcoming.

kwmsmith:
Disambiguates the blaze.server.spider module from the data_spider
function.

Closes #1384.

### DIFF
--- a/blaze/server/__init__.py
+++ b/blaze/server/__init__.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 from .server import Server, to_tree, from_tree, api
-from .spider import spider, from_yaml
+from .spider import data_spider, from_yaml
 from .client import Client
 from .serialization import (
     SerializationFormat,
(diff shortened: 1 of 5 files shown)


kwmsmith:
reviewed by @sandhujasmine and @postelrich.

PR #1393, opened 2016-01-19
BUG: sql truncate

llllllllll:
from the mailing list: https://groups.google.com/a/continuum.io/forum/#!topic/blaze-dev/0LxfZ73pUaQ

DatetimeTruncate was not supported for sql expressions. It was being compiled into an `extract` instead of a `date_trunc`

This also makes `unit_aliases` public for testing and adds some extra test utility functions.


llllllllll:
merging on pass

PR #1394, opened 2016-01-21
ENH: Adds 'ignore_existing' kwarg to create_index

llllllllll:
Allows users to optionally ignore existing indicies.

This also allows people to pass a tuple of column names instead of a list.


llllllllll:
ping, @kwmsmith any strong feelings on this?

PR #1398, opened 2016-01-26
BUG: revert a regression in the sql backend

llllllllll:
Leaves the test marked xfail so that we can fix this later.

pr that introduced the regression: #1396


llllllllll:
merging on pass

PR #1399, opened 2016-01-26
BUG: fix the weakkeydict key errors

llllllllll:
I have gotten a bunch of magical mysterious weakkeydict errors. This just uses a non data descriptor instead and replaces the `dshape` attribute on instance to get the same caching. This removes any strange cache issues and should be faster.

This change is taken from: #1348 but this particular bug is causing me a lot of problems so I am submitting it as its own change.

### DIFF
--- a/blaze/expr/expressions.py
+++ b/blaze/expr/expressions.py
@@ -12,16 +12,12 @@ import toolz
 from toolz import concat, memoize, partial, first
 from toolz.curried import map, filter
 
-from datashape import dshape, DataShape, Record, Var, Mono, Fixed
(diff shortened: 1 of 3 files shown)


kwmsmith:
+1: besides fixing the issue, it simplifies the code -- I'm always rusty on weak refs, and when combining weakkeydicts with caching / memoization, the complexity was making other bugfixes difficult.

LGTM.

llllllllll:
They, the only reason I normally prefer the weakkeydict approach is that the descriptor can be general because you don't need to know the name of the attribute.

PR #1409, opened 2016-02-09
coalesce and cast

llllllllll:
`cast` is needed when blaze is incorrect about the types of an expression. For example, the case I opened earlier about reductions of `var * A` being typed `A` instead of `?A` when the collection could be empty. To make it easy to work around this we should be able to explicitly overwrite the type.

`coalesce` is useful for a lot of queries.

### DIFF
--- a/blaze/compute/tests/test_postgresql_compute.py
+++ b/blaze/compute/tests/test_postgresql_compute.py
@@ -13,8 +13,26 @@ import pandas as pd
 
 import pandas.util.testing as tm
 
-from blaze import (atan2, by, compute, concat, cos, data, greatest, join,
-                   least, radians, sin, sqrt, symbol, transform)
+from datashape import dshape
+from odo import odo, drop, discover
+from odo.utils import tmpfile
+from blaze import (
+    data,
+    atan2,
+    by,
+    coalesce,
+    compute,
+    concat,
+    cos,
+    greatest,
+    join,
+    least,
+    radians,
+    sin,
+    sqrt,
+    symbol,
+    transform,
+)
 from blaze.interactive import iscorescalar
 from blaze.utils import example, normalize
 
@@ -593,7 +611,6 @@ def test_sample(sql):
     s2 = odo(result2, pd.DataFrame)
     assert len(s) == len(s2)
 
-
 @pytest.fixture
 def gl_data(sql_two_tables):
     u_data, t_data = sql_two_tables
@@ -609,3 +626,15 @@ def test_greatest(gl_data):
 def test_least(gl_data):
     u, t = gl_data
     assert odo(least(u.a.max(), t.a.max()), int) == 1
+
+
+def test_coalesce(sqla):
+    t = symbol('t', discover(sqla))
+    assert (
+        odo(compute(coalesce(t.B, -1), {t: sqla}), list) ==
+        [(1,), (1,), (-1,)]
+    )
+    assert (
+        odo(compute(coalesce(t.A, 'z'), {t: sqla}), list) ==
+        [('a',), ('z',), ('c',)]
+    )
--- a/blaze/compute/tests/test_postgresql_compute.py
+++ b/blaze/compute/tests/test_postgresql_compute.py
@@ -631,10 +631,10 @@ def test_least(gl_data):
 def test_coalesce(sqla):
     t = symbol('t', discover(sqla))
     assert (
-        odo(compute(coalesce(t.B, -1), {t: sqla}), list) ==
+        compute(coalesce(t.B, -1), {t: sqla}, return_type=list) ==
         [(1,), (1,), (-1,)]
     )
     assert (
-        odo(compute(coalesce(t.A, 'z'), {t: sqla}), list) ==
+        compute(coalesce(t.A, 'z'), {t: sqla}, return_type=list) ==
         [('a',), ('z',), ('c',)]
     )
--- a/blaze/expr/expressions.py
+++ b/blaze/expr/expressions.py
@@ -4,7 +4,17 @@ from keyword import iskeyword
 import re
 
 import datashape
-from datashape import dshape, DataShape, Record, Var, Mono, Fixed
+from datashape import (
+    dshape,
+    DataShape,
+    Record,
+    Var,
+    Mono,
+    Fixed,
+    promote,
+    Option,
+    Null,
+)
 from datashape.predicates import isscalar, iscollection, isboolean, isrecord
 import numpy as np
 from odo.utils import copydoc
@@ -16,12 +26,14 @@ from ..compatibility import _strtypes, builtins, boundmethod, PY2
 from .core import Node, subs, common_subexpression, path
 from .method_dispatch import select_functions
 from ..dispatch import dispatch
-from .utils import hashable_index, replace_slices
+from .utils import hashable_index, replace_slices, maxshape
 from ..utils import attribute
 
 
 __all__ = [
     'Apply',
+    'Cast',
+    'Coalesce',
     'Coerce',
     'ElemWise',
     'Expr',
@@ -35,6 +47,8 @@ __all__ = [
     'Slice',
     'Symbol',
     'apply',
+    'cast',
+    'coalesce',
     'coerce',
     'discover',
     'label',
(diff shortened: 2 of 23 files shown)


llllllllll:
ping @kwmsmith I would like someone to look this one over because I am adding some new expressions.

llllllllll:
freshly rebased against master

llllllllll:
@kwmsmith I would like to merge this soon

llllllllll:
If you don't have any comments I plan on merging by the end of tomorrow

blaze/expr/expressions.py
```diff
@@ -790,6 +804,92 @@ def coerce(expr, to):
     return Coerce(expr, dshape(to) if isinstance(to, _strtypes) else to)
 
 
+class Cast(Expr):
+    """Cast an expression to a different type.
+
+    This is only an expression time operation.
```
kwmsmith: Can you provide a brief `Examples` section?

blaze/expr/expressions.py
```diff
@@ -790,6 +804,92 @@ def coerce(expr, to):
     return Coerce(expr, dshape(to) if isinstance(to, _strtypes) else to)
 
 
+class Cast(Expr):
```
kwmsmith: We'll now have both `cast` and `coerce` expressions with this.  Is the essential that `cast` is for the expression's dshape, while `coerce` is for the underlying backend?

blaze/expr/expressions.py
```diff
@@ -790,6 +804,92 @@ def coerce(expr, to):
...
+        yield expr.rhs
+
+
+class Coalesce(Expr):
+    """SQL like coalesce.
+
+    coalesce(a, b) = {
+        a if a is not NULL
+        b otherwise
+    }
```
kwmsmith: A brief `Examples` section would be very useful here as well.

blaze/expr/expressions.py
```diff
@@ -790,6 +804,92 @@ def coerce(expr, to):
...
+    """
+    __slots__ = '_hash', 'lhs', 'rhs', 'dshape'
+    __inputs__ = 'lhs', 'rhs'
```
kwmsmith: Is `Coalesce` always binary?  Wouldn't it make sense to be variadic?  Or is it variadic in SQL and we can't express variadic things in Blaze's multiple dispatch, so we're limited to binary?

llllllllll: This is variadic in sql but we don't yet have a good way to handle this in blaze. When we do support that this would be a good candidate for a variadic (and then concat)

blaze/expr/expressions.py
llllllllll: Yeha, coerce will actually affect the data, cast will just reinterperate the data

blaze/expr/expressions.py
```diff
@@ -845,6 +945,7 @@ def ndim(expr):
 schema_method_list.extend([
     (isscalar, set([label, relabel, coerce])),
     (isrecord, set([relabel])),
+    (lambda ds: isinstance(ds, Option), {coalesce}),
```
kwmsmith: `coalesce` is valid for any `isscalar` dshape, right? Not just `Option` types?  Or am I misunderstanding coalesce?

blaze/expr/tests/test_expr.py
```diff
@@ -187,3 +188,69 @@ def test_pickle_roundtrip():
...
+
+    # check optional lhs and rhs with promotion
+    y = symbol('y', '?int64')
+    expr = coalesce(v, y)
+    # rhs and lhs are optional so this might be null
+    # there are no either types in datashape so we are a type large enough
+    # to hold either result
+    assert_dshape_equal(expr.dshape, dshape('?int64'))
+    assert expr.lhs.isidentical(v)
+    assert expr.rhs.isidentical(y)
```
kwmsmith: Some error / edge cases for testing:
* non-scalar dshapes: `Record`, others?;
* non-integer dshapes: `string` comes to mind;
* non-atomic expression arguments with compatible dshapes.

blaze/expr/expressions.py
llllllllll: it is not meaningful to coalesce a non-optional type with anything because it is just the identity.

docs/source/whatsnew/0.10.0.txt
```diff
@@ -10,6 +10,17 @@ New Expressions
...
+* Adds :func:`~blaze.expr.expressions.coalesce` expression which takes two
+  arguments and returns the first non missing value. If both are missing then
+  the result is missing. For example: ``coalesce(1, 2) == 1``,
+  ``coalesce(None, 1)`` == 1, and ``coalesce(None, None) == None``.
+  This is inspired by the sql function of the same name (:issue:`1409`).
+* Adds :func:`~blaze.expr.expressions.cast` expression as a type level function
+  for reinterpreting an expression as another dshape. This is based on C++
+  ``reinterpret_cast``, or just normal C casts. For example:
+  ``symbol('s', 'int32').cast('uint32').dshape == dshape('uint32')``. This
+  expression is the identity function at runtime but allows users to alter the
```
kwmsmith: "is the identity function at runtime" -- can you expand on this?  Do you mean it has no effect on the execution of `compute` on the expression?

docs/source/whatsnew/0.10.0.txt
```diff
@@ -10,6 +10,17 @@ New Expressions
 * The ``sample`` expression allows random sampling of rows to facilitate
   interactive data exploration (:issue:`1410`).  It is implemented for the
   Pandas, Dask, SQL, and Python backends.
+* Adds :func:`~blaze.expr.expressions.coalesce` expression which takes two
+  arguments and returns the first non missing value. If both are missing then
+  the result is missing. For example: ``coalesce(1, 2) == 1``,
+  ``coalesce(None, 1)`` == 1, and ``coalesce(None, None) == None``.
+  This is inspired by the sql function of the same name (:issue:`1409`).
+* Adds :func:`~blaze.expr.expressions.cast` expression as a type level function
+  for reinterpreting an expression as another dshape. This is based on C++
```
kwmsmith: Perhaps "Adds the :func:`~blaze.expr.expressions.cast` expression to reinterpret an expression's dshape"?

docs/source/whatsnew/0.10.0.txt
llllllllll: I can put this in the docs, but cast is defined as

```python
@dispatch(Expr, object)
    return data
```

kwmsmith:
@llllllllll Apologies for taking so long to get to this -- I've been pinned down on a number of internal things lately.

This looks good to me; my only question is on the names; new users may be confused on the distinction between `cast` and `coerce`, and `coalesce` isn't obvious at first what it does unless you're coming from SQL :)

Because of the possibly confusing or non-obvious names, it's important that we have good docstrings with good examples.

llllllllll:
No worries, I just wanted to make sure didn't get forgotten. I completely agree about the naming and documentation issue. I will go through and try to add some good examples and maybe some prose docs about the difference between cast and coerce. If you have an alternative name for `cast` please let me know. This operation is not at all the same as sql cast which can change values, this is really more a C cast, but idk if our users will have the same intuition for what a C cast is vs a sql cast.

kwmsmith:
> If you have an alternative name for cast please let me know.

Taking a page from Numpy / pandas, which have `astype()`: what about `as_dshape` instead of `cast`?  I'm trying to capture the idea that we're just changing the dshape of the expression, not the underlying data.

llllllllll:
numpy astype actually does conversions though, the equivalent in numpy is `view`

kwmsmith:
For `cast`, what happens if the user casts a dataframe with integer column to a dataframe with float column?

```python
bz.compute(df.int_column.cast('float32'))
```

Since `cast` is a no-op at compute time, the result here would be an integer pandas series, correct?  So in that sense, it isn't analogous to Numpy's `view`, since the `cast` is completely ignored at compute time.  I don't think I can come up with any Numpy / Pandas analogue for `cast` here, can you?

llllllllll:
I guess it is closer to view which is a C style cast. Really this is just a type level function for allowing us to use different expressions when blaze gets the type wrong. The use case I have for this is when there are bugs in the blaze type system I need to be able to fix them in my app code before I have time to fix them in blaze. For example, reductions over var returning A instead of ?A

kwmsmith:
Yes, I totally understand the usecase for correcting blaze / datashape's option / non option types.  That makes sense.  It's for the other cases when the user tries to cast from integer to floats, for example, and it has no effect in the computed results that gets more confusing.

Is dealing with option types the only usecase here, or are there others?  Because my preference would be to keep this limited to handling casts between option / non-option types of the same underlying type for now, and only expand it to more general casts if there's a clear usecase for that.

llllllllll:
The use case is for any time that the blaze type system is incorrect. This was happening before with reductions on timedeltas giving back floats. This was corrected but but I would like to have the freedom to work around this class of issue in the future.

kwmsmith:
I'm OK with this for the option / non-option case, of course, and I might envision other cases as well (`string` & `string[10]`, for example).  But if it turns out users misuse `cast` frequently and it leads to confusion--the `int` to `float` cast, for example, and the user expecting some change in the result--then we may need to modify the types `cast` is allowed to cast between.

My preference would be to only allow `cast` for the clear pain points, and expand it as other cases become clear in the future.

I'm OK with merging this as long as we're able to restrict the cases for `cast` in the future if it becomes a place where users get frequently tripped up.

blaze/expr/expressions.py
llllllllll: done

blaze/expr/expressions.py
llllllllll: done

llllllllll:
Added more tests, including some xfail tests around promotion of mixed shaped records which I think should work in the future

blaze/compute/tests/test_postgresql_compute.py
```diff
@@ -609,3 +626,15 @@ def test_greatest(gl_data):
 def test_least(gl_data):
     u, t = gl_data
     assert odo(least(u.a.max(), t.a.max()), int) == 1
+
+
+def test_coalesce(sqla):
+    t = symbol('t', discover(sqla))
+    assert (
+        odo(compute(coalesce(t.B, -1), {t: sqla}), list) ==
```
kwmsmith: nit: for consistency, we've been using `compute(expr, sql_data, return_type=list)` for simple cases like this rather than an explicit `odo` call.

blaze/compute/tests/test_postgresql_compute.py
```diff
@@ -609,3 +626,15 @@ def test_greatest(gl_data):
...
+        [(1,), (1,), (-1,)]
+    )
+    assert (
+        odo(compute(coalesce(t.A, 'z'), {t: sqla}), list) ==
```
kwmsmith: ibid.

kwmsmith:
@llllllllll just some consistency nits, otherwise +1.

As I mentioned above, in the future, we may need to restrict `cast`'s applicability if we find that non power-users misuse it frequently.  And if we find a better name for `cast`, then I'm fine with a rename in the future.

Also, please make a separate issue to add sections for `coerce` and `cast` in the blaze docs.

llllllllll:
I like the `return_type` use here, I think it reads nicely. I added these exprs to the API section, but I will add an issue to write actual docs for this

kwmsmith:
LGTM.

llllllllll:
Thanks for looking at this

llllllllll:
merged on cli

PR #1410, opened 2016-02-09
Random sampling

kwmsmith:
Mimics Pandas' `df.sample()` interface.

The main usecase is to randomly downsample to a cached dataset, and then use that as the basis for faster exploration and expression building.

So far, the Python and SQL backends are implemented; still have to implement the Pandas / Dask backend.


blaze/compute/sql.py
```diff
@@ -842,6 +842,24 @@ def compute_up(t, s, **kwargs):
     return s.order_by(*cols)
 
 
+@dispatch(Sample, sa.Table)
+def compute_up(t, s, **kwargs):
+    n = t.n if t.n is not None else int(t.frac * s.count().scalar())
```
llllllllll: This is going to actually execute a sql query at this point when you call `.scalar()`. Can we look into pushing this into a subselect to reduce the IO overhead. Also, should we special case postgres>=9.5 wich has the TABLESAMPLE clause on selects?

blaze/expr/collections.py
```diff
@@ -235,6 +250,47 @@ def tail(child, n=10):
...
+
+
+@copydoc(Sample)
+def sample(child, n=None, frac=None):
+    if n is frac is None:
+        raise TypeError("sample() missing 1 required argument, 'n' or 'frac'.")
+    if n is not None and frac is not None:
+        raise ValueError("n ({}) and frac ({}) cannot both be specified.".format(n, frac))
+    if n is not None:
+        n = int(n)
```
llllllllll: Should we maybe use `operator.index` to reject things like `float` and `Decimal`?

blaze/expr/collections.py
```diff
@@ -235,6 +250,47 @@ def tail(child, n=10):
...
+        if n < 1:
+            raise ValueError("n must be positive, given {}".format(n))
+    if frac is not None:
+        frac = float(frac)
+        if frac < 0.0 or frac > 1.0:
```
llllllllll: `if not 0.0 >= frac >= 1.0:` might be clearer.

blaze/expr/collections.py
```diff
@@ -235,6 +250,47 @@ def tail(child, n=10):
...
+            raise ValueError("0 < frac < 1.0, given {}".format(frac))
```
llllllllll: could we prefix this with `sample requires` so it is clear what the lhs is showing. Also I think that those should be inclusive bounds.

blaze/compute/sql.py
kwmsmith: Yes, there's a lot of room here for performance improvements.  I'll look into the subselect.  Related questions:  is sqla sophisticated enough to translate `s.count()` into an O(1) operation if the underlying table is indexed?  Is this dialect / engine specific?  Or does sqla take the safe approach an always do a table scan regardless?

blaze/expr/collections.py
kwmsmith: Yes, and good to know about `op.index`, thanks for pointing it out.

blaze/expr/collections.py
kwmsmith: Agreed.

blaze/expr/collections.py
kwmsmith: Agreed.

blaze/compute/sql.py
llllllllll: I don't think sqlalchemy is going to decide any of that. The way to check this would be to output the sql query (you can use `blaze.utils.literal_compile` for this to get any bind params formatted in) and then feed it to `explain` in your db. This will show the query plan which should include any index use

blaze/compute/python.py
```diff
@@ -666,6 +667,12 @@ def compute_up(t, seq, **kwargs):
         return take(t.n, seq)
 
 
+@dispatch(Sample, Sequence)
+def compute_up(t, seq, **kwargs):
+    nsamp = t.n if t.n is not None else int(t.frac * len(seq))
+    return random.sample(seq, min(nsamp, len(seq)))
```
llllllllll: I wonder if an optimization might be to just be identity if `n >= len(seq)`. is the contract that `sample` is in random order or just that it is a random sampling?

blaze/compute/tests/test_sql_compute.py
```diff
@@ -517,6 +517,10 @@ def test_head():
     assert str(compute(t.head(2), s)) == str(select(s).limit(2))
 
 
+def test_sample():
+    assert str(compute(t.sample(n=5), s)) == str(select(s).order_by(sa.func.random()).limit(5))
```
llllllllll: should we have a test with `frac` also?

blaze/compute/python.py
kwmsmith: I'm fine with this as-is, since both Pandas' `sample` and `random.sample` both implement random order when `n == len(seq)`, and it provides a way for users to randomize the order of all rows if that's desired.

llllllllll: Okay, that sounds good. I just wanted to make sure this was explicitly chosen

Issue #1411, opened 2016-02-11
compute on sql returns selectable


postelrich:
breaking out from #1401 

    import blaze as bz
    d = bz.Data('sqlite:///%s::iris' % bz.utils.example('iris.db'))
    bz.compute(d.sepal_width.sum())
    
Returns a sqlalchemy select object instead of executing the sql.  This is inconsistent with the expectation of compute.

Issue #1401, opened 2016-01-27
`compute` does not compute


kwmsmith:
I'm breaking this topic out into its own issue to try and keep things topical.  It has overlap with issues #1357 and #1304.

## Proposed improvements to `compute()` API

For reasons described below, we propose the following changes to `compute()`:

* `compute()` always see computation through to completion, unless instructed otherwise by a default argument override. Notably, this would change the behavior of `compute()` for SQL-backed computations.
* We propose that, for table-like results, `compute()` by default always returns a Pandas dataframe or Dask dataframe for results that will fit in memory (for some modifiable heuristic for "what fits in memory").
* For tabular results that don't fit in memory, my initial preference is to fall back to `odo`'s failure mode or raise a `NotImplementedError` asking the user to reduce / summarize the computation further.  
* The new `compute()` can be instructed to return the native, backend-specific result (like a SQLA select object) via overriding a `return_native=True` or `return_type='native'` default argument at call time (I'm open to better names here).
* For array-like things, `compute()` returns a NumPy array as its default for in-memory results, and raises or returns a `dask.array` for OOC results.  This behavior is similarly overrideable.

## Reasoning and usecases

Currently, Blaze's `compute()` returns different types of results depending on the backend.  For instance, for a Pandas-backed dataframe `a`:

```python
a = bz.Data(pd.DataFrame(...))
bz.compute(a.col.sum())
```

this code actually triggers a Pandas computation, and yields a scalar.

If we use the same logical data but use a SQL DB, the results are significantly different:

```python
db = bz.Data('postgresql://...')
bz.compute(db.col.sum())
```

This will result in a `sqlalchemy.sql.selectable.Select` object.  It is not obvious without searching in the documentation that the "right way" to see the computation to completion is to use `odo`:

```python
db = bz.Data('postgresql://...')
bz.odo(bz.compute(db.col.sum()), float)
```

The current behavior has led to significant confusion and lost time on the part of new users, especially for new users who know nothing of SQL and who are specifically using Blaze as a way to abstract away SQL entirely.

The benefits to these proposed changes to `compute()` are several:

* Blaze's user-facing API is more consistent across backends.  Pandas and SQL backed computations always result in a (fully realized) Pandas dataframe by default when that result can fit in memory, and results in a `dask.dataframe` otherwise (and similarly for array-based computations with NumPy / `dask.array` as the default).
[...]

Issue #1358, opened 2015-12-11
Data -> data and other API inconsistencies


kwmsmith:
For historical reasons as I understand it, `Data` was originally a class / type in Blaze.  It has since moved to a function.  Nearly everything else in the end-user API is `lowercase_with_underscores`, and there's seemingly no reason besides backwards compatibility for `Data` to be capitalized.

This fix would provide a `data` callable that fills the same role as `Data`.  `Data` would be reimplemented to raise a deprecation warning for 1 minor release cycle, after which it will raise an exception pointing users to `data`, and then it will be removed entirely.

This is related to #1357.

kwmsmith:
Along with this, I vote to deprecate `Symbol` in favor of `symbol`.  Again, once #1357 creates `UnboundSymbol`s, the `Symbol` constructor is no longer needed.

If there are other renamings that we should do, please speak up here.

llllllllll:
I think `Symbol` is already for internal use just the the other expr classes. Users should go through `symbol` for the memoization. One issue is that we import * in init so we expose far to many attributes through our public api.

cpcloud:
`Symbol` and `symbol` are not synonymous. `Symbol` is the type of thing that `symbol` returns, and `symbol` hides some caching behind a function.

kwmsmith:
> Symbol and symbol are not synonymous. Symbol is the type of thing that symbol returns, and symbol hides some caching behind a function.

Right, in #1357 we're proposing renaming the `Symbol` type to `UnboundSymbol`.  With that renaming, I see no reason not to deprecate `Symbol` and point everyone to use `symbol` instead.

kwmsmith:
> I think Symbol is already for internal use just the the other expr classes. Users should go through symbol for the memoization. One issue is that we import * in init so we expose far to many attributes through our public api.

Good point, I'll take a look through our public API and see what we can / should tighten up.

g24swint:
I am super new at using blaze (as in, played with it for only a couple of days), but I like this change as it was puzzling that I didn't end up with a Data after calling Data().

kwmsmith:
Thanks for the input @g24swint.

kwmsmith:
@postelrich  Here's the issue we discussed, should be a quick one.

PR #1412, opened 2016-02-11
#1411 add return_type logic to compute

postelrich:


llllllllll:
Currently the behavior of having `compute` return selects is very important for many applications. This allows users to seamlessly weave blaze into places where sqlalchemy is expected. Until there is a plan to replace this with a `compile` function this will remove an imporant piece of functionality from the blaze backend. The standard way to materialize an expression is not `compute` but instead it is `odo`.

postelrich:
@llllllllll this is an effort to standardize what is returned by compute.  Right now, to my knowledge, the sql abstraction is the only one to not compute as expected.  It causes confusion to users because the implied behavior is to not need to odo the results of compute.

postelrich:
See #1401 for more info

llllllllll:
The sql backend abstraction matches all of the other backends for compute. If I feed compute pandas structures I will get out a pandas structure. If I feed it numpy I get numpy out. It follows that if I feed compute sqlalchemy structures I should get sqlalchemy structures back out.

I think it is fair to say that this function is better named `compile` instead of compute. I do not think that adding a post compute is the correct way to solve this issue because it makes it much harder to weave blaze and sqlalchemy. I think we need to hold off on this until we have a clear dperecation path for the current compute verbs and give our users time to port all of code that depends on this old behavior.

kwmsmith:
@llllllllll see #1401, specifically:

> By providing an overrideable optional argument, users who, for example, want the sqlalchemy object for use in other SQLA code can easily get what they need.

We're changing the default behavior to be more consistent across backends; this decision is made based on the input from many users who are consistently bitten by `compute`'s current inconsistency with SQL and who have no idea what to do with a SQLA Select object.

We realize this is a breaking change; but as you know, to get to the behavior you want, you just have to set `post_compute=None` when calling `compute`.  That should handle your usecase--are we wrong there?

blaze/compute/sql.py
```diff
@@ -1181,7 +1181,11 @@ def compute_up(expr, data, **kwargs):
 
 
 @dispatch(Expr, ClauseElement)
-def post_compute(_, s, **kwargs):
+def post_compute(expr, s, **kwargs):
+    if isscalar(expr.dshape):
+        return select(s).scalar()
```
kwmsmith: @postelrich this is a good start; it'd be better to leverage `odo.into` here, since it can handle more cases.

```python
return odo(s, <scalar dtype from dshape>)
```

postelrich:
@kwmsmith actually no, setting `post_compute=None` doesn't negate the behavior.  The post_compute includes the code that wraps it in a `select`.  

```
In [1]: import blaze as bz

In [2]: x = bz.Data('sqlite:///%s::iris' % bz.utils.example('iris.db'))

In [3]: bz.compute(x.petal_width.sum(), post_compute=False)
Out[3]: <sqlalchemy.sql.elements.Label object at 0x118358ba8>

In [4]: print(bz.compute(x.petal_width.sum(), post_compute=False))
sum(iris.petal_width)
```

blaze/compute/sql.py
```diff
@@ -1181,7 +1181,11 @@ def compute_up(expr, data, **kwargs):
...
+    elif iscollection(expr.dshape) or istabular(expr.dshape):
+        return select(s).execute().fetchall()
```
kwmsmith: Similarly, we should do:

```python
return odo.into(s, pd.DataFrame)
```

The `fetchall()` will return a nested Python datastructure, not a Pandas object.

It'd be even better if we could detect when the SQLA object is a single column and `odo` it into a `pd.Series`

llllllllll:
I think that the `post_compute` should take a `materialize` kwarg that maybe can default to true. The issue is that we have post compute registered for `ClauseElement` to wrap it in a `select`. We also unbox up `ScalarSelect`. The boxing still needs to happen for these cases. If we just made this a seperate argument these 3 post_compute functions can decide what to return. I am also a little concerned about how this affect the odo graph. Right now anything that is a sqlalchemy object -> other will not be used because we will be computing it out into a python list of tuples (for tables). This will mean that we cannot fast path from blaze expr against sql into some other structure, we always need to go to list first.

postelrich:
@kwmsmith , also does it make sense to have this return the native result from a sqlalchemy query?  For collections, it returns a list of tuples.  Does that make sense or should it be odo'ed to something else?

blaze/compute/sql.py
postelrich: this answers the question I just posted below.

kwmsmith:
@llllllllll 

> he issue is that we have post compute registered for ClauseElement to wrap it in a select. We also unbox up ScalarSelect

Both the `ClauseElement` and `ScalarSelect`  `post_compute` dispatches are trivial: is it unreasonable for someone wanting the SQLA result to detect these cases and wrap in a select or unbox depending?  That would be my preference, and that way all SQL `post_compute` implementations can return scalars / dataframes / series.

> I am also a little concerned about how this affect the odo graph. Right now anything that is a sqlalchemy object -> other will not be used because we will be computing it out into a python list of tuples (for tables). This will mean that we cannot fast path from blaze expr against sql into some other structure, we always need to go to list first.

Agreed, we will use `odo.into` to return a scalar or a Pandas object, depending.  We won't be using `.fetchall()`.

kwmsmith:
Strange, why didn't Travis-CI trigger for 19da248?

postelrich:
Are we okay with the latest change?  Good to start updating the tests?

llllllllll:
Should we have a releaase that has a deprecation warning on the old behavior to give users time to migrate?

kwmsmith:
> Are we okay with the latest change? Good to start updating the tests?

@postelrich  yes, start updating the tests.

kwmsmith:
> Should we have a releaase that has a deprecation warning on the old behavior to give users time to migrate?

@llllllllll in general, yes, I agree.  To be clear, you mean 1 full release cycle that keeps the old behavior and issues a deprecation warning, and then the next release switches over to the new behavior?

llllllllll:
@kwmsmith Correct, 0.10 should have the old behavior and optionally the new behavior with a flag. Then in the next release we continue to support the flag being true only and but raise a deprecation warning if the flag is set. Finally we can remove the flag in the release after that.

postelrich:
@llllllllll also want to point out that the other part of #1401 includes allowing the user to select the format the data is returned in.  It includes a `native` format.  Not sure if it would make sense for it to return the `Select` as the native format for the sql module.

postelrich:
@llllllllll @kwmsmith Since this isn't even at a 1.0 yet, would going through deprecation be overkill?  The fact that sql is the only abstraction that isn't consistent with compute seems more like just fixing a bug than changing a feature.

kwmsmith:
> Since this isn't even at a 1.0 yet, would going through deprecation be overkill? The fact that sql is the only abstraction that isn't consistent with compute seems more like just fixing a bug than changing a feature.

Going through a full 3-release cycle deprecation process is overkill, yes, given that Blaze doesn't have that many users.

For big changes like this, I do want to have a 2 release cycle deprecation process, though.  We'll keep the old behavior around with a warning for the 0.10 release, and then switch over to the new behavior completely for the 0.11 release.

So @postelrich, after you get the tests working with the new behavior, please add a branch in `post_compute` to keep the old behavoir, and have it take this branch by default.

We're also doing more frequent (monthly at least) Blaze releases now that the release process has been streamlined, so it won't be long before the new behavior is fully implemented.

The 0.10 release will have several API deprecations, I'm hoping to get as many as we can in place at once.

kwmsmith:
> also want to point out that the other part of #1401 includes allowing the user to select the format the data is returned in. It includes a native format. Not sure if it would make sense for it to return the Select as the native format for the sql module.

@llllllllll @postelrich what are your thoughts on the `return_type` optional argument, as described in #1401?  It's syntactic sugar, since we can accomplish what we want with `post_compute=None` to return the un-odo-ed result, but it's clearer to new users.

llllllllll:
I understand and agree with the motivations in #1401 and I am glad we are taking steps to get there; however, the reason I am being sort of grumpy about this is that it makes one workflow, namely the one that some of my applications are built on, harder.

> @llllllllll also want to point out that the other part of #1401 includes allowing the user to select the format the data is returned in.

I see this mentioned in #1401 but this pr does not implement it. This change seems like it will make it difficult to get the `native` result because it will need to properly pass post_compute when using sqlalchemy. I imagined that this would be a change that live in `compute` itself, not in the post_compute for a particular backend.

> The fact that sql is the only abstraction that isn't consistent with compute seems more like just fixing a bug than changing a feature.

I tried to address this already. In blaze, the output of compute is always in the same backend as the input. In pandas, that means that dataframes and series are the inputs and output. In sqlalchemy, tables and selects are the input and output. I think that this is very consistent with the model that blaze has. I see this is a change in the definition of compute from "compute this expression in this backend" to "compute *and materialize* this expression in this backend". What does the dask backend do here? I think that has a similar experience where the 'values' are actually deferred expressions also.

The reason I asked for a release cycle to deal with this is that I depend on the old sql behavior and I am also using blaze master (or very close to it) in a lot of cases. In order to stay current I need to update all of my existing code and I need some time to do that.

Thank you again for contributing this. I do appreciate that you are trying to help us move forward, I just need to defend my use case ;)

llllllllll:
I like having a `return_backend` field. This means I can easily go from any backend to say, pandas in one consistent step. `post_compute=None` also has some slight differences because it will sometimes return non-selectables. I think that we could move the coercion logic from post_compute to compute after any given post_compute is called. This means I can provide my own post_compute and _still_ coerce to pandas after.

kwmsmith:
> In blaze, the output of compute is always in the same backend as the input. In pandas, that means that dataframes and series are the inputs and output. In sqlalchemy, tables and selects are the input and output. I think that this is very consistent with the model that blaze has. I see this is a change in the definition of compute from "compute this expression in this backend" to "compute and materialize this expression in this backend". 

A couple of comments here:
* If you look at the Mongo backend, it has a couple of [`post_compute()` implementations](https://github.com/blaze/blaze/blob/master/blaze/compute/mongo.py#L400-L442) that take the Mongo query and convert it to nested Python data structures.  This is exactly the spirit of what we're trying to do here: once the SQLA query is built inside `compute`, ensure it's run to completion and gives back a core backend object--in this case, a Pandas DataFrame.
* For the longer term, we're pushing blaze in a heterogeneous direction, supporting heterogeneous joins when we can, etc.  This immediately raises the question of "which backend should the result be in?"  For table-like things, we'll always default to Pandas / dask.dataframe; for array-like things, Numpy / dask.array, for NoSQL things, Python data structures, etc.  This lets us have a consistent story.
* We definitely want to support escape hatches so that things like your usecase can get the SQLA query directly (or Mongo query, etc.).

> What does the dask backend do here? I think that has a similar experience where the 'values' are actually deferred expressions also.

Agreed, but that's OK, since dask is a core backend with a Numpy / pandas API (dask.array, dask.dataframe), and end users get something with a familiar API after `compute`ing.  Our focus here is to abstract over backends that users don't want to deal with, with escape hatches for more advanced cases.

kwmsmith:
> I like having a return_backend field. This means I can easily go from any backend to say, pandas in one consistent step. post_compute=None also has some slight differences because it will sometimes return non-selectables. I think that we could move the coercion logic from post_compute to compute after any given post_compute is called. This means I can provide my own post_compute and still coerce to pandas after.

You make a good case, I agree.  Thoughts on `return_backend` vs. `return_type` vs. a different name?  I'm partial to `return_type`.  Perhaps `odo_type`?

What values would this argument take, with what meanings?  Here's one proposal:
* `None` -- the current default; returns whatever backend type `compute` provides.
* `'core'` -- if tabular, odos to a DataFrame / dask.dataframe if necessary; if array-like, odos to Numpy / dask.array, etc.  Will become the default after version 0.10.
* a specific type, like `pd.DataFrame`, `dask.dataframe`, `dask.bag`, `list`, `Temp(CSV)` etc -- passed through to the `odo` step to allow precise control.

@postelrich sorry for the churn, but could you add in a `return_type` kwarg (name subject to change) that defaults to `None` to [this implementation](https://github.com/blaze/blaze/blob/master/blaze/compute/core.py#L369-L370), and checks for the other cases listed here?  Also, you'll need to revert the changes to `post_compute`.

llllllllll:
I would prefer `return_backend` or just `backend` which can be a string. This gives us the duck-typing of `compute` with the control I think most users require. If you know the exact type you want, you should use:

```python
odo(expr, type)
```

instead of `compute`.

The argument could be a string that is registered to a `materialze` function. This function will be passed the computed data the datashape of the result, and any `odo_kwargs` needed. The implementation would be very similar (if not the same) to https://github.com/blaze/blaze/pull/1413/files#diff-72e112ee086443eeb30ba60beae284beR57. In that PR different serialization formats needed control over materializing of expressions on the server in the same way that all computations really want to control this. Maybe we could merge these and make the `SerializationFormat` just accept a blaze backend string. We could use a `RegexDispatcher` so that users can register their own materialization formats with:

```python
@backend.register('pandas')
def materialize_pandas(data, dshape, odo_kwargs):
    if istabular(dshape):
        tp = pd.DataFrame
    elif iscollection(dshape):
        tp = pd.Series
    else:
        return coerce_scalar(data, str(dshape), odo_kwargs)

    return odo(data, tp, **odo_kwargs)
```

This is a moderate deviation from what is proposed in this pr but I think it satisfies all of our use cases. What do you all thing?

postelrich:
@kwmsmith I think the best method would be a large refactor where all the current compute functions get renamed to compile.  Then you have one compute function which calls compile and then odos depending on user input.  The default odo would be to a pandas/numpy series,dataframe,array as appropriate.

kwmsmith:
> I would prefer return_backend or just backend which can be a string.

Thinking about it over the weekend -- `compute(expr, sql_data, backend='pandas')` implies to me that the *compute* backend is being specified, which is misleading, since all computation is in SQL and then the result is `odo`'d to Pandas afterwards.

> If you know the exact type you want, you should use: `odo(expr, type)` instead of compute.

This is a great point and one that I forgot in the course of this discussion, thanks for bringing it up.

`odo(expr, type, **kwargs)` provides a way for more advanced usecases when odoing requires extra kwargs, and lets us keep the default behavior streamlined.

My gut sense is that the `materialize` part of your proposal @llllllllll is too heavyweight for now and is out of scope.

I'd much rather start with a simple thing and see if the extra complexity of another multipledispatch is warranted.

At the core, what we're after is:

```python
compute(expr, data)
```

Will always return a:
* dataframe (either Pandas or Dask, depending on the particulars) for table-like exprs,
* ndarray (or dask.array) for array-like exprs,
* Python collection for unstructured things (like a general NoSQL query result), or
* a Python scalar, as the expression dictates.

There is an optional argument `return_type` (understanding that the name of this arg is not settled yet) that has the default setting of `'core'`, which means `compute` will return a core type.

If someone wants `compute` to give back the internal SA query instead of a dataframe, then the current proposal on the table would be to override `return_type` (again, understanding we haven't finalized on the name for the kwarg yet):

```python
compute(expr, data, return_type='native')
```

To keep the `compute` API streamlined for now, we specifcally don't provide a way for users to add more odo kwargs.  Inside the `compute` top-level implementation, we only have to check for a few values of `return_type`, which keeps that logic somewhat straightforward as well.

If someone passes in an actual type for `return_type` instead of a string, then that type could be passed on to `compute`'s internal `odo` call directly.  This clearly won't work for the cases when extra kwargs need to be passed in to `odo`, for that case, see below.

This is the 80% solution, and we specifically want to keep it simple for now.

For the remaining 20% of usecases, to take control over the final conversion and  specify more odo arguments, then they'd use:

```python
odo(expr, OdoSupportedType, data=<symbol: data mapping needed by compute>, odo_arg1=..., odo_arg2=...)
```

The last version here would be the fully general case, intended for users with specific needs.

@postelrich @llllllllll  I'll implement a prototype of this implementation on top of this PR so we can kick the tires.

postelrich:
implemented Kurt's POC.  

Note that the iscollection is only checking to see if the dimension > 1 and istabular requires the dimension == 2.  Can pandas DataFrames handle multidimensional datasets?  If so I need to change the way the type checking is done.

blaze/compute/core.py
```diff
@@ -367,8 +371,14 @@ def swap_resources_into_scope(expr, scope):
 
 
 @dispatch(Expr, dict)
-def compute(expr, d, **kwargs):
-    """ Compute expression against data sources
+def compute(expr, d, return_type='native', **kwargs):
+    """ Compute expression against data sources.
+
+    Args:
```
kwmsmith: Could you edit the docstring to fit with the other docstring styles?  Like [join, for example ](https://github.com/blaze/blaze/blob/master/blaze/expr/collections.py#L409-L446).

blaze/compute/core.py
```diff
@@ -367,8 +371,14 @@ def swap_resources_into_scope(expr, scope):
...
+        expr: blaze expression
+        d: data source to compute expression on
+        return_type (optional): type to return data as. Defaults to 'core' which
```
kwmsmith: For this release `return_type` defaults to `'native'`.  We should also note in the docstring that we'll be changing it to `'core'` in version 0.11.

blaze/compute/core.py
```diff
@@ -367,8 +371,14 @@ def swap_resources_into_scope(expr, scope):
...
+                                returns as scalar, list, series, or dataframe as appropriate.
```
kwmsmith: Can you reflow these and make sure the line lengths aren't too long?

blaze/compute/core.py
kwmsmith: We'll also want to describe what the `'core'` and `'native'` options do and why they're different.

blaze/compute/core.py
```diff
@@ -407,6 +417,18 @@ def compute(expr, d, **kwargs):
     if post_compute_:
         result = post_compute_(expr3, result, scope=d4)
 
+    # return the backend's native response
+    if return_type == 'native':
+        warnings.warn("compute's `return_type` parameter will default to 'core' in blaze version >= 0.11.", BlazePendingDeprecationWarning)
```
kwmsmith: Could you format this message a bit more, and break it into 2 lines?

blaze/compute/core.py
kwmsmith: One more: we should note that the user can pass in a `type` explicitly as a third option besides the `'core'` and `'native'` string options.

blaze/interactive.py
```diff
@@ -252,6 +253,28 @@ def coerce_scalar(result, dshape, odo_kwargs=None):
         return result
 
 
+def coerce_core(result, dshape, odo_kwargs=None):
+    """Coerce data to a core data type."""
+    if iscoretype(result):
+        pass
```
kwmsmith: `pass` -> `return result` here.

blaze/utils.py
```diff
@@ -310,3 +310,13 @@ def __get__(self, instance, owner):
...
 
         return self._f(instance)
+
+
+class BlazeDeprecationWarning(DeprecationWarning):
+    """Blaze specific deprecation warning."""
+    pass
+
+
+class BlazePendingDeprecationWarning(PendingDeprecationWarning):
```
kwmsmith: As discussed, let's remove this subclass and just use the `BlazeDeprecationWarning` here.

kwmsmith:
@postelrich For testing, we'll also need to add tests for sqlalchemy / postgres, because the `return_type` argument options affect that backend the most.

kwmsmith:
As discussed, we'll need to update the SQL test files, by adding an explicit `return_type='native'` to the existing calls to `compute`, and also add some additional tests that call `compute(..., return_type='core')` for the test files with a particular backend:
* blaze/compute/tests/test_postgresql_compute.py
* blaze/compute/tests/test_mysql_compute.py
* blaze/compute/tests/test_sql.py
* blaze/compute/tests/test_sql_compute.py
* blaze/compute/tests/test_sparksql.py

postelrich:
note those 2 commits with #1358 should be #1411

kwmsmith:
Closing this PR, moving over to #1417.

PR #1413, opened 2016-02-12
server performance

llllllllll:
The main meat of this pr is split into two sections:

### profiler

Without the ability to profile the server it is much harder to understand why queries are slow. I was profiling my client but to get a full picture of the runtime I needed to profile to server. This is mainly designed as a developer tool so it normally will write the pstats data to the server. If you are profiling your queries it is assumed you have access to the server. If you want the data sent back you can pass a file like object to `compute` with the keyword `profiler_output`. This will make the server send the data back with the response. This is really useful if you are running your server behind a load balancer and don't want to hunt for the stats.

### fastmsgpack_format

Profiling the client showed that we spent a disgusting amount of time deserializing tabular data with the msgpack format. In a query of mine I was calling `object_hook` some 15.5m times! This extends the serialization formats to make it easier to tune the performance. This also more efficiently sends arrays and tables. This is a massive perfomance improvement over the old msgpack implementation.

I also optimized the `object_hook` function because it is used for `json` and the old `msgpack` formats. We are also using it for the metadata in `fastmsgpack` so we want this to be really fast.


blaze/server/serialization.py
```diff
@@ -34,6 +87,101 @@ def _coerce_str(bytes_or_str):
...
+
+
+def fastmsgpack_data_dumps(data):
+    return {
+        '__!bytes': pd.to_msgpack(
+            None,
+            data,
+            compress=compress,
+            default=fastmsgpack_default,
+            encoding='latin1',
```
jreback: FYI in 0.18.0 default encoding will be utf8; I think the latin1 was originally a copy paste error

llllllllll: when using `utf-8` with a python3 server and a python2 client I got failures in the c++ code for msgpack about a dict not being hashable. I explicitly set this to `latin1` to avoid this. To be fair, the python2 client was using pandas 0.16.1 so it might be fixed on master.

jreback: could be msgpack was fixed up a bit on 0.17 but should be compat though

blaze/server/serialization.py
```diff
@@ -34,6 +87,101 @@ def _coerce_str(bytes_or_str):
...
+    raw = list(msgpack_unpack(
+        BytesIO(_l1(data)),
+        object_hook=object_hook,
+    ))
+    if len(raw) == 1:
+        return raw[0]
+    return raw
+
+
+def fastmsgpack_default(ob):
```
llllllllll: this is a workaround for: https://github.com/pydata/pandas/pull/12307

blaze/server/serialization.py
```diff
@@ -2,13 +2,65 @@
...
+from toolz import identity
 
-from ..compatibility import pickle as pickle_module, unicode
+from ..compatibility import pickle as pickle_module, unicode, PY2
+from ..interactive import coerce_scalar
 from ..utils import json_dumps, object_hook
 
 
-SerializationFormat = namedtuple('SerializationFormat', 'name loads dumps')
+_serializationformat = namedtuple(
```
kwmsmith: Any reason for this other than as the baseclass for immutable subclasses, like `SerializationFormat` below?

Any reason to prefer subclassing a `namedtuple` over specifying `__slots__` on `SerializationFormat` itself?

blaze/server/serialization.py
```diff
@@ -2,13 +2,65 @@
...
+
+
+class SerializationFormat(_serializationformat):
+    def __new__(cls,
+                name,
+                loads,
+                dumps,
+                data_loads=None,
+                data_dumps=None,
+                materialize=None):
```
kwmsmith: `__slots__ = ()` ?

blaze/server/serialization.py
llllllllll: In the original implementation we only had `loads` and `dumps`. We used tuple indexing instead of named access. We have since changed and now always use the named fields. I think we can get rid of this entirely and just subclass `object` without slots.

blaze/server/serialization.py
llllllllll: I am not sure that we need slots here. If users want to mutate this object then so be it. Adding the slots might make it harder to do other things later.

kwmsmith: If we don't need slots (and don't need to subclass from a `namedtuple`) that's great by me -- my comment was more of a question if you wanted to set an empty slots in conjunction with the subclass, which is now a moot point :)

blaze/server/serialization.py
```diff
@@ -1,14 +1,75 @@
...
+    dumps : callable[any -> bytes]
+        The function that serializes python objects to some other format.
+    data_loads : callable[any -> any], optional
+    data_dumps : callble[any -> any], optional
+        Specialized functions for loading and writing only the 'data' field of
+        blaze server responses. This allows us to define a more efficient
+        serialzation format for moving large amounts of data while still
+        having a rich representation for the rest of the metadata.
+    materialize : callble[(any, DataShape, **kwargs) -> any], optional
+        The function used to materialze the result of compute into a form
```
richafrank: Typos: "callble" (twice), "materialze", "serialzation"

blaze/server/serialization.py
```diff
@@ -34,6 +96,101 @@ def _coerce_str(bytes_or_str):
...
+        return bs
+
+
+def fastmsgpack_data_loads(data):
+    raw = list(msgpack_unpack(
+        BytesIO(_l1(data)),
+        object_hook=fastmsgpack_object_hook,
+    ))
+    if len(raw) == 1:
+        return raw[0]
```
richafrank: What is this case handling?  How are collections of length one serialized?

blaze/server/server.py
```diff
@@ -420,11 +493,29 @@ def compserver(payload, serial):
...
+
+        profiler.disable()
+        if profiler_output == ':response':
+            from pandas.compat import BytesIO
+            file = BytesIO()
+        else:
+            file = open(_prof_path(profiler_output, expr), 'wb')
+
+        with file:
+            marshal.dump(Stats(profiler).stats, file)
```
richafrank: How did you choose `marshal`?

llllllllll: I just looked at how `cProfile` writes this file. It didn't offer a way to pass a file like object so I just reimplemented the function here allowing us to give an in memory buffer or on disk file.

blaze/server/serialization.py
llllllllll: All objects are sent as lists. This case is so that if I have only sent back 1 dataframe we will get the single dataframe instead of a list of length 1 containing the dataframe. This logic is borrowed from pandas.io.packers.read_msgpack. We are reimplementing this instead of using that function so that we can forward the object hook in pandas 0.16.

blaze/server/server.py
llllllllll: added a comment about this.

blaze/server/serialization.py
llllllllll: added a comment about this.

blaze/server/serialization.py
```diff
@@ -34,6 +97,132 @@ def _coerce_str(bytes_or_str):
...
+except ImportError:
+    compress = None
+
+
+def _final_base(values):
+    base = values.base
+    while True:
+        try:
+            base = base.base
+        except AttributeError:
```
richafrank: Is there a benefit to stopping on `AttributeError` versus checking `is None`?  The latter seems clearer to me (maybe faster?).

llllllllll: when the base is a buffer then the buffer passed will be the base.

For example:
```python
In [1]: np.frombuffer(b'abc', dtype='uint8').base
Out[1]: b'abc'
```

richafrank: Aha TIL

blaze/server/server.py
```diff
@@ -393,38 +493,101 @@ def from_tree(expr, namespace=None):
...
+                    403,
+                )
+
+            profiler_output = requested_profiler_output
+            profiler = Profile()
+            profiler.enable()
+            # ensure that we stop profiling in the case of an exception
+            stack.callback(profiler.disable)
+
+        ns = payload.get('namespace', dict())
```
richafrank: Any reason not to use a dict literal?

llllllllll: This was just reindented, not changed. we can change this to a literal though.

blaze/server/server.py
```diff
@@ -393,38 +493,101 @@ def from_tree(expr, namespace=None):
...
+            'datashape': pprint(expr.dshape, width=0),
+            'data': serial.data_dumps(result),
+            'names': expr.fields
+        }
+        if profiling:
+            import marshal
+            from pstats import Stats
+
+            # remove our callback because we are going to exit now
+            stack.pop_all()
```
richafrank: Is it worth making this more robust to people adding other ctx managers to the stack?  Maybe exit the stack unconditionally after producing the `response`?

blaze/server/server.py
richafrank: Ah ok.

PR #1414, opened 2016-02-12
Deprecate compute-in-repr behavior.

kwmsmith:
Start process to replace Expr.__repr__ with a more standard
implementation.

Adds a `peek` method to do what `__repr__` used to do.

@llllllllll For the life of me I can't figure out how to configure the warning filters so that the warning only shows up once -- do you have any guidance here?


blaze/interactive.py
```diff
@@ -320,8 +320,21 @@ def table_length(expr):
...
+def _warning_repr(self):
+    # DeprecationWarnings are ignored by default, so we use a UserWarning here
+    # instead to ensure users are aware of this significant API change.
+    warnings.warn(_warning_msg)
+    return expr_repr(self)
+
+
+Expr.__repr__ = _warning_repr
+Expr.peek = lambda x: print(expr_repr(x))
 Expr._repr_html_ = lambda x: to_html(x)
```
llllllllll: we should also make the `_repr_html_` not implicitly compute when in jupyter.

blaze/interactive.py
```diff
@@ -320,8 +320,21 @@ def table_length(expr):
...
-Expr.__repr__ = expr_repr
+def _warning_repr(self):
+    # DeprecationWarnings are ignored by default, so we use a UserWarning here
+    # instead to ensure users are aware of this significant API change.
+    warnings.warn(_warning_msg)
+    return expr_repr(self)
+
+
+Expr.__repr__ = _warning_repr
+Expr.peek = lambda x: print(expr_repr(x))
```
llllllllll: I really love this name, it totally conveys the expected use case of "peeking" at the result while working on some expression!

kwmsmith: Good -- I wanted something shorter than `preview`...

llllllllll: One comment I have is that if we are going to call `print`, could we accept a file argument so that users can pass a `StringIO` or other output stream?

blaze/interactive.py
```diff
@@ -390,9 +426,38 @@ def table_length(expr):
...
+representation and will no longer implicitly compute.  Use the `peek()`
+method to see a preview of the expression's results.\
+"""
+
+use_new_repr = False
+def _choose_repr(self):
+    if use_new_repr:
+        return new_repr(self)
+    else:
+        # DeprecationWarnings are ignored by default, so we use a UserWarning here
```
llllllllll: we are actually using a deprecation warning here

llllllllll:
Other than the comment this looks good. Regarding the visibility of the warning: I think that we should not be forcing warnings to be more visible than they are by default. If users do not want to see deprecation warnings then that is on them.

kwmsmith:
@llllllllll thanks for the feedback.  I addressed your comment, updated whatsnew, and tweaked the `__repr__` implementations for `Symbol`, `_Data`, and `Expr` to be more consistent:

```python
In [1]: import blaze as bz

In [2]: t = bz.symbol('t', dshape='var * {name: string, id: int}')

In [3]: t
Out[3]: <`t` symbol; dshape='var * {name: string, id: int32}'>

In [4]: d = bz.data([['Alice', 1], ['Bob', 2]], dshape='var * {name: string, id: int32}', name='d')

In [5]: d
Out[5]: <'list' data; _name='d', dshape='var * {name: string, id: int32}'>

In [6]: d.name
Out[6]:
    name
0  Alice
1    Bob

In [7]: bz.interactive.use_new_repr = True

In [8]: d.name
Out[8]: <`Field` expression; dshape='var * string'>

In [9]: t.id ** 2
Out[9]: <`Pow` expression; dshape='var * int64'>

In [10]: str(t.id ** 2)
Out[10]: 't.id ** 2'

In [11]: d.name.peek()
Out[11]:
    name
0  Alice
1    Bob
```

Things to point out:
* The `__repr__`s are all boring reprs now that show the `dshape` along with the type of the thing.
* Long `dshape`s are truncated with a `...` at the end.
* The string form of an expression--whether compound or atomic--uses the underlying `_name` of the leaf terms to keep the string form looking nice.

kwmsmith:
I'm still looking into the best way to address the doctest failures.

llllllllll:
The doctest failures look like they are just relying on the new symbol repr, I think it is safe to just change them. I think we might want to include a bit more data in the reprs as we go, but we can do that after merging this. One immediate thing I am thinking of is including the field name in Field repr.

kwmsmith:
> The doctest failures look like they are just relying on the new symbol repr, I think it is safe to just change them. 

+1 -- although I didn't expect you to say that :)

> I think we might want to include a bit more data in the reprs as we go, but we can do that after merging this. One immediate thing I am thinking of is including the field name in Field repr.

+1.

llllllllll:
I'm really glad to see this feature merged, thanks!

Issue #1418, opened 2016-02-16
Tests for input validators


sandhujasmine:
Add/complete tests for input validators.

Start w/ adding tests for following (blaze.expr.collections) that raise exceptions.

- distinct
- schema_concat: not sure how this is exposed to user
- merge
  - merge tests are in test_collections.py - should validation test go there or in test_interactive?
  - _project() method raises error but haven't yet found how/when this is invoked
- join: already exists in test_collections
- concat: already exists in test_collections
- isin: already exists in test_collections
- shift: added a test in test_collections

Add tests to either test_interactive.py or test_collections.py

sandhujasmine:
__status__

- <s>distinct</s>
- schema_concat - not sure how this is exposed to user
- <s>merge</s>
  - merge tests are in test_collections.py - should validation test go there or in test_interactive?
  - _project() method raises error but haven't yet found how/when this is invoked
- <s>join</s> already exists in test_collections
- <s>concat</s> already exists in test_collections
- <s>isin</s> already exists in test_collections
- <s>shift</s> added a test in test_collections

PR #1420, opened 2016-02-17
TST: Add input validation tests

sandhujasmine:
Fixes GH 1418

Issue [status](https://github.com/blaze/blaze/issues/1418#issuecomment-184864228)

### DIFF
--- a/blaze/expr/tests/test_collections.py
+++ b/blaze/expr/tests/test_collections.py
@@ -108,12 +108,20 @@ def test_join_option_types():
     assert join(b, a, 'x').dshape == dshape('var * {x: int}')
 
 
-def test_join_mismatched_schema():
+def test_join_exceptions():
+    'mismatched schema; no shared fields'
     a = symbol('a', 'var * {x: int}')
-    b = symbol('b', 'var * {x: string}')
-
     with pytest.raises(TypeError):
+        b = symbol('b', 'var * {x: string}')
         join(a, b, 'x')
+
+    with pytest.raises(ValueError):
+        b = symbol('b', 'var * {z: int}')
+        join(a, b)
+
+    with pytest.raises(ValueError):
+        b = symbol('b', 'var * {x: int}')
+        join(a, b, how='inner_')
 
 
 def test_join_type_promotion():
@@ -214,3 +222,9 @@ def test_shift():
     assert repr(t.shift(1)) == 'shift(t, n=1)'
     assert repr(t.shift(0)) == 'shift(t, n=0)'
     assert repr(t.shift(-1)) == 'shift(t, n=-1)'
+
+
+def test_shift_not_int():
+    a = symbol('a', '3 * {x: int32, y: int32}')
+    with pytest.raises(TypeError):
+        a.x.shift(1.3)
--- a/blaze/expr/tests/test_collections.py
+++ b/blaze/expr/tests/test_collections.py
@@ -41,6 +41,14 @@ def test_merge_common_subexpression():
     assert result.isidentical(t.a)
 
 
+def test_merge_exceptions():
+    t = symbol('t', 'var * {x:int}')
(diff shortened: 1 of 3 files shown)


blaze/expr/tests/test_collections.py
```diff
@@ -108,13 +108,21 @@ def test_join_option_types():
...
 
 
-def test_join_mismatched_schema():
+def test_join_exceptions():
+    'mismatched schema; no shared fields'
     a = symbol('a', 'var * {x: int}')
-    b = symbol('b', 'var * {x: string}')
-
     with pytest.raises(TypeError):
+        b = symbol('b', 'var * {x: string}')
```
llllllllll: We should move this out of the context manager. As a general rule context invariant code should not be placed in the context manager when possible because then readers need to know if the context manager affects the code in any way. In this particular case, if `symbol` raises a `TypeError` then this test will pass when it shouldn't

sandhujasmine: Indeed - thanks @llllllllll 
Fixed and tested.

llllllllll:
How do you feel about testing the contents of the error strings?
something like:

```python
with pytest.raises(Error) as e:
    ...
assert '...' in str(e.value)
assert str(e.value) == '...'
```

We might want to test that the error message includes the correct column or incorrect value.

blaze/expr/tests/test_collections.py
```diff
@@ -108,13 +108,21 @@ def test_join_option_types():
     assert join(b, a, 'x').dshape == dshape('var * {x: int}')
 
 
-def test_join_mismatched_schema():
+def test_join_exceptions():
+    'mismatched schema; no shared fields'
```
llllllllll: Can we use triple quoted strings for docstrings. It took me a second to realize this was the docstring.

llllllllll:
Thank you for the test cases!

blaze/expr/tests/test_collections.py
sandhujasmine: > How do you feel about testing the contents of the error strings?

I feel good about it - I had added but wasn't sure if it was overkill. It's easy to add so I'll update.

> Can we use triple quoted strings for docstrings

Yes, will fix.

sandhujasmine: > How do you feel about testing the contents of the error strings?
> something like:
> ```
> with pytest.raises(Error) as e:
>    ...
> ```

Instead of doing an exact match would it suffice to check substring? This will keep it concise and not require string formatting. I don't think exact match buys us much more once we've validated the correct exception was raised.

```
with pytest.raises(TypeError) as excinfo:
    join(a, b, 'x')

assert 'Schemata of joining columns do not match,' in str(excinfo.value)
```

The exact message for this exception is:
`
'Schemata of joining columns do not match,'
                ' no promotion found for %s=%s and %s=%s' % (
                    on_left[n], left_types[n], on_right[n], right_types[n],
                )
`

llllllllll: that is fine, we might wasa to just check that the expected column names appear in the string

sandhujasmine:
@kwmsmith, @llllllllll 

I was thinking to move exception tests for objects defined in collections module from test_interactive.py to test_collections.py, so moving:
`test_sample_failures(), test_distinct_failures(), test_merge_failure()`

I added the last two here but seems like the general convention is for tests for a module to be in associated test_<module>.py

What do you think?

llllllllll:
Moving the tests sounds good. Keeping tests organized makes it easier for developers to find the place to add new tests which encourages better testing.

kwmsmith:
> I was thinking to move exception tests for objects defined in collections module from test_interactive.py to test_collections.py

:+1:

llllllllll:
Thank you for these tests and cleaning up the existing ones!

PR #1421, opened 2016-02-18
1358 rename Data to data

### LINKED ISSUE #1358: Data -> data and other API inconsistencies

postelrich:
closes #1358 

rename Data to data and add deprecation warning to Data.

### DIFF
--- a/blaze/utils.py
+++ b/blaze/utils.py
@@ -310,3 +310,13 @@ class attribute(object):
             return self
 
         return self._f(instance)
+
+
+class BlazeDeprecationWarning(DeprecationWarning):
+    """Blaze specific deprecation warning."""
+    pass
+
+
+class BlazePendingDeprecationWarning(PendingDeprecationWarning):
+    """Blaze specific pending deprecation warning."""
+    pass
--- a/blaze/__init__.py
+++ b/blaze/__init__.py
@@ -15,7 +15,10 @@ from multipledispatch import halt_ordering, restart_ordering
 halt_ordering()  # Turn off multipledispatch ordering
 
 from datashape import dshape, discover
-from .utils import ignoring
+from .utils import ignoring, BlazeDeprecationWarning, BlazePendingDeprecationWarning
+import warnings
+warnings.simplefilter('once', BlazeDeprecationWarning)
+warnings.simplefilter('once', BlazePendingDeprecationWarning)
 from .expr import (Symbol, TableSymbol, symbol, ndim, shape)
 from .expr import (by, count, count_values, distinct, head, join, label, like,
         mean, merge, nunique, relabel, selection, sort, summary, var,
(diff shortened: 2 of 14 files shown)


blaze/utils.py
```diff
@@ -310,3 +310,13 @@ def __get__(self, instance, owner):
             return self
 
         return self._f(instance)
+
+
+class BlazeDeprecationWarning(DeprecationWarning):
```
llllllllll: is there a reason to make these classes rather than just using the superclass directly?

llllllllll:
How do people feel about the name, `bound_expression` or something? `data` feels too close to what `resource` is.

kwmsmith:
> How do people feel about the name, bound_symbol or something? data feels too close to what resource is.

As you know, we discussed this at length in another issue thread.

We bikeshedded the naming question again among 4-5 people before starting this PR.  We decided to stay with `data`:
* It's a very common end-user entry point, so we want to keep it short;
* Given the target end user--Data Scientist, Business Analyst, Data Analyst, etc.--`bound_symbol` was thought to be too far removed from their working vocabulary;
* `data`, `compute`, and `expression` are the three main concepts we use when talking about blaze, so it's a big plus to have those three concepts reflected in the API directly.

blaze/utils.py
kwmsmith: We originally thought it would be good to have a subclass to make it easily filterable, but now I'm with you, and see no really good reason not to just use `DeprecationWarning` directly.

postelrich: The idea was to so that we can set `warnings.simplefilter` to only display the warning once but I couldn't get it to work.  I created the derived class so that I wouldn't effect how other warnings are displayed, especially since deprecation warnings are ignored by default.

postelrich:
@llllllllll I agree that `data`is too generic of a name (especially as part of a data science library/distribution) and not descriptive enough of what it's doing.  But I also agree with @kwmsmith against naming it something too technical.

At the risk of starting another lengthy discussion, I propose that if we were to change the name that it be simple.  So something like `blaze.ignite` because you're firing up a powerful connection to your data.  Or really simple like `blaze.on` or `blaze.open`.

But we shouldn't delay progress, rather just keep the discussion open and make a decision before we get to a 1.0.

llllllllll:
@kwmsmith Sorry, I forgot that we had come to a conclusion in the other thread.

kwmsmith:
Closing this PR, see #1431.

Issue #1423, opened 2016-02-19
Bug in sample(frac=) expression


sandhujasmine:
When randomly sampling a fraction of the results for an SQL backend, it always returns 1 row - tested against postgres. Will update test_postgres_compute() to include a failing test right after opening this issues so it can be referenced in the xfail.

The following expression always returns 1 row regardless of the number of rows contained in the `nyc` table (test in postgres backend): `result2 = compute(t.sample(frac=0.5), nyc)`

sandhujasmine:
PR #1425 addresses this issue.

PR #1424, opened 2016-02-19
TST: Add failing test for GH 1423

sandhujasmine:
Add failing test and mark with xfail for [GH 1423](https://github.com/blaze/blaze/issues/1423)

To run the test using docker provisioning.sh script, after odo [PR#428](https://github.com/blaze/odo/pull/428) is merged; run the ./provisioning.sh and add the blaze/examples/data as a mount point for docker so it can load data into postgres.

Example:
`provisioning.sh -t ~/tmp/docker/ -m ${BLAZE_PROJ_PATH}/blaze/blaze/examples/data restart`

### DIFF
--- a/blaze/compute/tests/test_postgresql_compute.py
+++ b/blaze/compute/tests/test_postgresql_compute.py
@@ -575,11 +575,21 @@ def test_interactive_len(sql):
     assert len(t) == int(t.count())
 
 
-def test_sample(sql):
-    t = symbol('t', discover(sql))
-    result = compute(t.sample(n=1), sql)
+def test_sample_n(nyc):
+    t = symbol('t', discover(nyc))
+    result = compute(t.sample(n=14), nyc)
     s = odo(result, pd.DataFrame)
-    assert len(s) == 1
-    result2 = compute(t.sample(frac=0.5), sql)
+    assert len(s) == 14
+
+
+@pytest.mark.xfail(raises=AssertionError,
+                   reason='Test for GH Issue # 1423')
+def test_sample_frac(nyc):
+    """
+    sample(frac=x) always returns 1 row.
+    """
+    t = symbol('t', discover(nyc))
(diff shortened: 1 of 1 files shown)


blaze/compute/tests/test_postgresql_compute.py
```diff
@@ -575,11 +575,21 @@ def test_interactive_len(sql):
...
-def test_sample(sql):
-    t = symbol('t', discover(sql))
-    result = compute(t.sample(n=1), sql)
+def test_sample_n(nyc):
+    t = symbol('t', discover(nyc))
+    result = compute(t.sample(n=14), nyc)
     s = odo(result, pd.DataFrame)
-    assert len(s) == 1
-    result2 = compute(t.sample(frac=0.5), sql)
+    assert len(s) == 14
```
llllllllll: should we check that the sorted results are the same ?

llllllllll:
Thanks for adding an xfail test. This is great way to codify tracking this issue!

blaze/compute/tests/test_postgresql_compute.py
sandhujasmine: Not sure I understand - could you explain a bit more please.

The `sample()` method will by default do a random sampling of the data in the SQL query by invoking `ORDER BY random()`.

In both cases we're doing a random sampling; however when using  `n=100` we have LIMIT defined as 100 while for `frac=0.5`, the LIMIT is computed as `COUNT(*) * 0.5`

llllllllll: sorry, maybe I incorrectly assumed that we were doing a sample of the whole table which should just shuffle it. If this is a subset then I think that we should test that all the records exist in the base table as a sanity check that we are actually sampling correcectly.

kwmsmith:
Closing as this is subsumed in #1452.

PR #1425, opened 2016-02-19
FIX: sample(frac=...) on SQL.

kwmsmith:
Have to add a `from_obj` clause inside the sub-select.

@sandhujasmine please try this against your datasets.


blaze/compute/tests/test_sql_compute.py
```diff
@@ -19,11 +19,10 @@
 
 from odo import odo, drop
 from blaze.compatibility import xfail
-from blaze.compute.sql import compute, select, lower_column, compute_up
-from blaze.expr import (
-    symbol, transform, summary, by, sin, join,
-    floor, cos, merge, nunique, mean, sum, count, exp, datetime as bz_datetime
-)
+from blaze.compute.sql import (compute, select, lower_column, compute_up,
```
llllllllll: I have a style request: can we put parenthesized imports on their own lines sorted alphabetically like:

```python
from a import (
    b,
    c,
)
```

I have been updating these as I see them. The reason I prefer this style is that it makes it very easy to trace imports visually and understand where names are coming from. As it is now it is hard to quickly see what names are available, or the fact that sum is being aliased here.

If you have strong feeling about this please let me know,

sandhujasmine:
@kwmsmith - It doesn't appear to be subsampling now. See below.
```
In [7]: odo(compute(sub_parts.nrows, prsx), int)
Out[7]: 40040

In [8]: odo(compute(sub_parts.sample(frac=0.75).nrows, prsx), int)
Out[8]: 40040
```
Also added a [failing test](https://github.com/blaze/blaze/pull/1424) in blaze.

kwmsmith:
Closing, as this is subsumed by #1452.

PR #1427, opened 2016-02-22
WIP: changed from_yaml to use pathes relative to yaml file, not relative to CWD

brendancol:
To help with portability of yaml config files, I've changed `from_yaml` to look relative to the yaml `path.name` instead of the current working directory.  

So my thinking is:
  - if `source` is a relative file path (`../my_csv.csv`), then make it absolute path relative to the yaml config
  - if `source` is a uri (`http://www.google.com`), then let it be...
  - if `source` includes `~`, then let it be and allow `expanduser` to deal with it...
  - if `source` contains a protocol and relative path (`sqlite://../examples/iris/db`) then make it absolute path relative to yaml files.

What other situations should i consider?

### DIFF
--- a/blaze/server/spider.py
+++ b/blaze/server/spider.py
@@ -7,7 +7,6 @@ import sys
 import argparse
 import importlib
 
-import toolz
 import yaml
 
 from odo import resource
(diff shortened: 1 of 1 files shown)


blaze/server/spider.py
```diff
@@ -105,9 +104,27 @@ def from_yaml(path, ignore=(ValueError, NotImplementedError), followlinks=True,
...
+            # convert relative pathes to absolute relative to yaml file
+            if (not os.path.isabs(source) and
+               not source.startswith('http') and
+               '~' not in source):
+                os.chdir(yaml_dir)
+                src = os.path.abspath(source)
+                os.chdir(owd)
+
+                # handle 'sqlite:///../path/with/protocol/test.db'
+                if ('://' in source):
```
llllllllll: Cases like this make me think we should just bracket this block with a chdir. This way we don't need to manipulate the paths directly.

llllllllll:
I would also recommend making this feature optional. I could imagine someone expecting the paths to be relative to the cwd of the server which is much more common behavior.

kwmsmith:
This PR is absorbed into #1460, closing.

PR #1431, opened 2016-02-25
Complete `Data` to `data` rename

kwmsmith:
This builds on PR #1421:
* All calls to `Data` are replaced with `data`;
* Calls to `Data` are deprecated;
* `InteractiveSymbol` is deprecated and replaced by `_Data`;
* Calls to `data` create an instance of `_Data`;
* Calls to `odo.resource` are replaced by `blaze.data` throughout the codebase and testsuite.

In version 0.11, the following changes will take place, to complete the transition:
* `InteractiveSymbol` will be removed;
* The `_Data` class will be renamed `Data`, and calls to `data` will create a `Data` instance;
* Just like Blaze's expressions, the public API is all lowercase function-like objects, and they create instances of their corresponding classes--`join` and `Join`, `head` and `Head`, etc.


blaze/interactive.py
```diff
@@ -41,19 +41,27 @@
     not_an_iterator.append(pymongo.database.Database)
 
 
-class InteractiveSymbol(Symbol):
-    """Interactive data.
+class _Data(Symbol):
```
llllllllll: should we just call this `Data` to match the other expression/smart constructor pairs?

kwmsmith: Yes, that's exactly the ultimate intent, and in version 0.11 we'll remove `InteractiveSymbol` and rename `_Data` to `Data`.  I made note of this in the whatsnew document.

llllllllll: Ah, I should read the whole change before commenting; sorry

PR #1436, opened 2016-03-03
ENH: Added logging to compute endpoint

richafrank:
Includes expression being computed and total time to compute


llllllllll:
Just update the issue in the whatsnew. Thanks you for adding the logs!

richafrank:
Updated the issue in the whatsnew.  Thanks!

PR #1441, opened 2016-03-10
Remove Table and TableSymbol

kwmsmith:
Both of these have been long deprecated.  This PR removes them, and migrates the `TableSymbol` tests to just test `Symbol` instead.


kwmsmith:
@llllllllll will merge in the next day or so.

llllllllll:
Sounds good, these have been deprecated since I started working on blaze ;)

PR #1442, opened 2016-03-10
Fix bug in Pandas compute_up() on Broadcast exprs.

kwmsmith:
Adds broadcast tests for Pandas frames / series and Dask dataframes.

### DIFF
--- a/blaze/compute/tests/test_pandas_compute.py
+++ b/blaze/compute/tests/test_pandas_compute.py
@@ -14,7 +14,7 @@ from string import ascii_lowercase
 
 from blaze.compute.core import compute
 from blaze.compute.pandas import pdsort
-from blaze import dshape, discover, transform
+from blaze import dshape, discover, transform, broadcast_collect
 from blaze.expr import symbol, join, by, summary, distinct, shape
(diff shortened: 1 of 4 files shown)


blaze/compute/tests/test_pandas_compute.py
```diff
@@ -45,6 +45,20 @@
...
+    t = symbol('t', 'var * {a: int64}')
+    bcast = broadcast_collect(expr=(2 * t.a - t.a**2))
+    result = compute(bcast, s)
+    assert_series_equal(2 * s - s**2, result)
+
+
+def test_frame_broadcast():
+    bcast = broadcast_collect(expr=t.amount * t.id)
+    result = compute(bcast, df)
+    assert_series_equal(df.amount * df.id, result)
```
llllllllll: can we swap the assertion to make the lhs the result and the rhs the expected? I know that this is equivalent but that pattern is more common

llllllllll:
Thanks for the test coverage, looks good!

blaze/compute/tests/test_pandas_compute.py
kwmsmith: Done.

PR #1445, opened 2016-03-14
MAINT: explicitly pass return_type='native' in internal cals to compute

llllllllll:
Still got warnings after the last change because we used the deprecated default


llllllllll:
@kwmsmith We might need to update our tests. I would highly recommend making warnings failures inside the test suite to catch this.

kwmsmith:
> I would highly recommend making warnings failures inside the test suite to catch this.

Agreed, I'll make an issue to track.

PR #1451, opened 2016-03-18
Merge core and blaze pkgs

kwmsmith:
To simplify the blaze packaging, we merged the core and blaze packages into one recipe / package.  This is consistent with the anaconda distribution of blaze, which is now a single package as well.


conda.recipe/meta.yaml
```diff
@@ -15,31 +15,40 @@ source:
 requirements:
   build:
```
llllllllll: why do we have any build deps other than python and setuptools?

kwmsmith: That's a good question -- I'll investigate.  I knew at one point, but I've since forgotten.  It would be good to whittle this list down to a minimum.

kwmsmith: When only enabling `python` and `setuptools`, `conda build` gives me this exception:

```
RuntimeError: Setuptools downloading is disabled in conda build. Be sure to add all dependencies in the meta.yaml
```

So unfortunately we have to include all packages in `build` that are required during the `python setup.py install` process. :(

llllllllll: we don't have any dependencies in our setup.py though, we are not building anything

kwmsmith: We have installation requirements:
* https://github.com/blaze/blaze/blob/master/setup.py#L55-L56
* https://github.com/blaze/blaze/blob/master/requirements-strict.txt

If there's a way to consolidate the `requirements*.txt` files and `meta.yaml`, I'm all ears -- I hate having duplicated stuff.  I haven't found a good way to improve this, however.

llllllllll: The way we do this in zipline is we load up the setup.py and pull the reqs out that way:

https://github.com/quantopian/zipline/blob/master/conda/zipline/meta.yaml
https://github.com/quantopian/zipline/blob/master/setup.py

kwmsmith:
@llllllllll Unfortunately we have to handle our dependencies differently on windows vs. posix, and we have python version specific constraints as well.  We've used conda selectors for this, but that feature isn't supported by setuptools.

I don't have the time in the near future to refactor things so that both setuptools and conda use the same requirements files and handle the platform / python version matrix correctly.

The core of this PR is the merging of the `core` and `all` conda recipes, which we've accomplished.  I'll make an issue to note the desire to refactor the dependency management.

llllllllll:
That's fine, removing the duplication is nice to have but not required

PR #1452, opened 2016-03-18
Fix sample with `frac` argument on postrgres backend.

### LINKED ISSUE #1423: Bug in sample(frac=) expression

kwmsmith:
Subsumes #1425 and #1424.

Closes #1423.


kwmsmith:
ping @sandhujasmine

sandhujasmine:
Looks good - it contains tests for various `sample()` configurations.

And `sample(frac=` is working correctly; however, saw small inconsistencies between using sample() for a pandas backend vs DB backend.

1. When using sample(n=176) if the result set has 175 rows, for the DB backend we get all rows but pandas backend raises an error. The pandas sample works a little bit differently.

2. Also, in an example, found that using frac=0.3 for DB returns 53 rows while pandas (read from CSV file) returns 52 rows; basically one is rounding down and the other is rounding (ceil) up.

Would be good to have consistent behavior if looking at same data either from files or loaded up in DB. This is more when testing the same symbolic (unbound) expression against different resources.

Maybe we should work towards merging this as is, then add failing tests to reproduce above cases, then work on solutions?

Issue #1095, opened 2015-05-14
exception in test_dask.py (without numba)


cowlicks:
I removed numba from Blaze. And this exception was coming up intermittently, so I rewrote the failing test to have a while loop at the problem point like this:

```python
def test_mean():
    count = 0
    expr = sx.mean(axis=0)
    result = compute(expr, dask_ns)
    expected = compute(expr, numpy_ns)
    assert isinstance(result, Array)
    if expr.dshape.shape:
        while True:
            count += 1
            print(count)
            result2 = np.array(result)
    else:
        result2 = float(result)
    assert eq(result2, expected)
```
running this I get
 
```python
================================================================================================ test session starts ================================================================================================
platform linux2 -- Python 2.7.9 -- py-1.4.26 -- pytest-2.7.0 -- /home/blake/.miniconda/envs/dask-hang/bin/python
rootdir: /home/blake/github/blaze, inifile: 
collected 25 items 

blaze/compute/tests/test_dask.py::test_mean 1
2
3
4
5
6
7
8
9
10
11
12
13
14
15
16
17
18
19
FAILED

===================================================================================================== FAILURES ======================================================================================================
_____________________________________________________________________________________________________ test_mean _____________________________________________________________________________________________________

    def test_mean():
        count = 0
        expr = sx.mean(axis=0)
        result = compute(expr, dask_ns)
        expected = compute(expr, numpy_ns)
        assert isinstance(result, Array)
        if expr.dshape.shape:
            while True:
                count += 1
                print(count)
>               result2 = np.array(result)

blaze/compute/tests/test_dask.py:75: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../dask/dask/array/core.py:651: in __array__
    x = self.compute()
../dask/dask/array/core.py:692: in compute
    result, = compute(self, **kwargs)
../dask/dask/array/core.py:487: in compute
    results = get(dsk, keys, **kwargs)
../dask/dask/array/core.py:1020: in get
[...]

PR #1454, opened 2016-03-19
Skip dask test on windows + Python 2.7.

kwmsmith:
Get nondeterministic test failures relating to dask's `get_async`.

I want to return to this perhaps with @mrocklin's help, but for now, I'm marking it with a `skipif()` on Windows and Python 2.7.

### DIFF
--- a/blaze/compute/tests/test_dask.py
+++ b/blaze/compute/tests/test_dask.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import pytest
+import sys
 
 dask = pytest.importorskip('dask')
 
@@ -65,6 +66,8 @@ exprs = [2 * sx + 1,
 
 
 @pytest.mark.parametrize('expr', exprs)
+@pytest.mark.skipif(sys.platform == 'win32' and sys.version_info[:2] == (2, 7),
+		    reason="Nondeterministic exceptions relating to get_async")
 def test_compute(expr):
     result = compute(expr, dask_ns)
     expected = compute(expr, numpy_ns)


kwmsmith:
Looks like this is hitting the same bug as reported in #1095.  Perhaps it showed up because I'm running the testsuite on a new Windows VM (win 7 64 bit).

mrocklin:
Which test fails?

kwmsmith:
On any given run, I've seen expr1, expr8, and expr9 each fail:
* Expr1: `sx.sum(axis=0)`
* Expr8: `sx.sum()`
* Expr9: `sx - sx.sum()`

They all have `sx.sum()` in common.  This only reveals itself in python 27.

mrocklin:
Ah, well, even though this is non-deterministic it might not be incorrect.  Perhaps the tests should be changed to use something like `np.allclose`?

You're right that it would be nice to to be able to guarantee order though, at least in the single-threaded case.

kwmsmith:
> Perhaps the tests should be changed to use something like `np.allclose`?

Agreed with you here -- I'll make that change.

I misspoke when I said "failiure", I meant "error", as in, we see nondeterministic errors, sometimes `KeyErrors

kwmsmith:
> Perhaps the tests should be changed to use something like `np.allclose`?

Agreed -- I'll make that change.

I misspoke when I said "failiure"; I should have said "error".  We see nondeterministic errors, sometimes `KeyErrors`, and other times (minimally informative) Blaze `NotImplementedErrors` ("don't know how to compute ..."), and other times all tests pass fine.  Since tests do eventually pass, I suspect the problem isn't in the `eq()` implementation, but something more subtle.

I haven't had time to bound the issue; given that it's not reproducible, I suspect there's some potentially deep threading issue or somesuch.  My starting point would be to see if I can demonstrate it in dask without blaze; failing that, then I'll see if it's something to do with the dask backend for blaze.

FYI, I won't be able to put much time on this for a couple of weeks, due to other priorities.

PR #1456, opened 2016-03-21
Add test to ensure sampling is bounded in SQL

kwmsmith:
Also converts some `odo(compute(...))` calls to `compute(..., return_type=...)`.


llllllllll:
+1

Issue #1457, opened 2016-03-22
Support HAVING clause for DB backends; incorrectly using WHERE for aggregate functions


sandhujasmine:
Blaze does not currently support HAVING clause. It tries to use the WHERE clause with aggregate functions which results in an error.

Here's an example SQL query:
```
SELECT 
    tbl0."A",
    sum(tbl0."B") AS quant
FROM 
    tbl0
GROUP BY
    tbl0."A" HAVING sum(tbl0."B") > 125
```

Writing this using a Blaze expression:
```
# big_sql contains a postgres table with following shape and values:
# ds is dshape("var * {A: string, B: int64}")
# column 'A' contains: list("".join(['a'*25, 'b'*25, 'c'*25, 'd'*25]))
# column 'B' contains: np.random.randint(1, 10, (100,)).tolist()
# Add this as a failing test to test_postgres_compute.py
ds = discover(big_sql)
nn = symbol('nn', ds)
g1 = by(nn['A'], quant=nn.B.sum())

# Syntax below is consistent with pandas syntax
expr = g1[g1.quant > 125]
res = compute(expr, big_sql)
bres = odo(res, list)
```

Doing a print(res) shows blaze writes this using WHERE clause which fails during execute since we must HAVING with aggregate (sum()) function.

```
print(res)
SELECT tbl0."A",
    sum(tbl0."B") AS quant
FROM 
    tbl0
WHERE 
    sum(tbl0."B") > 125
GROUP BY
    tbl0."A"
```

mattfrei27:
This is still a problem. See http://stackoverflow.com/questions/42171715/blaze-generating-invalid-sql-for-simple-sql-having-style-query for a reproducible example.

PR #1458, opened 2016-03-22
TST: Add failing test for HAVING feature request

sandhujasmine:
Add failing test and mark as xfail for requested feature in GH 1457.
#1457

Test can be run using odo [docker provisioning](/blaze/odo/pull/428#issuecomment-199887590)

### DIFF
--- a/blaze/compute/tests/test_postgresql_compute.py
+++ b/blaze/compute/tests/test_postgresql_compute.py
@@ -91,8 +91,8 @@ def big_sql(url):
     except sa.exc.OperationalError as e:
         pytest.skip(str(e))
     else:
-        t = odo(zip(list("".join(['a' * 25, 'b' * 25, 'c' * 25, 'd' * 25])),
-                    np.random.randint(1, 10, (100,)).tolist()), t)
+        t = odo(zip("".join(['a' * 25, 'b' * 25, 'c' * 25, 'd' * 25]),
+                    range(1, 101)), t)
         try:
             yield t
         finally:
@@ -699,6 +699,28 @@ def test_coalesce(sqla):
     )
 
 
+def test_having_exp_group_by(big_sql):
+    '''
+    The following three tests (test_having, test_having_multiple_conditions,
+    test_having_where) expect big_sql column B to contain range(1, 101). Verify
+    results of 'group by' before proceeding - just to ensure the data in
+    big_sql isn't changed inadvertently.
+    '''
+    ds = discover(big_sql)
+    nn = symbol('nn', ds)
+    g1 = by(nn['A'], quant=nn.B.sum())
+
+    # expected result of group by for big_sql table
+    #    A  quant
+    # 0  a    325
+    # 1  b    950
+    # 2  c   1575
+    # 3  d   2200
+    # ensure this is correct for one of the tests
+    g1_res = compute(g1.sort('A'), big_sql, return_type=list)
+    assert g1_res == [('a', 325), ('b', 950), ('c', 1575), ('d', 2200)]
+
+
 def test_having(big_sql):
     '''
     The blaze expression generated in 'res' below is incorrect. It invokes the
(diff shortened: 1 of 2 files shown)


blaze/compute/tests/test_postgresql_compute.py
```diff
@@ -696,3 +697,35 @@ def test_coalesce(sqla):
...
+
+        SELECT tbl0."A", sum(tbl0."B") AS quant
+            FROM tbl0
+        GROUP BY
+            tbl0."A" HAVING sum(tbl0."B") > 125
+    '''
+    ds = discover(big_sql)
+    nn = symbol('nn', ds)
+    g1 = by(nn['A'], quant=nn.B.sum())
+    g1_res = odo(compute(g1, big_sql), list)
```
kwmsmith: Can you remove the `odo` call and convert this to `compute(g1, big_sql, return_type=list)` instead, for consistency?

blaze/compute/tests/test_postgresql_compute.py
```diff
@@ -696,3 +697,35 @@ def test_coalesce(sqla):
...
+    assert len(g1_res) == 4
+
+    expr = g1[g1.quant > 125]
```
kwmsmith: Can we add another `assert` or a separate test that has a compound selection predicate, something like:

```
g1[g1.A.like(...) & (g1.quant > 125)]
```

Also, another test case that combines `WHERE`, `GROUP BY`, and `HAVING`, something like:

```
nn_sel = nn[nn.B < 100]
g1 = by(nn_sel['A'], quant=...)
expr = g1[g1.quant > 125]
...
```

blaze/compute/tests/test_postgresql_compute.py
```diff
@@ -91,7 +91,8 @@ def big_sql(url):
     except sa.exc.OperationalError as e:
         pytest.skip(str(e))
     else:
-        t = odo(zip(list('a'*100), list(range(100))), t)
+        t = odo(zip(list("".join(['a'*25, 'b'*25, 'c'*25, 'd'*25])),
```
llllllllll: style nit: can we put spaces around binary operators, without them expressions are very crowded

sandhujasmine: Fixed in latest push.

blaze/compute/tests/test_postgresql_compute.py
sandhujasmine: Fixed in latest commit.

blaze/compute/tests/test_postgresql_compute.py
sandhujasmine: Added 2 more tests: one for a more complex condition on HAVING and another one that uses WHERE and HAVING together.

blaze/compute/tests/test_postgresql_compute.py
```diff
@@ -696,3 +697,99 @@ def test_coalesce(sqla):
...
+    #    SELECT tbl0."A", tbl0."B"
+    #    FROM tbl0
+    #    WHERE tbl0."B" > 5
+    s1 = nn[nn.B > 5]
+
+    g1 = by(s1['A'], quant=s1.B.sum())
+    g1_res = compute(g1, big_sql, return_type=pd.DataFrame)
+    assert len(g1_res) == 4
+
+    expr = g1[(g1.quant < 100) & (g1.quant > 50)]
```
sandhujasmine: Looks like this part of the expression is not being correctly computed for Python 2.7 .. will setup a 2.7 environment and try to repro this error.

sandhujasmine:
HAVING tests were previously using randomly generated data - updated data so it is fixed and updated tests accordingly. `test_having_where()` should now consistently fail since the compiled SQL for this is incorrect - still working on this in #1457

blaze/compute/tests/test_postgresql_compute.py
sandhujasmine: It's not a Python version issue. The SQL compiled is incorrect - still working on #1457

sandhujasmine:
Here's how to reproduce the incorrect SQL for the two conditionals on group_by expression.

```
from blaze import by, compute, discover, symbol, data, odo, drop
import pandas as pd

big_sql = data('postgresql://jsandhu@0.0.0.0/test::tbl0', dshape='var * {A: string, B: int64}')
drop(big_sql)

big_sql = odo(zip("".join(['a' * 25, 'b' * 25, 'c' * 25, 'd' * 25]), range(1, 101)), big_sql)
ds = discover(big_sql)

nn = symbol('nn', ds)

s1 = nn[nn.B > 5]

for data_ in (nn, s1):
    g1 = by(data_['A'], quant=data_.B.sum())

    expr = g1[(g1.quant > 325) & (g1.quant < 2000)]
    bz_res = compute(expr, big_sql)
    print('SQL for ', data_, ': \n')
    print(bz_res)
    print('\nResult for g1[(g1.quant > 325) & (g1.quant < 2000)]:')
    print(compute(expr, big_sql, return_type=pd.DataFrame))
    print('==========\n')
```

PR #1459, opened 2016-03-23
Error checking when adding to blaze server.

kwmsmith:
Does basic error checking when adding a new data set to a blaze server (`/add` endpoint):
* prevent adding a dataset when there's a name collision with existing dataset;
* check that the data is `discover`able to verify the data is loadable and queryable;
* capture any other exception and return appropriate error code and traceback string.

I also reformatted the indentation in other tests for consistency.

### DIFF
--- a/blaze/server/server.py
+++ b/blaze/server/server.py
@@ -36,8 +36,9 @@ DEFAULT_PORT = 6363
 
 
 class RC(object):
-    """Simple namespace for HTTP status codes.
-    https://en.wikipedia.org/wiki/List_of_HTTP_status_codes 
+    """
+    Simple namespace for HTTP status codes.
+    https://en.wikipedia.org/wiki/List_of_HTTP_status_codes
     """
 
     OK = 200
@@ -69,11 +70,8 @@ def _get_option(option, options, default=_no_default):
             return default
 
         # Provides a more informative error message.
-        raise TypeError(
-            'The blaze api must be registered with {option}'.format(
-                option=option,
-            ),
-        )
+        msg = 'The blaze api must be registered with {option}'
+        raise TypeError(msg.format(option=option))
 
 
 def ensure_dir(path):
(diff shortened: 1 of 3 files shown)


blaze/server/server.py
```diff
@@ -70,12 +67,9 @@ def _register_api(app, options, first_registration=False):
     Register the data with the blueprint.
     """
     _get_data.cache[app] = _get_option('data', options)
-    _get_format.cache[app] = dict(
-        (f.name, f) for f in _get_option('formats', options)
-    )
-    _get_auth.cache[app] = (
-        _get_option('authorization', options, None) or (lambda a: True)
-    )
+    _get_format.cache[app] = dict((f.name, f) for f in _get_option('formats', options))
```
llllllllll: this should probably be a dict comprehension now

llllllllll:
It is hard to see what actually changed here, could you split this into two commits?

llllllllll:
Also, since this is mainly a style change, I would like to say that I don't find `dict(key=value)` as clear as `{'key': value}`. For dictionary literals I think this violates the "one obvious way to do something" principle.

blaze/server/tests/test_server.py
```diff
@@ -583,72 +518,75 @@ def test_isin(test, serial):
...
-        assert response1.status_code == 200
+        payload = serial.dumps({'iris': iris_path})
+        empty_server.post('/add',
+                          headers=mimetype(serial),
+                          data=payload)
+        # Try to add to existing 'iris'
+        resp = empty_server.post('/add',
+                                 headers=mimetype(serial),
+                                 data=payload)
+        assert resp.status_code == 409 # CONFLICT
```
llllllllll: are these error codes not defined somewhere, and if not, should we make a small namespace for them like: `rc.CONFLICT`

kwmsmith:
@llllllllll I split the commit up; 1a7fda0 is the core of the bugfixes, while 80da240 is the style and indentation fixup.

blaze/server/tests/test_server.py
```diff
@@ -660,24 +588,59 @@ def test_cant_add_data_to_server(iris_server, serial):
...
-    assert response1.status_code == 422
+    response1 = iris_server.post('/add',
+                                 headers=mimetype(serial),
+                                 data=blob)
+    assert response1.status_code == RC.UNPROCESSABLE_ENTITY
+
+
+@pytest.mark.parametrize('serial', all_formats)
+def test_add_data_twice_error(empty_server, serial):
+    with temp_server():
```
llllllllll: should this be a pytest.yield_fixture?

blaze/server/tests/test_server.py
```diff
@@ -660,24 +588,59 @@ def test_cant_add_data_to_server(iris_server, serial):
...
+        # add iris
+        iris_path = example('iris.csv')
+        payload = serial.dumps({'iris': iris_path})
+        empty_server.post('/add',
+                          headers=mimetype(serial),
+                          data=payload)
+        # Try to add to existing 'iris'
+        resp = empty_server.post('/add',
+                                 headers=mimetype(serial),
+                                 data=payload)
```
llllllllll: should we use different payloads and assert that even though we got a CONFLICT that we are still served the data from the original payload?

blaze/server/server.py
```diff
@@ -607,21 +584,48 @@ def log_time(start=time()):
...
+        payload_not_mm_msg = ("Need a dictionary-like payload; instead was "
+                              "given %s of type %s.")
+        return (payload_not_mm_msg % (payload, type(payload)),
+                RC.UNPROCESSABLE_ENTITY)
 
-    payload_not_mm_msg = ("Cannot update blaze server with a %s payload, since"
-                          " it is not a mutable mapping (dictionary like).")
-    if not isinstance(payload, collections.MutableMapping):
-        return (payload_not_mm_msg % type(payload), 422)
+    if len(payload) > 1:
```
llllllllll: Why is this blocked, could we not change the code below to be:

```python
for name, resource_uri in payload.items():
    ....
```

blaze/server/tests/test_server.py
kwmsmith: The `CONFLICT` is triggered on the name of the dataset, not the contents; I'm fine if we keep the payload the same.

I'll add the test that the server still serves the original.

blaze/server/server.py
kwmsmith: I blocked this since for now, our only usecases are adding a single dataset at a time.  Generalizing to multiple additions in one `/add` would complicate things, as you point out.

I'm fine generalizing once there's a clear need, but for now, I'm pleading YAGNI :)

llllllllll: That sounds good, I haven't had a use for `/add` yet so I don't mind keeping it minimal

PR #1460, opened 2016-03-24
Allow option to load yaml resources relative to yaml file

kwmsmith:
This builds on PR #1427.

@llllllllll, this should address your comments in that PR thread.  Further thoughts?


blaze/server/spider.py
```diff
@@ -72,28 +73,47 @@ def data_spider(path,
...
+              followlinks=True,
+              hidden=False,
+              relative_to_yaml_dir=False):
     """Construct a dictionary of resources from a YAML specification.
 
     Parameters
     ----------
-    path : str
-        Path to a YAML specification of resources to load
+    path : file
```
llllllllll: this should be something like `f` or `file` if it is a file instead of a path to a file.

blaze/server/spider.py
```diff
@@ -72,28 +73,47 @@ def data_spider(path,
...
+@contextmanager
+def manage_chdir(root):
+    """Context manager that changes to ``root`` directory on ``__entry__`` and
+    changes back to ``os.getcwd()`` on __exit__.
+    """
+    root = os.path.abspath(root)
+    cwd = os.path.abspath(os.getcwd())
+    if root != cwd:
+        os.chdir(root)
+    yield
```
llllllllll: this should be in a try/finally so that if an exception is raised in the `manage_chdir` context we still correctly exit.

blaze/server/spider.py
```diff
@@ -72,28 +73,47 @@ def data_spider(path,
...
+                                            followlinks=followlinks,
+                                            hidden=hidden,
+                                            extra_kwargs=extra_kwargs)}
 
 
-def from_yaml(path, ignore=(ValueError, NotImplementedError), followlinks=True,
-              hidden=False):
+@contextmanager
+def manage_chdir(root):
+    """Context manager that changes to ``root`` directory on ``__entry__`` and
```
llllllllll: this data model method here is `__enter__`; however, I think it is fine to just say on enter or on exit (without the underscores)

blaze/server/spider.py
```diff
@@ -72,28 +73,47 @@ def data_spider(path,
...
+                                            ignore=ignore,
+                                            followlinks=followlinks,
+                                            hidden=hidden,
+                                            extra_kwargs=extra_kwargs)}
 
 
-def from_yaml(path, ignore=(ValueError, NotImplementedError), followlinks=True,
-              hidden=False):
+@contextmanager
+def manage_chdir(root):
```
llllllllll: I think an easier name for this might be `pushd`. `manage` doesn't quite tell me how it is being managed but the pushd/popd analog might be more clear

blaze/server/spider.py
```diff
@@ -72,28 +73,47 @@ def data_spider(path,
...
+    """Context manager that changes to ``root`` directory on ``__entry__`` and
+    changes back to ``os.getcwd()`` on __exit__.
+    """
+    root = os.path.abspath(root)
+    cwd = os.path.abspath(os.getcwd())
```
llllllllll: `os.getcwd` returns the absolute path

blaze/server/spider.py
llllllllll: could we call this `path`, I am not sure what this is the root of.

blaze/server/spider.py
```diff
@@ -72,28 +73,47 @@ def data_spider(path,
...
+    if root != cwd:
```
llllllllll: is this check just an optimization?

kwmsmith:
@llllllllll thanks for the good feedback and quick review.

@brendancol thoughts / input?

PR #1462, opened 2016-03-28
String broadcastables

kwmsmith:
This starts the process of expanding Blaze's string column support to include `upper` and `lower`.  This is useful since having first-class (and optimized) support for common string operations is useful for the string-munging pain points that users hit.

I'm marking these as part of the "experimental" API currently, since I'm not wild about the `str_upper`, `str_lower`, etc. naming scheme.  I'd like to find a better naming system for these if we can.

We have immediate need for `upper` and `lower`, so I'm putting these in for 0.10.

Regarding naming schemes: I like the Pandas' style `df.col.str.upper().str.replace(...)`.  We could expand that to include `df.col.dt.datetimemethod()` for datetimes as well.

We'll have to think about which string and datetime methods we want to support, for which backends, and what are the semantics when the method in question returns multiple values.  All of that is outside the scope for this PR.

This PR also deprecates `strlen` and adds `str_len` for consistency.

### DIFF
--- a/blaze/compute/pandas.py
+++ b/blaze/compute/pandas.py
@@ -286,6 +286,8 @@ def compute_up(expr, data, **kwargs):
 
 string_func_names = {
     'strlen': 'len',
+    'strupper': 'upper',
+    'strlower': 'lower',
 }
 
 
--- a/blaze/compute/pandas.py
+++ b/blaze/compute/pandas.py
@@ -284,11 +284,10 @@ def compute_up(expr, data, **kwargs):
     return compute_up(expr._child.distinct().count(), data, **kwargs)
 
 
-string_func_names = {
-    'strlen': 'len',
-    'strupper': 'upper',
-    'strlower': 'lower',
-}
+string_func_names = {'str_len': 'len',
+                     'strlen': 'len',
+                     'str_upper': 'upper',
+                     'str_lower': 'lower'}
 
 
 @dispatch(UnaryStringFunction, Series)
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -88,6 +88,7 @@ from ..expr import (
     notnull,
     nunique,
     std,
+    str_len,
     strlen,
     var,
 )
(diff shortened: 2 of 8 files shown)


kwmsmith:
ping @ltransom

necaris:
FWIW :+1: for the Pandas style of `df.col.str.*`, especially because that lets us do more with `df.col.dt.*`, `df.col.int.*`, and so on.

llllllllll:
I am also +1 on the `str` and `dt` descriptor namespaces.

sandhujasmine:
Tested on postgres test data in blaze-benchmarks. All looks good.

sandhujasmine:
Starting with [postgresDB](http://www.postgresql.org/docs/9.1/static/functions-string.html) and documenting what's available.
Still need to experiment with functions directly in postgres and SQLA. These are standard SQL string functions so expect these to be present in MySQL but needs verification.

First two already implemented in blaze.

Pandas                  | Postgres            | SQLA                   
-------------------------|-----------------------|--------------------------------------------------------------------
lower()||
upper()||
cat()                       |'Post' \|\| 42         | `sqlalchemy.sql.functions.concat` <br> + should work for string columns, `(table1.col2 + table2.col2)`
                              | bit_length()        |                          
len()                       | char_length()     |
strip, lstrip, rstrip    | trim([leading \| trailing \| both] [characters] from string) |


Replacing a substring can be done using two functions from above:
1. first find the substring position via `position(substring in string)`
2. then use `overlay(string placing substring from int [for int])`

In Postgres, it can also be done using `replace()` function

Additional string functions in Postgres:

Pandas                  | Postgres            | Example    | Result             | SQLA                   
-------------------------|-----------------------|----------------|--------------------|-------------------------------
           | regexp_split_to_array(string text, pattern text [, flags text ])       | regexp_split_to_array('hello world', E'\\s+') | {hello,world}
split()  | regexp_split_to_table(string text, pattern text [, flags text])        | regexp_split_to_table('hello world', E'\\s+') gives two rows | hello <br> world
replace() | replace(string text, from text, to text) | | |

sandhujasmine:
Created an epic to capture this feature but I'm not able to create the epic and add issue to it. Might be a permissions issue? 

Here's the high level epic: #1476
Here's one issue for the fist bullet point: #1475

PR #1467, opened 2016-04-01
ENH: dispatch(Field, collections.Mapping) for compute_up

llllllllll:
I also re-exported broadcast_collect through the top-level api, some tests were importing it from there but I cannot understand where it gets there from. These passed on travis but not locally for me so I assume it was coming from and import * on one of the backends

### DIFF
--- a/blaze/__init__.py
+++ b/blaze/__init__.py
@@ -19,6 +19,7 @@ from .utils import ignoring
 import warnings
 from .expr import (
     Symbol,
+    broadcast_collect,
     by,
     cast,
     coalesce,
--- a/blaze/compute/core.py
+++ b/blaze/compute/core.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
-from collections import defaultdict, Iterator
+from collections import defaultdict, Iterator, Mapping
 from datetime import date, datetime, timedelta
 import itertools
 import numbers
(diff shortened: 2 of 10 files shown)


kwmsmith:
With a little grepping, there are a few other places where we're `@dispatch`ing on `dict` rather than `Mapping`.  Thoughts on converting these as well?

```
$ ag --python '^@dispatch.*dict'
blaze/compute/core.py
381:@dispatch(Expr, dict)
454:@dispatch(Field, dict)

blaze/compute/python.py
771:@dispatch(Field, dict)

blaze/expr/core.py
368:@dispatch((tuple, list), dict)
373:@dispatch(Node, dict)
386:@dispatch(object, dict)

blaze/expr/expressions.py
342:@dispatch(Symbol, dict)

blaze/interactive.py
185:@dispatch(_Data, dict)
```

llllllllll:
I will go over these cases and make sure the conversion makes sense. Also, fun think that I learned while doing this, `collections.abc.Mapping` does not implement `__subclasshook__`, so classes must be explicitly registered or actual subclasses. I think it is because the interface is ambigious with a sequence, for example, `list` implements iter, len, and getitem.

llllllllll:
I also found some isinstance checks that I think are improved by checking against Mapping

llllllllll:
ping @kwmsmith tests are passing

kwmsmith:
LGTM, thanks.

PR #1471, opened 2016-04-01
BUG: use reconstruct_select in sql relabel

mtydykov:
Fixes column relabeling where the relabeling is over other selects.

### DIFF
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -1022,10 +1022,11 @@ def compute_up(expr, data, **kwargs):
     names = data.c.keys()
     assert names == expr._child.fields
     d = dict(zip(names, getattr(data, 'inner_columns', data.c)))
-    return reconstruct_select([d[col].label(new_col) if col != new_col
-                               else d[col]
-                               for col, new_col
-                               in zip(expr._child.fields, expr.fields)], data)
+    return reconstruct_select(
+        (d[col].label(new_col) if col != new_col else d[col]
+         for col, new_col in zip(expr._child.fields, expr.fields)),
+        data,
+    )
 
 
 @dispatch(FromClause)
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -1022,10 +1022,10 @@ def compute_up(expr, data, **kwargs):
     names = data.c.keys()
     assert names == expr._child.fields
     d = dict(zip(names, getattr(data, 'inner_columns', data.c)))
-    return sa.select(
-        d[col].label(new_col) if col != new_col else d[col]
-        for col, new_col in zip(expr._child.fields, expr.fields)
-    )
+    return reconstruct_select([d[col].label(new_col) if col != new_col
+                               else d[col]
+                               for col, new_col
+                               in zip(expr._child.fields, expr.fields)], data)
 
 
 @dispatch(FromClause)
--- a/blaze/compute/tests/test_postgresql_compute.py
+++ b/blaze/compute/tests/test_postgresql_compute.py
@@ -410,10 +410,10 @@ def test_distinct_on(sql):
 
 def test_relabel_columns_over_selection(big_sql):
     t = symbol('t', discover(big_sql))
-    result = compute(t[t['B'] == 2].relabel(B='b'),
+    result = compute(t[t['B'] == 2].relabel(B=u'b'),
                      big_sql, return_type=pd.DataFrame)
-    expected = pd.DataFrame({"A": ['a'], "b": [2]})
-    tm.assert_frame_equal(result.sort(axis=1), expected.sort(axis=1))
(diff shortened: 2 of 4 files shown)


blaze/compute/sql.py
```diff
@@ -1022,10 +1022,10 @@ def compute_up(expr, data, **kwargs):
     names = data.c.keys()
     assert names == expr._child.fields
     d = dict(zip(names, getattr(data, 'inner_columns', data.c)))
-    return sa.select(
-        d[col].label(new_col) if col != new_col else d[col]
-        for col, new_col in zip(expr._child.fields, expr.fields)
-    )
+    return reconstruct_select([d[col].label(new_col) if col != new_col
```
llllllllll: I find ternary expressions harder to read when they are split like this, could we just break earlier like
```python
return reconstruct_select(
    (d[col].label(new_col) if col != new_col else d[col]
     for col, new_col in zipl(...),
    data,
)
```

llllllllll:
Can you add a test to both blaze/compute/test_sql_compute and blaze/compute/test_postgresql_compute

blaze/compute/tests/test_sql_compute.py
```diff
@@ -556,6 +556,16 @@ def test_relabel_table():
...
 
 
+def test_relabel_columns_over_selection():
+    result = compute(t[t['amount'] > 150].relabel(name='NAME'), s, return_type
+    = 'native')
+    expected = str(sa.select(
+        [s.c.name.label('NAME'), s.c.amount, s.c.id]
+    ).where(s.c.amount > 150))
+    import pdb; pdb.set_trace()
+    assert str(result) == str(expected)
```
llllllllll: we should use bz.utils.literal_compile here to make sure that the expression actually matches. We will also want to use `normalize` on the output to ignore whitespace differences.

blaze/compute/tests/test_sql_compute.py
```diff
@@ -556,6 +556,16 @@ def test_relabel_table():
     assert str(result) == str(expected)
 
 
+def test_relabel_columns_over_selection():
+    result = compute(t[t['amount'] > 150].relabel(name='NAME'), s, return_type
```
llllllllll: can we reformat this to not line break on the passing of a kwarg

blaze/compute/tests/test_postgresql_compute.py
```diff
@@ -410,6 +424,17 @@ def test_distinct_on(sql):
     assert odo(computation, tuple) == (('a', 1), ('b', 2))
 
 
+def test_relabel_columns_over_selection(sql_more_data):
```
llllllllll: for the postgresql test we should actually compute this and check the resulting data instead of just checking the sql string that it emits.

blaze/compute/tests/test_postgresql_compute.py
```diff
@@ -65,6 +65,20 @@ def sql(url):
             drop(t)
 
 
+@pytest.yield_fixture
```
llllllllll: is there a reason we need to add a new fixture here?

mtydykov: I wanted more data than existing fixtures have (so that I can filter by it and test relabeling over a selection).

llllllllll: `big_sql` probably has enough data for this test. Also, the dshape and the data do not match up here.

blaze/compute/tests/test_postgresql_compute.py
```diff
@@ -410,6 +408,14 @@ def test_distinct_on(sql):
     assert odo(computation, tuple) == (('a', 1), ('b', 2))
 
 
+def test_relabel_columns_over_selection(big_sql):
+    t = symbol('t', discover(big_sql))
+    result = compute(t[t['B'] == 2].relabel(B='b'),
+                     big_sql, return_type=pd.DataFrame)
+    expected = pd.DataFrame({"A": ['a'], "b": [2]})
+    tm.assert_frame_equal(result.sort(axis=1), expected.sort(axis=1))
```
llllllllll: the column order here should be based on the order of the fields in the dshape. I do not think that we should sort the columns of the `result` dataframe because it could mask bugs where the columns are not in the correct order. we could also be explicit about this with `expected` by using: `pd.DataFrame([['a', 2]], columns=['A', 'b'])`

blaze/compute/tests/test_sql_compute.py
mtydykov: So just to confirm: we're checking both the query and the result, correct?

llllllllll: just the result, we check the query in test_sql.py

mtydykov: Whoops - thought this was a comment for `test_postgresql_compute`. Makes more sense now.

Issue #1241, opened 2015-09-10
?string join with string fails


cpcloud:

cpcloud:
@llllllllll I believe this is happening when promoting to object. Long term it would probably be best to not use numpy for promotion, short term we can special case strings.

llllllllll:
Yeah, the failure mode is that I am checking if `promote(a, b) != object_` because I assumed that meant that no promotion was found. I agree with both rolling our own promotion function and special casing strings for now

kwmsmith:
Closed by #1478.

PR #1478, opened 2016-04-11
Add test for join dshape with string option types.

kwmsmith:
Also add xfail test for `'outer'` join with mixed option / non option
types.

This is to test a case with `join` after blaze/datashape#213 bugfix.


kwmsmith:
Closes #1241.

Issue #1475, opened 2016-04-08
Survey of `str` methods for DB backends similar to Pandas `str` namespace


sandhujasmine:
Starting with [postgresDB](http://www.postgresql.org/docs/9.1/static/functions-string.html) and documenting what's available.
Still need to experiment with functions directly in postgres and SQLA. These are standard SQL string functions so expect these to be present in MySQL but needs verification.

First two already implemented in blaze.

Pandas                  | Postgres            | SQLA                   
-------------------------|-----------------------|--------------------------------------------------------------------
lower()||
upper()||
cat()                       |'Post' \|\| 42         | `sqlalchemy.sql.functions.concat` <br> + should work for string columns, `(table1.col2 + table2.col2)`
                              | bit_length()        |                          
len()                       | char_length()     |
strip, lstrip, rstrip    | trim([leading \| trailing \| both] [characters] from string) |

Replacing a substring can be done using two functions from above:
1. first find the substring position via `position(substring in string)`
2. then use `overlay(string placing substring from int [for int])`

In Postgres, it can also be done using `replace()` function

Additional string functions in Postgres:

Pandas                  | Postgres            | Example    | Result             | SQLA                   
-------------------------|-----------------------|----------------|--------------------|-------------------------------
           | regexp_split_to_array(string text, pattern text [, flags text ])       | regexp_split_to_array('hello world', E'\\s+') | {hello,world}
split()  | regexp_split_to_table(string text, pattern text [, flags text])        | regexp_split_to_table('hello world', E'\\s+') gives two rows | hello <br> world
replace() | replace(string text, from text, to text) | | |

PR #1479, opened 2016-04-13
Add str_cat() to pandas and sql to concatenate string columns

sandhujasmine:
- Added str_cat() to concatenate two string columns with `sep=` kwarg.
- Added tests for pandas and tests to validate the compiled SQL query.
- Added test for postgres backend.

Also, still need to implement str_cat() when inputs are Select as opposed to ColumnElement, which is a bit more complex

### DIFF
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -1208,7 +1208,8 @@ def compute_up(expr, data, **kwargs):
 
 @dispatch(StrCat, ColumnElement, ColumnElement)
 def compute_up(expr, lhs_data, rhs_data, **kwargs):
-    res = (sa.sql.functions.concat(lhs_data, expr.sep,
+    res = (sa.sql.functions.concat(lhs_data,
+                                   expr.sep,
                                    rhs_data).label(lhs_data.name))
     return res
 
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -1207,7 +1207,54 @@ def compute_up(expr, data, **kwargs):
     return sa.sql.functions.char_length(data).label(expr._name)
 
 
-@dispatch(StrCat, (ColumnElement, ColumnElement), (ColumnElement, Select))
+def str_cat_lhs_rhs_data(lhs_data, rhs_data):
+    err = 'cannot concat columns from different tables'
+
+    if isinstance(rhs_data, Select):
+        # lhs should only ever have 1 column
+        # rhs can have more than one columns if chained StrCat?
+        c_rhs = rhs_data.columns.keys()[0]
+
+        if isinstance(lhs_data, ColumnElement):
+            # (ColumnElement, Select)
+            # If RHS is Select, then this becomes the table from which to
+            # concat columns. Append LHS column to RHS so we don't end up with
+            # a cross join.
+            c_lhs = lhs_data.name
+            rhs_data.append_column(lhs_data)
+            alias_ = rhs_data.alias()
+
+            lhs_data = alias_.columns.get(c_lhs)
+
+        else:
+            # lhs could have more than one column if we chain StrCat
+            # still experimenting with this
+            if len(lhs_data.columns.keys()) > 1:
+                msg = ("WIP: this is likely chaining StrCat and has a where "
+                       "clause. SQL is not consistent with pandas result")
+                warnings.warn(msg)
+
+            c_lhs = lhs_data.columns.keys()[0]
+
+            # (Select, Select)
+            c_append = lhs_data.froms[0].columns.get(c_lhs)
+            rhs_data.append_column(c_append)
+
+            alias_ = rhs_data.alias()
+
+        lhs_data = alias_.columns.get(c_lhs)
+        rhs_data = alias_.columns.get(c_rhs)
+
+    else:
+        if lhs_data._from_objects != rhs_data._from_objects:
+            raise ValueError(err)
+
+    return lhs_data, rhs_data
+
+
+@dispatch(StrCat,
+          (ColumnElement, ColumnElement),
+          (ColumnElement, Select))
 def compute_up(expr, lhs_data, rhs_data, **kwargs):
     """
     concat two string columns element wise. If a row in either column is NULL,
@@ -1218,21 +1265,7 @@ def compute_up(expr, lhs_data, rhs_data, **kwargs):
     e.g.
         t.name.str_cat(t.comment.str_cat(t.sex, sep=' -- '), sep=' ++ ')
     """
-    err = 'cannot concat columns from different tables'
-
-    if isinstance(rhs_data, Select):
-        if lhs_data._from_objects != rhs_data._from_objects[0].froms:
-            raise ValueError(err)
-
-        # this gives desired result by mutating the select in place instead
-        # of chaining
-        rhs_data.append_column(lhs_data)
-        a_rhs = rhs_data.alias()
-        rhs_data = a_rhs.columns._all_columns[0]
-        lhs_data = a_rhs.columns._all_columns[1]
-    else:
-        if lhs_data._from_objects != rhs_data._from_objects:
-            raise ValueError(err)
+    lhs_data, rhs_data = str_cat_lhs_rhs_data(lhs_data, rhs_data)
 
     if expr.sep is None:
         # this will automatically handle null values correctly
@@ -1250,6 +1283,35 @@ def compute_up(expr, lhs_data, rhs_data, **kwargs):
                                                        expr.sep,
                                                        rhs_data)
                          ).label(expr.lhs._name)
+        res = select(res)
+    return res
+
+
+@dispatch(StrCat, Select, Select)
+def compute_up(expr, lhs_data, rhs_data, **kwargs):
+    """
+    Assumes: StrCat operates on data from same table
+    Not sure why I cannot just add (Select, Select) to function above
+    """
+    lhs_data, rhs_data = str_cat_lhs_rhs_data(lhs_data, rhs_data)
+
+    if expr.sep is None:
+        # this will automatically handle null values correctly
+        res = (lhs_data + rhs_data).label(expr.lhs._name)
+    else:
+        # only use concat function if both columns have non null values
+        # this is consistent with how pandas works
+        res = \
+            sa_expr.case([
+                          (sa.or_(lhs_data == sa_expr.null(),
+                                  rhs_data == sa_expr.null()),
+                           sa_expr.null())
+                          ],
+                         else_=sa.sql.functions.concat(lhs_data,
+                                                       expr.sep,
+                                                       rhs_data)
+                         ).label(expr.lhs._name)
+
         res = select(res)
     return res
 
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -34,6 +34,7 @@ import sqlalchemy as sa
 from sqlalchemy import sql, Table, MetaData
 from sqlalchemy.engine import Engine
 from sqlalchemy.sql import Selectable, Select, functions as safuncs
+from sqlalchemy.sql import expression as sa_expr
 from sqlalchemy.ext.compiler import compiles
 from sqlalchemy.sql.elements import ClauseElement, ColumnElement, ColumnClause
 from sqlalchemy.sql.selectable import FromClause, ScalarSelect
@@ -1208,9 +1209,28 @@ def compute_up(expr, data, **kwargs):
 
 @dispatch(StrCat, ColumnElement, ColumnElement)
 def compute_up(expr, lhs_data, rhs_data, **kwargs):
-    res = sa.sql.functions.concat(lhs_data,
-                                  expr.sep,
-                                  rhs_data).label(expr.lhs._name)
+    """
+    concat two string columns element wise. If a row in either column is NULL,
+    then return a NULL otherwise perform concat() with the user defined 'sep'
+    between the two elements.
+    """
+    if expr.sep is None:
+        # this will automatically handle null values correctly
+        res = (lhs_data + rhs_data).label(expr.lhs._name)
+    else:
+        # only use concat function if both columns have non null values
+        # this is consistent with how pandas works
+        res = \
+            sa_expr.case([
+                          (sa.or_(lhs_data == sa_expr.null(),
+                                  rhs_data == sa_expr.null()),
+                           sa_expr.null())
+                          ],
+                         else_=sa.sql.functions.concat(lhs_data,
+                                                       expr.sep,
+                                                       rhs_data)
+                         ).label(expr.lhs._name)
+        res = select(res)
     return res
 
 
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -1253,7 +1253,7 @@ def str_cat_lhs_rhs_data(lhs_data, rhs_data):
 
 
 @dispatch(StrCat,
-          (ColumnElement, ColumnElement),
+          (Select, ColumnElement),
           (ColumnElement, Select))
 def compute_up(expr, lhs_data, rhs_data, **kwargs):
     """
@@ -1267,34 +1267,8 @@ def compute_up(expr, lhs_data, rhs_data, **kwargs):
     """
     lhs_data, rhs_data = str_cat_lhs_rhs_data(lhs_data, rhs_data)
 
-    if expr.sep is None:
-        # this will automatically handle null values correctly
-        res = (lhs_data + rhs_data).label(expr.lhs._name)
-    else:
-        # only use concat function if both columns have non null values
-        # this is consistent with how pandas works
-        res = \
-            sa_expr.case([
-                          (sa.or_(lhs_data == sa_expr.null(),
-                                  rhs_data == sa_expr.null()),
-                           sa_expr.null())
-                          ],
-                         else_=sa.sql.functions.concat(lhs_data,
-                                                       expr.sep,
-                                                       rhs_data)
-                         ).label(expr.lhs._name)
-        res = select(res)
-    return res
-
-
-@dispatch(StrCat, Select, Select)
-def compute_up(expr, lhs_data, rhs_data, **kwargs):
-    """
-    Assumes: StrCat operates on data from same table
-    Not sure why I cannot just add (Select, Select) to function above
-    """
-    lhs_data, rhs_data = str_cat_lhs_rhs_data(lhs_data, rhs_data)
-
+    # try following to get rid of case statement below
+    # res = (lhs_data + expr.sep + rhs_data).label(expr.lhs._name)
     if expr.sep is None:
         # this will automatically handle null values correctly
         res = (lhs_data + rhs_data).label(expr.lhs._name)
@@ -1312,7 +1286,8 @@ def compute_up(expr, lhs_data, rhs_data, **kwargs):
                                                        rhs_data)
                          ).label(expr.lhs._name)
 
-        res = select(res)
+    res = select(res)
+
     return res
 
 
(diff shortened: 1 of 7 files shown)


blaze/compute/tests/test_pandas_compute.py
```diff
@@ -680,6 +680,21 @@ def test_str_lower():
...
+    df = pd.DataFrame(data, columns=['name', 'comment', 'num'])
+
+    def test_str_cat(self):
+        res = compute(self.s.name.str_cat(self.s.comment), self.df)
+        assert all(self.df.name.str.cat(self.df.comment) == res)
+
+    def test_str_cat_sep(self):
+        res = compute(self.s.name.str_cat(self.s.comment, sep=' -- '), self.df)
+        assert all(self.df.name.str.cat(self.df.comment, sep=' -- ') == res)
+
```
kwmsmith: Can you refactor this to use `pytest.fixture`s and top-level functions?   e.g.:

```python
@pytest.fixture(scope='module')
def strcat_sym():
    return symbol('s', dshape=...)

@pytest.fixture(scope='module')
def strcat_df():
    return pd.DataFrame(...)

def test_str_cat(strcat_sym, strcat_df):
    ...
def test_str_cat_sep(strcat_sym, strcat_df):
    ...
```

blaze/compute/tests/test_pandas_compute.py
```diff
@@ -680,6 +680,21 @@ def test_str_lower():
     assert_series_equal(expected, result)
 
 
+class TestStrCat():
+    ds = dshape('3 * {name: string[10], comment: string[25], num: int32}')
```
kwmsmith: Can you make one of these string columns variable length (`string` rather than `string[25]`)?

blaze/expr/tests/test_strings.py
```diff
@@ -28,3 +27,16 @@ def test_str_upper_schema(ds):
     assert (expr_upper.schema.measure ==
             expr_lower.schema.measure ==
             dshape('%sstring' % ('?' if '?' in ds else '')).measure)
+
```
kwmsmith: We should add tests for the following cases:
* Calling the functional form (rather than the method form) of `str_cat()` works and does proper error checking for both the lhs and rhs types / dshapes,
* test the resulting dshape for str_cat, specifically, what happens when strcat is passed in option or non-option string columns?

blaze/expr/strings.py
```diff
@@ -74,6 +84,54 @@ def schema(self):
...
+class StrCat(ElemWise):
+    """
+    Concatenate two string columns together with optional 'sep' argument.
+
+    >>> import pandas as pd
+    >>> from blaze import symbol, compute, dshape
+
+    >>> ds = dshape('3 * {name: string[10], comment: string[25], num: int32}')
+    >>> s = symbol('s', dshape=ds)
+    >>> data = [('al', 'good', 0), ('suri', 'not good', 1), ('jinka', 'ok', 2)]
```
kwmsmith: Can you add a row with `None` in there for one of the strings, to show what happens when a string is missing, and that `str_cat()` propagates `None` in this case?

blaze/compute/tests/test_pandas_compute.py
kwmsmith: As mentioned elsewhere, please add a row with one or both of the strings set to `None`, to test `str_cat` on missing data.

blaze/compute/sql.py
```diff
@@ -1205,6 +1206,14 @@ def compute_up(expr, data, **kwargs):
     return sa.sql.functions.char_length(data).label(expr._name)
 
 
+@dispatch(StrCat, ColumnElement, ColumnElement)
+def compute_up(expr, lhs_data, rhs_data, **kwargs):
+    res = (sa.sql.functions.concat(lhs_data,
+                                   expr.sep,
+                                   rhs_data).label(lhs_data.name))
```
kwmsmith: nit: you don't need the outer `()`s here...

blaze/expr/strings.py
```diff
@@ -74,6 +84,54 @@ def schema(self):
...
+    def _dshape(self):
+        '''
+        since pandas supports concat for string columns, do the same for blaze
+        '''
+        shape, schema = self.lhs.dshape.shape, DataShape(String())
+        return DataShape(*(shape + (schema,)))
+
+
+@copydoc(StrCat)
+def str_cat(expr, to_concat, sep=None):
```
kwmsmith: To emphasize the symmetry and to match the corresponding `StrCat` instance attributes, can you rename `expr` to `lhs` and `to_concat` to `rhs`?

kwmsmith:
Another test to add to both the pandas and SQL backends: verify that chaining `str_cat()`s works as expected: `t.a.str_cat(t.b).str_cat(t.c)` -- since strcat is binary as we've implemented it, this is how users would accomplish n-ary concatenation.

llllllllll:
How is `t.a.str_cat(t.b)` different than `t.a + t.b`? also, as a seperate could we have some optimize pass to fold those chained concats (this should also fold chained `bz.concat` nodes). Right now chaining these calls has really bad runtime

sandhujasmine:
It's the same as `t.a + t.b` except it excepts and adds the `sep` kwarg similar to pandas `str.cat()`; but the motivation was to add this to the `str` namespace similar to pandas. Per the comments in #1462 - there was consensus on wanting to implement pandas style `str` and `dt` namespace. This is just implementing the `cat()` feature of the pandas str namespace.

We still need an `str` namespace but started with implementing some more of the functions that will eventually go here. `str_upper, str_lower, str_cat,` will eventually be in str namespace and loose the `str_` prefix.

Here's a survey of some of the methods in the `str` namespace and how they maybe implemented in blaze #1475 -- still working on it.

blaze/compute/tests/test_pandas_compute.py
```diff
@@ -45,17 +45,12 @@
...
-
-    df_add_null = dfbig.append(DataFrame([row],
+@pytest.fixture(scope='module')
+def df_add_null():
+    rows = [(None, 'M', 300, 6),
+            ('first', None, 300, 6),
+            (None, None, 300, 6)]
+    df_add_null = dfbig.append(DataFrame(rows,
                                          columns=dfbig.columns
                                          ), ignore_index=True)
```
sandhujasmine: Add NULL data to rows and test.

blaze/compute/tests/test_pandas_compute.py
```diff
@@ -680,20 +697,23 @@ def test_str_lower():
...
 
 
-class TestStrCat():
-    ds = dshape('3 * {name: string[10], comment: string[25], num: int32}')
-    s = symbol('s', dshape=ds)
-    data = [('alice', 'this is good', 0), ('suri', 'this is not good', 1),
-            ('jinka', 'this is ok', 2)]
-    df = pd.DataFrame(data, columns=['name', 'comment', 'num'])
+def test_str_cat():
+    res = compute(tbig.name.str_cat(tbig.sex), dfbig)
```
sandhujasmine: Use `dfbig` existing dataframe for tests. This includes a variable length and fixed length string in the dshape:
`{name: string, sex: string[1], amount: int, id: int}`

blaze/compute/tests/test_pandas_compute.py
sandhujasmine: Reused `dfbig` existing dataframe - also added a fixture for testing null values in rows. See latest changes.

blaze/compute/tests/test_pandas_compute.py
sandhujasmine: done.

blaze/compute/tests/test_pandas_compute.py
```diff
@@ -45,17 +45,12 @@
...
-    else:
-        row = [None, None, 300, 6]
-
-    df_add_null = dfbig.append(DataFrame([row],
+@pytest.fixture(scope='module')
+def df_add_null():
+    rows = [(None, 'M', 300, 6),
+            ('first', None, 300, 6),
+            (None, None, 300, 6)]
+    df_add_null = dfbig.append(DataFrame(rows,
```
sandhujasmine: Add NULL data to rows and test.

blaze/expr/strings.py
```diff
@@ -110,7 +110,15 @@ def _dshape(self):
...
+        shape = self.lhs.dshape.shape
+        if isinstance(self.lhs.schema.measure, Option):
+            schema = self.lhs.schema
+        elif isinstance(self.rhs.schema.measure, Option):
+            schema = self.rhs.schema
+        else:
+            # convert fixed length string to variable length string
+            schema = DataShape(String())
+
         return DataShape(*(shape + (schema,)))
```
sandhujasmine: Fixed schema of output as follows:
- if `lhs` or `rhs` contain NULL values, then result will contain NULL
- if neither contain NULL values, then just return String() dshape

blaze/expr/tests/test_strings.py
```diff
@@ -29,14 +46,19 @@ def test_str_upper_schema(ds):
...
 
-class TestStrCatExceptions():
-    ds = dshape('3 * {name: string[10], comment: string[25], num: int32}')
-    s = symbol('s', dshape=ds)
+@pytest.mark.parametrize('ds', lhsrhs_ds)
+def test_str_schema(ds):
+    t = symbol('t', ds)
+    expr = t.name.str_cat(t.comment)
+    assert (expr.schema.measure ==
+            dshape('%sstring' % ('?' if '?' in ds else '')).measure)
```
sandhujasmine: Add parametrized test to check the resulting schema is as expected.

blaze/expr/tests/test_strings.py
```diff
@@ -29,14 +46,19 @@ def test_str_upper_schema(ds):
...
-    def test_str_cat_exception_non_string_sep(self):
-        with pytest.raises(TypeError):
-            self.s.name.str_cat(self.s.comment, sep=123)
 
-    def test_str_cat_exception_non_string_col_to_cat(self):
-        with pytest.raises(TypeError):
-            self.s.name.str_cat(self.s.num)
+def test_str_cat_exception_non_string_col_to_cat(strcat_sym):
+    with pytest.raises(TypeError):
+        strcat_sym.name.str_cat(strcat_sym.num)
```
sandhujasmine: Updated tests to check exception raised by str_cat() function.

blaze/compute/sql.py
```diff
@@ -1208,9 +1209,28 @@ def compute_up(expr, data, **kwargs):
...
-                                  expr.sep,
-                                  rhs_data).label(expr.lhs._name)
+    """
+    concat two string columns element wise. If a row in either column is NULL,
+    then return a NULL otherwise perform concat() with the user defined 'sep'
+    between the two elements.
+    """
+    if expr.sep is None:
+        # this will automatically handle null values correctly
+        res = (lhs_data + rhs_data).label(expr.lhs._name)
```
sandhujasmine: Use the `+` to concatenate if no sep - it's easier to read. I assume it is faster since SQL doesn't have to process the case statement but mostly added this so we can clearly see what the simple case should be doing.

blaze/compute/sql.py
```diff
@@ -1208,9 +1209,28 @@ def compute_up(expr, data, **kwargs):
...
+                          (sa.or_(lhs_data == sa_expr.null(),
+                                  rhs_data == sa_expr.null()),
+                           sa_expr.null())
+                          ],
+                         else_=sa.sql.functions.concat(lhs_data,
+                                                       expr.sep,
+                                                       rhs_data)
+                         ).label(expr.lhs._name)
+        res = select(res)
     return res
```
sandhujasmine: Since `concat()` ignores NULL values, need a case statement here to only use `concat()` if both `lhs` and `rhs` are not null. This result is consistent with pandas for the `na_rep=None` default.

*todo: extend `str_cat()` to include na_rep kwarg*

blaze/expr/strings.py
sandhujasmine: Done. I was trying to describe what `lhs` and `rhs` indicate but I guess that's more clear.

blaze/expr/tests/test_strings.py
```diff
@@ -47,11 +48,12 @@ def test_str_upper_schema(ds):
 
 
 @pytest.mark.parametrize('ds', lhsrhs_ds)
-def test_str_schema(ds):
+def test_str_cat_schema_shape(ds):
     t = symbol('t', ds)
     expr = t.name.str_cat(t.comment)
     assert (expr.schema.measure ==
             dshape('%sstring' % ('?' if '?' in ds else '')).measure)
+    assert expr.lhs.shape == expr.rhs.shape == expr.shape
```
sandhujasmine: - tests shape and schema of the result

blaze/expr/tests/test_strings.py
sandhujasmine: Done.

blaze/expr/strings.py
```diff
@@ -74,6 +84,64 @@ def schema(self):
...
+    Name: name, dtype: object
+
+    """
+    __slots__ = '_hash', 'lhs', 'rhs', 'sep'
+    __inputs__ = 'lhs', 'rhs'
+
+    def _dshape(self):
+        '''
+        since pandas supports concat for string columns, do the same for blaze
+        '''
```
kwmsmith: There's a `datashape.promote()` utility function here that's worth looking into.  It should handle this logic for you, although it's worth testing with a mixture of fixed / variable width strings:

```python
datashape.promote(lhs.dshape, rhs.dshape)
```

blaze/expr/tests/test_strings.py
```diff
@@ -10,6 +9,24 @@
            'var * ?string',
            'string']
 
+lhsrhs_ds = ['var * {name: string, comment: string[25]}',
+             'var * {name: string[10], comment: string}',
+             'var * {name: string, comment: string}',
+             'var * {name: ?string, comment: string}',
+             'var * {name: string, comment: ?string}',
+             '10 * {name: string, comment: ?string}']
```
kwmsmith: This is great -- looks like you covered a lot of cases here.

kwmsmith:
@sandhujasmine It's looking good!  Looks like the biggest remaining piece is the `Select, Select` case for SQL.  How's that coming?

sandhujasmine:
> Another test to add to both the pandas and SQL backends: verify that chaining str_cat()s works as 
> expected: t.a.str_cat(t.b).str_cat(t.c) -- since strcat is binary as we've implemented it, this is how users > would accomplish n-ary concatenation.

This is currently failing for SQL - debugging and trying to figure out the issue.

blaze/compute/sql.py
```diff
@@ -1207,13 +1207,25 @@ def compute_up(expr, data, **kwargs):
...
+        t.name.str_cat(t.comment.str_cat(t.sex, sep=' -- '), sep=' ++ ')
     """
+    if isinstance(rhs_data, Select):
+        # this gives desired result by mutating the select in place instead
+        # of chaining
+        rhs_data.append_column(lhs_data)
+        a_rhs = rhs_data.alias()
+        rhs_data = a_rhs.columns._all_columns[0]
+        lhs_data = a_rhs.columns._all_columns[1]
+
```
sandhujasmine: Added code and tests for chaining str_cat() as: t.a.str_cat(t.b).str_cat(t.c)

Still want to look at the SQLA Select object's attributes a bit more closely but tests are passing so it is one way to do it.

sandhujasmine:
> It's looking good! Looks like the biggest remaining piece is the Select, Select case for SQL. How's that coming?

I missed this comment - I'll work on this next. This will solve following case:
```
t = s[s.amount >= 200]

# need to implement compute_up for (Select, Select) in the following case
expr = t.name.str_cat(t.comment.str_cat(t.sex, sep=' -- '), sep=' ++ ')
```

sandhujasmine:
In working on the (Select, Select) case, learned a few things:

1. Currently implementation gives inconsistent results between pandas and SQL if `str_cat()` used on different tables. In pandas, `res = s.name.str.cat(s.comment)` produces the same dimension result so `res.shape == s.name.shape`; while SQL does a cross join so `len(res) = len(s.name) * len(s.comment)`

2. This is also the case for (Select, Select) example above.

3. Also need a test for chaining str_cat() on different tables.

Q: Should str_cat() work (as pandas does) when used on different tables or should it raise an error requiring a join before str_cat()?

blaze/compute/tests/test_sql_compute.py
```diff
@@ -860,6 +860,7 @@ def test_str_cat(sep):
     assert normalize(result) == normalize(expected)
 
 
+@pytest.mark.xfail(reason="raise exception code needed for refactored StrCat")
 @pytest.mark.parametrize("expr",
                          [t.name.str_cat(t_str_cat.comment),
                           t.name.str_cat(t_str_cat.comment.str_cat(t_str_cat.name))])
```
sandhujasmine: This is currently failing since the code was refactored but the checks to ensure the data is coming from same table are not in place yet. Will fix that after resolving the issue with chaining StrCat containing a where clause. This is just a reminder.

sandhujasmine:
Chaining StrCat() with WHERE clause gives inconsistent results from pandas with current implementation. Create a table 'smp01' in to test DB with following data:
```
def create_table(url='postgresql://localhost/test::smp01'):
    ds = dshape(""" var * {name: ?string,
                       sex: ?string,
                       amount: int,
                       id: int,
                       comment: ?string}
              """)
    rows = [('Alice', 'F', 100, 1, 'Alice comment'),
            (None, 'M', 300, 2, None),
            ('Drew', 'F', 100, 4, 'Drew comment'),
            ('Bob', 'M', 100, 5, 'Bob comment 2'),
            ('Drew', 'M', 200, 5, None),
            ('first', None, 300, 4, 'Missing info'),
            (None, None, 300, 6, None)]

    t = data(url, dshape=ds)
    t = data(odo(rows, t))

    return t


create_table()
```

blaze expression to chain StrCat:

```
ds2 = dshape(""" var * {name: ?string,
                        sex: ?string,
                        amount: int,
                        id: int,
                        comment: ?string}
               """)

r1 = resource('postgresql://localhost/test::smp01', dshape=ds2)

s = symbol('s', ds2)

# Chain str_cat without WHERE clause works
compute(s.name.str_cat(s.comment.str_cat(s.sex)), r1, return_type=pd.DataFrame)
# Out[6]: 
#                   name
# 0  AliceAlice commentF
# 1                 None
# 2    DrewDrew commentF
# 3    BobBob comment 2M
# 4                 None
# 5                 None
# 6                 None

t = s[s.amount <= 200]

# str_cat with WHERE clause seems ok
compute(t.comment.str_cat(t.sex), r1, return_type=pd.DataFrame)
# Out[7]: 
#           comment
# 0  Alice commentF
# 1   Drew commentF
# 2  Bob comment 2M
# 3            None

# chaining str_cat with WHERE clause is broken -- this gives too many results
expr = t.name.str_cat(t.comment.str_cat(t.sex))
```
Looking at the last `expr = t.name.str_cat(t.comment.str_cat(t.sex))`, the current logic compiles to following SQL:
```
SELECT 
    anon_1.name || anon_1.comment AS name 

FROM 
    (SELECT 
        anon_2.comment || anon_2.sex AS comment,
        smp01.name AS name 
     FROM 
        (SELECT 
            smp01.sex AS sex,
            smp01.comment AS comment 
         FROM 
            smp01 
         WHERE 
            smp01.amount <= 200) AS anon_2,
            smp01) AS anon_1

```
Here's one correct SQL expression:

```
SELECT 
    anon_1.name || anon_1.comment AS name 

FROM 
    (SELECT 
        anon_2.name,
        anon_2.comment || anon_2.sex AS comment
     FROM 
        (SELECT 
            smp01.name AS name,
            smp01.sex AS sex,
            smp01.comment AS comment 
         FROM 
            smp01 
         WHERE 
            smp01.amount <= 200) AS anon_2) AS anon_1
```

Constructing a new select statement with the correct columns + where clause is another approach. There is a `_whereclause` attribute in `Select`. I experimented a bit with this but didn't dig too far -- everything is assumed to come from same table so conceptually this does make sense.

@kwmsmith - getting cross-eyed looking at this - will look for your insights here. Thanks!

kwmsmith:
Closing this PR as it's subsumed by #1496.

PR #1481, opened 2016-04-15
Make Blaze server `/add` endpoint more flexible

necaris:
Allow the `/add` endpoint of the Blaze server to take a more complex
payload. Rather than simply a name and resource URI string, the server
can now accept a dictionary with 'source' and 'extra' fields. The
'source' is the URI string as before, but the 'extra' field can be a
dictionary of keyword arguments that are passed through to the
`blaze.resource` constructor.

For instance, in the test case the `encoding` argument is passed through
to allow reading a CSV in a non-default encoding.


blaze/server/server.py
```diff
@@ -607,14 +607,22 @@ def addserver(payload, serial):
...
     if name in data:
         msg = "Cannot add dataset named %s, already exists on server."
         return (msg % name, RC.CONFLICT)
 
     try:
-        data.update({name: resource(resource_uri)})
+        if isinstance(resource_info, dict):
+            # Extract resource creation arguments
+            source = resource_info['source']
+            extra = resource_info.get('extra', {})
```
kwmsmith: It's an edge case, but some `resource()` implementations in odo do take both positional and kwargs, and treat the positional args in a special way.  For instance:
    https://github.com/blaze/odo/blob/master/odo/backends/sql.py#L545

I also realize that the original code didn't handle this case, but this is an opportunity to do it correctly :)

Can you refactor this so that the `resource_info` has both an `args` sequence and a `kwargs` dict / mapping, ala:

```python
source = ...
args = resource_info.get('args', [])
kwargs = resource_info.get('kwargs', {})
...
data.update({name: resource(source, *args, **kwargs)})
```

blaze/server/tests/test_server.py
```diff
@@ -612,6 +612,61 @@ def test_bad_add_payload(temp_server, serial):
 
 
 @pytest.mark.parametrize('serial', all_formats)
+def test_add_expanded_payload(temp_server, serial):
+    # Ensure that the expanded payload format is accepted by the server
+    iris_path = example('iris.csv')
+    blob = serial.dumps({
+        'iris': {
+            'source': iris_path,
+            'extra': {'delimiter': ','}
```
kwmsmith: As mentioned above, can you rename this to `kwargs` rather than `extra`?

blaze/server/tests/test_server.py
```diff
@@ -612,6 +612,61 @@ def test_bad_add_payload(temp_server, serial):
...
+    # Ensure that the expanded payload format actually passes the arguments
+    # through to the resource constructor
+    iris_path = example('iris-latin1.tsv')
+    csv_kwargs = {'delimiter': '\t', 'encoding': 'iso-8859-1'}
+    blob = serial.dumps({
+        'iris': {
+            'source': iris_path,
+            'extra': csv_kwargs
+        }
+    })
```
kwmsmith: Style nit (here and above) please reformat with the following style:

```python
blob = serial.dumps({'iris': {'source': iris_path,
                              'extra': csv_kwargs}})
```

We've settled on this indentation style for Blaze going forward, and are gradually updating files to conform.

necaris:
@kwmsmith Thanks for the quick review! Addressed your comments -- added `args`, renamed `extra`, and updated styles :smile:

kwmsmith:
While we're on the topic, @necaris, one gap here is that there's no `.add()` method on the client side, which is a natural extension to this.  Could you look into adding that method as well?  It looks to be straightforward, but if not, feel free to say so and we can defer to later.

necaris:
@kwmsmith Is https://github.com/blaze/blaze/pull/1481/commits/74416c6cbc6233551a4c4645cdee62ec9c33f786 what you had in mind?

necaris:
@kwmsmith FYI, I've added more tests as well as support for an `imports` key in the payload dictionary, to bring it up to full parity with the server YAML file.

llllllllll:
I am sort of afraid of having `/add` enabled by defaut, especially now that users can send imports. This adds a lot of attack surface to the blaze server. In our blaze server deployment we specifically needed to wrap out dataset in a mappingproxy to prevent datasets from being added, I think it might be nice to have a flag to `Server`/`blaze_api`.` that turns this feature off.

kwmsmith:
> I think it might be nice to have a flag to Server/blaze_api.` that turns this feature off.

@llllllllll good to know, and good input.  So would an `allow_add` flag that's by default `False`, and that's overridable when firing up the blaze server instance, address this adequately?

necaris:
@llllllllll @kwmsmith I've added an `allow_add` flag to the `blaze.server.Server` constructor -- it defaults to `False`, and if it's not set then the `/add` route is not defined _at all_. This results in requests for `/add` returning a 404 Not Found error.

It might be more appropriate to return a Forbidden error, possibly with a message body indicating that the server instance doesn't support adding datasets -- thoughts?

kwmsmith:
@necaris looks good -- the only addition is to add a commandline argument to the `blaze-server` entry point to flip the `allow_add` bit:

    https://github.com/blaze/blaze/blob/master/blaze/server/spider.py#L148-L169

necaris:
@kwmsmith Added!

blaze/server/spider.py
```diff
@@ -164,6 +164,8 @@ def _parse_args():
                    help='Call ``blaze.data`` on hidden files')
     p.add_argument('--yaml-dir', action='store_true',
                    help='Load path-based resources relative to yaml file directory.')
+    p.add_argument('--allow-dynamic-addition', action='store_true',
```
kwmsmith: It's good this is a long option name to make it less likely to trigger accidentally.

necaris: :smile: I considered `-a` but I figured that would be too easy to do by mistake!

kwmsmith:
@necaris LGTM, thanks!

PR #1482, opened 2016-04-16
BUG: Fix test_mysql_compute.py::test_agg_sql expected value

pratapvardhan:
As of today, mysql tests are skipped on travis. Locally, `test_mysql_compute.py::test_agg_sql` fails, this PR fixes it valid input for `normalize` function.

Any specific reason for not validating mysql tests on travis?


pratapvardhan:
Does this seem fine, reviewers?

llllllllll:
I think the issue is just getting mysql setup on travis. I have no easy way to test this myself, but this change seems correct.

PR #1490, opened 2016-04-25
Accept more flexible content-type specification

### LINKED ISSUE #1489: Blaze Server rejects valid Content-Types with option information
Firefox <44 sends an additional `charset=` option with the content-type header, even when it's not required (see the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=918742)) -- this, or any other option, makes the content-type unrecognizable by Blaze server.

necaris:
Closes #1489. The `Content-Type` header can contain option information in a semicolon-separated portion after the main content-type string. This PR allows those more elaborate headers to be accepted by the server as valid serialization formats.

To make this more extensible in future, it might be a good idea to use a helper function such as that [available in Werkzeug](http://werkzeug.pocoo.org/docs/0.11/http/#werkzeug.http.parse_options_header) to handle this.

### DIFF
--- a/blaze/server/server.py
+++ b/blaze/server/server.py
@@ -486,7 +486,7 @@ def from_tree(expr, namespace=None):
         return expr
 
 
-mimetype_regex = re.compile(r'^application/vnd\.blaze\+(%s)$' %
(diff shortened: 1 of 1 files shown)


kwmsmith:
@necaris LGTM.

PR #1491, opened 2016-04-26
Consistency check at blaze-server startup.

kwmsmith:
Make yaml file default to `None` -- expectation is that if user doesn't
specify a yaml file explicitly, will use dynamic addition.

Adds consistency check at server startup.


kwmsmith:
@necaris This is a small tweak / fix to blaze server -- could you take a look?

necaris:
@kwmsmith Looks good! 👍

PR #1492, opened 2016-04-26
Brain dead server logging

kwmsmith:
First cut at logging blaze server.  Adds a few cmdline flags for setup and initialization.

Of note is capturing the full traceback when `/compute` hits an exception, which is the primary thrust of this PR.

Comments and suggestions for improvement welcome.

### DIFF
--- a/blaze/server/server.py
+++ b/blaze/server/server.py
@@ -1,5 +1,9 @@
 from __future__ import absolute_import, division, print_function
 
+import logging
+from logging import Formatter
+from functools import wraps
+import traceback
 import collections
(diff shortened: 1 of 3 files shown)


necaris:
@kwmsmith It may be a good idea to default to logging to `stdout` -- a lot of deployment environments like to manage logging on their own, e.g. I know Supervisor does this.

necaris:
@kwmsmith since #1521 supersedes this, can we close this in its favor?

kwmsmith:
Closing in favor of #1521.

PR #1496, opened 2016-04-28
str_cat with Pandas `.str.cat()` interface

kwmsmith:
* Ensured that chaining `.str_cat().str_cat()` rolls up concatenation ops in SQL query.
* Will be refactored as `.str.cat()` when `.str` accessor namespace is implemented.
* Tested against Pandas / SQL / postgres backends.

ping @sandhujasmine


blaze/compute/sql.py
```diff
@@ -1185,7 +1188,7 @@ def compute_up(t, s, **kwargs):
 string_func_names = {
     # <blaze function name>: <SQL function name>
     'str_upper': 'upper',
-    'str_lower': 'lower',
+    'str_lower': 'lower'
```
llllllllll: If we are making a style edit, should this just be:

```python
string_func_names = {'str_upper': upper,
                     'str_lower': lower}
```

blaze/expr/strings.py
```diff
@@ -74,6 +84,75 @@ def schema(self):
...
+        since pandas supports concat for string columns, do the same for blaze
+        '''
+        shape = self.lhs.dshape.shape
+        if isinstance(self.lhs.schema.measure, Option):
+            schema = self.lhs.schema
+        elif isinstance(self.rhs.schema.measure, Option):
+            schema = self.rhs.schema
+        else:
+            # convert fixed length string to variable length string
+            schema = DataShape(String())
```
llllllllll: should we preserve encoding?

blaze/compute/sql.py
```diff
@@ -1205,6 +1208,43 @@ def compute_up(expr, data, **kwargs):
...
+        )
+        rhs_col = first(rhs.inner_columns)
+    else:
+        rhs_col = rhs
+    if expr.sep:
+        sel = select((lhs_col + expr.sep + rhs_col).label(expr.lhs._name))
+    else:
+        sel = select((lhs_col + rhs_col).label(expr.lhs._name))
+    wheres = unify_wheres([lhs, rhs])
+    if wheres is not None:
```
llllllllll: I think we need to use `reconstruct_select` here because if the comes from a select with something like a groupby or sort this information will be lost.

kwmsmith:
Thanks for the pointers @llllllllll -- `reconstruct_select()` was exactly what I was looking for; it's been a while since I've touched the sql stuff.

blaze/compute/tests/test_sql_compute.py
```diff
@@ -809,14 +823,71 @@ def test_str_upper():
...
+               """
+    assert normalize(result) == normalize(expected)
+
+
+
+@pytest.mark.xfail(reason="raise exception code needed for refactored StrCat")
+@pytest.mark.parametrize("expr",
+                         [t.name.str_cat(t_str_cat.comment),
+                          t.name.str_cat(t_str_cat.comment.str_cat(t_str_cat.name))])
+def test_str_cat_runtime_exception(expr):
```
sandhujasmine: Are we checking that the str_cat operates on the same table? If not, then we should take this test out - I saw a comment that we need to do this check more consistently throughout other blaze operations so perhaps remove the test?

PR #1497, opened 2016-04-29
Allow map/apply in client/server when explicitly enabled

sandhujasmine:
Add serialization for builtins. Addresses the tests in PR #1493 
    
- added json_dumps and object_hook for python builtin functions
- data_loads is not consistent for all serializations. For fastmsgpack, it uses pandas to decode to a Series; while other methods decode to a list.

### DIFF
--- a/blaze/server/serialization.py
+++ b/blaze/server/serialization.py
@@ -226,6 +226,11 @@ fastmsgpack = SerializationFormat(
 )
 
 
+most_formats = frozenset(
+    g for _, g in globals().items() if isinstance(g, SerializationFormat) and
+    g is not fastmsgpack
+)
+
 all_formats = frozenset(
     g for _, g in globals().items() if isinstance(g, SerializationFormat)
 )
--- a/blaze/server/serialization.py
+++ b/blaze/server/serialization.py
@@ -1,4 +1,3 @@
-import sys
 from functools import partial
 import json as json_module
 
--- a/blaze/server/tests/test_server.py
+++ b/blaze/server/tests/test_server.py
@@ -458,29 +458,39 @@ def test_map_numpy_client_server_fastmsgpack(iris_server):
 
 @pytest.mark.xfail(reason="pickle does not produce same error")
 @pytest.mark.parametrize('serial', all_formats)
-def test_map_builtin_403_exception(iris_server, serial):
+def test_builtin_403_exception(iris_server, serial):
+    '''
+    ensure exception is raised when both map and apply are invoked.
+    exception is raised in check_request() when object_hook is invoked;
+    this is when the payload is loaded from the bytes object in reqeust.data
+    '''
     t = symbol('t', discover(iris))
 
-    expr = t.species.map(eval, 'str')
-    query = {'expr': to_tree(expr)}
-    response = iris_server.post('/compute',
-                                data=serial.dumps(query),
-                                headers=mimetype(serial))
+    for name in ('map', 'apply'):
+        func = getattr(t.species, name)
+        expr = func(eval, 'int')
+        query = {'expr': to_tree(expr)}
+        response = iris_server.post('/compute',
+                                    data=serial.dumps(query),
+                                    headers=mimetype(serial))
 
-    assert '403 FORBIDDEN'.lower() in response.status.lower()
+        assert '403 FORBIDDEN'.lower() in response.status.lower()
 
 
 @pytest.mark.xfail(reason="pickle does nto produce same error")
 @pytest.mark.parametrize('serial', all_formats)
-def test_map_builtin_501_exception(iris_server, serial):
+def test_builtin_501_exception(iris_server, serial):
     t = symbol('t', discover(iris))
-    expr = t.species.map(copy, 'str')
-    query = {'expr': to_tree(expr)}
-    response = iris_server.post('/compute',
-                                data=serial.dumps(query),
-                                headers=mimetype(serial))
 
-    assert '501 Not Implemented'.lower() in response.status.lower()
+    for name in ('map', 'apply'):
+        func = getattr(t.species, name)
+        expr = func(copy, 'int')
+        query = {'expr': to_tree(expr)}
+        response = iris_server.post('/compute',
+                                    data=serial.dumps(query),
+                                    headers=mimetype(serial))
+
+        assert '501 Not Implemented'.lower() in response.status.lower()
 
 
 @pytest.mark.parametrize('serial', most_formats)
--- a/blaze/server/tests/test_server.py
+++ b/blaze/server/tests/test_server.py
@@ -23,7 +23,7 @@ from blaze.utils import example
 from blaze import discover, symbol, by, CSV, compute, join, into, data
 from blaze.server.client import mimetype
 from blaze.server.server import Server, to_tree, from_tree, RC
-from blaze.server.serialization import all_formats, fastmsgpack
+from blaze.server.serialization import all_formats, most_formats, fastmsgpack
 
 
 accounts = DataFrame([['Alice', 100], ['Bob', 200]],
@@ -385,10 +385,11 @@ def test_server_accepts_non_nonzero_ables():
     Server(DataFrame())
 
 
-@pytest.mark.parametrize('serial', all_formats)
-def test_map_client_server(iris_server, serial):
-    print(serial)
-    test = iris_server
+def serialize_query_with_map_builtin_function(test, serial):
+    """
+    serialize a query that invokes the 'map' operation using a builtin function
+    return the result of the post operation along with expected result
+    """
     t = symbol('t', discover(iris))
     expr = t.species.map(len, 'int')
     query = {'expr': to_tree(expr)}
@@ -399,13 +400,31 @@ def test_map_client_server(iris_server, serial):
     respdata = serial.loads(response.data)
     result = serial.data_loads(respdata['data'])
 
-    istrue = result == compute(expr, {t: iris}, return_type=list)
-    try:
-        assert istrue
-    except:
-        # ofr fastmsgpack - pandas loads data into a Series - this is
-        # inconsistent with other serialization types that load a list
-        assert all(istrue)
+    exp_res = compute(expr, {t: iris}, return_type=list)
+    return (exp_res, result)
+
+
+@pytest.mark.parametrize('serial', most_formats)
+def test_map_client_server(iris_server, serial):
+    """
+    serialization for 'most_formats' returns a list so this is valid for
+    everything in most_formats
+    """
+    exp_res, result = serialize_query_with_map_builtin_function(iris_server,
+                                                                serial)
+
+    assert result == exp_res
+
+
+def test_map_client_server_fastmsgpack(iris_server):
+    """
+    serialization using fastmsgpack returns a Series object so our assertion
+    must be for all elements in Series
+    """
+    exp_res, result = serialize_query_with_map_builtin_function(iris_server,
+                                                                fastmsgpack)
+
+    assert all(result == exp_res)
 
 
 @pytest.mark.parametrize('serial', all_formats)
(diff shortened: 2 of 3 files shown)


blaze/utils.py
```diff
@@ -1,4 +1,9 @@
 from __future__ import absolute_import, division, print_function
+import types
+try:
+    import builtins
+except ImportError:
+    import __builtin__ as builtins
```
kwmsmith: You can use `from blaze.compatibility import builtins` here.

blaze/utils.py
```diff
@@ -187,6 +192,11 @@ def json_dumps(ds):
     return {'__!datashape': str(ds)}
 
 
+@dispatch(types.BuiltinFunctionType)
```
llllllllll: This dispatch is going to get hit for any function defined in C. This means that thinks like `np.sum` will get deserialized as `builtins.sum`. I think we might need some stronger checks here

blaze/server/tests/test_server.py
```diff
@@ -386,6 +386,44 @@ def test_server_accepts_non_nonzero_ables():
...
+                         data=serial.dumps(query),
+                         headers=mimetype(serial))
+    assert 'OK' in response.status
+    respdata = serial.loads(response.data)
+    result = serial.data_loads(respdata['data'])
+
+    istrue = result == compute(expr, {t: iris}, return_type=list)
+    try:
+        assert istrue
+    except:
```
llllllllll: could we catch only the error we are expecting?

blaze/utils.py
sandhujasmine: yep - this needs more work - just getting the conversation started 👍 

Thanks @llllllllll

llllllllll:
One issue I have with this idea is that users can create a string and then apply exec or eval to it and send that to the server to execute. I think sending functions might only be valid for the pickle backend where the server maintainers have decided security is not a concern.

blaze/server/tests/test_server.py
sandhujasmine: The question though: for fastmsgpack - pandas loads data into a Series - which is inconsistent with other serialization types that load a list.

`fastmsgpack_object_hook` is returning Series and DataFrames. Should it return lists to be consistent with other serialization objects. If it doesn't matter then I can break up the test so it does the correct assertion for different serializations.

It seems like all serialization should return consistent results, but I don't know what this effects.

llllllllll: The fastmsgpack was added to dramatically speed up the performance of the blaze server. The bottleneck in our production blaze servers and clients was actually serializing and deserializing the data. The general problem was the the dataframe was getting serialized as a list of records and then when rehydrating the data we would call object_hook tens of millions of times. object_hook was consistently in the top 5 ncalls when profiling, so even with a percall that rounded to 0s, it was a large part of the runtime. The fastmsgpack format is designed to send each block of the dataframe as a single buffer which is then compressed with blosc. This is why we have a seperate `data_(dumps|loads)` and `dumps|loads`. Basically, the data field of the response record is actually a seperately serialized dataframe which has been compressed. This allows us to send a smaller response which is then deserialized much more quickly.

There is also the potential to reduce the copying, If we deserialize as a list, then we need to take the bytes off the wire and turn them into python integers which is going to trigger thousands of allocations. This is really painful because we often just want to push that right back into an ndarray. I backported some code I added to pandas to prevent the copy and still give us a mutable buffer, which can be found here: https://github.com/blaze/blaze/blob/master/blaze/server/serialization.py#L117 (note, I actually just found a bug there, I will push a fix now). In the blaze version, we only slightly break the abstraction that bytes are immutable; however, the pandas version does this in a much more sane way: https://github.com/pydata/pandas/blob/master/pandas/io/packers.py#L325.

blaze/utils.py
sandhujasmine: More from the [docs](https://docs.python.org/2/library/types.html):
```
This module defines names for some object types that are used by the standard Python interpreter, but not for the types defined by various extension modules.
```

I now understand your comment better 👍 
But I believe this will be safe against functions compiled in extensions as per the docs and per some quick testing against `np.sum`

blaze/server/serialization.py
```diff
@@ -226,6 +226,11 @@ def fastmsgpack_materialize(data, dshape, odo_kwargs):
 )
 
 
+most_formats = frozenset(
+    g for _, g in globals().items() if isinstance(g, SerializationFormat) and
+    g is not fastmsgpack
+)
+
```
sandhujasmine: Added most_formats so I can test fastmsgpack separately from all other formats.

blaze/server/tests/test_server.py
```diff
@@ -385,6 +385,63 @@ def test_server_accepts_non_nonzero_ables():
...
+def test_map_client_server_fastmsgpack(iris_server):
+    """
+    serialization using fastmsgpack returns a Series object so our assertion
+    must be for all elements in Series
+    """
+    exp_res, result = serialize_query_with_map_builtin_function(iris_server,
+                                                                fastmsgpack)
+
+    assert all(result == exp_res)
+
```
sandhujasmine: > could we catch only the error we are expecting?

break apart 'fastmsgpack' test from `test_map_client_server()` into `test_map_client_server_fastmsgpack()`. The result of fastmsgpack is a Series object as opposed to a list so this makes the assertions cleaner.

blaze/utils.py
```diff
@@ -190,8 +190,7 @@ def json_dumps(ds):
 
 @dispatch(types.BuiltinFunctionType)
 def json_dumps(f):
-    if f.__module__ in ('__builtin__', 'builtins'):
```
sandhujasmine: This check is not required - accidently added it in - remove it for now.

blaze/utils.py
```diff
@@ -271,6 +282,26 @@ def register(typename, converter, converters=_converters):
...
+    else:
+        raise NotImplementedError("accepts numpy/pandas/builtin funcs only")
+
+    fcn = reduce(getattr, f.split('.')[1:], mod)
+    return fcn
+
+
+object_hook.register('numpy_pandas_function', numpy_pandas_function_from_str)
+
+
```
sandhujasmine: Added `json_dumps` and `object_hook` for pandas/numpy functions used in `map`. Added simple tests but need more interesting examples.

@kwmsmith

blaze/utils.py
```diff
@@ -193,6 +193,17 @@ def json_dumps(f):
...
+# conscturct a list of all numpy functions and their types
+
+@dispatch(Callable)
+def json_dumps(f):
+    if f.__module__.startswith('numpy') or f.__module__.startswith('pandas'):
+        fcn = ".".join([f.__module__, f.__name__])
+    else:
+        raise NotImplementedError("Only supports numpy or pandas functions")
+    return {'__!numpy_pandas_function': fcn}
+
```
sandhujasmine: Added `json_dumps` for pandas/numpy functions used in `map`. Only accepts pandas/numpy functions - the builtin functions dispatch on a separate function.

Is this a good way to ensure we only handle pandas/numpy functions? The builtin functions are handled by a different json_dumps that triggers for `types.BuiltinFunctionType`

@kwmsmith

blaze/utils.py
```diff
@@ -193,6 +193,17 @@ def json_dumps(f):
     return {'__!builtin_function': f.__name__}
 
 
+# conscturct a list of all numpy functions and their types
```
llllllllll: construct*

sandhujasmine: thanks - removed and replaced with a more relevant comment.

llllllllll:
As a general rule for the blaze server I would prefer if we assumed limited trust for any client that is connected. I think this is important because the blaze server box will have the credentials to access any of your data or have the data available directly. There is also the concern that the box is a shared resource and people will misuse the box because they do not know better. For example, a non-malicious denial of service could come from a data scientist using `np.savez` on a very large array. I think that this security concern is enough to warrant implementing this with a whitelist. This would allow server administrators to decide which functions are allowed or which are disallowed. I think that we could possibly have some sentinel value that says, "all are allowed" but I would strongly suggest not making this the default. I also don't think that we should manage a whitelist ourselves because then blaze is in the business of doing security audits.

sandhujasmine:
> One issue I have with this idea is that users can create a string and then apply exec or eval to it and send that to the server to execute. I think sending functions might only be valid for the pickle backend where the server maintainers have decided security is not a concern.

@llllllllll 
It's not implemented with an eval. It'll only look for the functions in the pandas, numpy or builtin namespace. But we might still need to limit this further as you suggest - will investigate further.

Could you please label this PR as Work In Progress?

llllllllll:
Providing access to the builtin namespace is providing `exec` and `eval`. For example:

```python
In [7]: bz.data(["print('hello from blaze!')"]).map(eval, '?string')
Out[7]: hello from blaze!

       
0  None

```

sandhujasmine:
> Providing access to the builtin namespace is providing exec and eval. For example:

I understand - thanks for the example. Will add a check for it.

blaze/utils.py
```diff
@@ -262,6 +278,27 @@ def register(typename, converter, converters=_converters):
...
+    reconstruct function from string representation
+    """
+    if f.startswith("numpy"):
+        mod = np
+
+    elif f.startswith("pandas"):
+        mod = pd
+
+    else:
+        raise NotImplementedError("accepts numpy/pandas/builtin funcs only")
```
sandhujasmine: Not sure if the server should raise an exception -- what is the best practice here? See resulting response object:

```
In [11]: expr = t.species.map(os.path.exists, 'str')
In [12]: response = test.post('/compute', data=serial.dumps(query), headers=mimetype(serial))
In [13]: response
Out[13]: <Response streamed [400 BAD REQUEST]>
```

@kwmsmith @llllllllll

sandhujasmine: How about a [flask NotImplemented exception (501 Not Implemented)](werkzeug.exceptions.NotImplemented)

blaze/utils.py
```diff
@@ -292,7 +300,7 @@ def numpy_pandas_function_from_str(f):
         mod = pd
 
     else:
-        raise NotImplementedError("accepts numpy/pandas/builtin funcs only")
+        raise wz_ex.NotImplemented("accepts numpy/pandas/builtin funcs only")
```
sandhujasmine: Raising http exceptions from `werkzeug.exceptions`, trying to be more specific; however, upon POST, the server will invoke [`check_request`](https://github.com/blaze/blaze/blob/master/blaze/server/server.py#L511) to see if the serialized data loads correctly. The `check_request` function expects [ValueError and traps it](https://github.com/blaze/blaze/blob/master/blaze/server/server.py#L215)

So, perhaps utils.py should only raise ValueError

@kwmsmith @llllllllll

kwmsmith:
Thanks for the input @llllllllll, it's good to have your usecase factored in as part of this work.

What about the following, to more clearly separate the locked-down server config from the trusted-client server config (all names are changeable):
* We create a separate `json_dumps_trusted` multipledispatch that is a superset of `json_dumps` with implementations that include whitelisted `builtins`, `numpy`, and `pandas` functions.
* We create a `object_hook_trusted` multipledispatch to go with `json_dumps_trusted`.
* We create `json_trusted` and `msgpack_trusted` `SerializationFormat` instances that use these `*_trusted` serializers / deserializers.
* We ensure that these `*_trusted` serialization formats are *not* included in the default `formats` argument to `Server`.
* The only way for a user to have `*_trusted` serialization formats is via explicitly instantiating `Server()` for now -- we don't provide commandline arguments to enable them.

@llllllllll overall, our target user is someone in an environment where using IPython notebook kernels with IPython notebook is acceptable, i.e., within a trusted environment, behind a firewall, etc.

llllllllll:
That seems to satisfy both of our use cases and protects users who don't know the difference. On a somewhat related note, which serialization formats have you been using for production?

kwmsmith:
> which serialization formats have you been using for production?

Currently, `json`, but we might use `msgpack` in the future.

llllllllll:
I would highly reccomend the `fast_msgpack` format for use with numpy and pandas, this was a really big performance improvement for us.

kwmsmith:
Closing -- this PR was absorbed into #1504.

PR #1499, opened 2016-05-04
Verify dataset is discoverable *before* adding it to server.

### LINKED ISSUE #1498: blaze server fall into inconsistency when handling adding some invalid file resource
When trying to add dataset to blaze server (via /add route), if the underlying file resource is given incorrectly, in some cases, the blaze server will return 422 error, but from now on fall into bad internal state (subsequent /datashape call returns 500 error). 

Those cases are:
1) invalid relative path, such as ./badfile; ../badfile;
2) invalid absolute path that ends with valid file ext, such as: 
   /Users/ywang/data/wrongfile.csv
   /Users/ywang/wrongdata/goodfile.csv

However, an when the invalid absolute path does not end with a valid file ext, such as:
/Users/ywang/data/wrongfile
/Users/ywang/wrong.data/goodfile.badext

Then the blaze server will simply return 422 error, and still in working state.

kwmsmith:
closes #1498.

### DIFF
--- a/blaze/server/server.py
+++ b/blaze/server/server.py
@@ -6,7 +6,6 @@ import errno
 import functools
 from hashlib import md5
 import os
-import re
 import socket
 from time import time
 from warnings import warn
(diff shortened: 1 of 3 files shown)


necaris:
@kwmsmith 👍 Looks good, tests pass on my checkout, I'm not able to get the server into an inconsistent state any more!

ywang007:
Thank you @kwmsmith . I tested a few cases and it all works fine. Putting discovery before adding makes it robust and gives consistent error. 👍

PR #1502, opened 2016-05-12
DOC: update compute docstring

llllllllll:
noticed that it listed `expr` as a `str` so I cleaned up some formatting


kwmsmith:
Thanks!

PR #1504, opened 2016-05-16
map/apply support in client/server

kwmsmith:
This picks up PR #1497 from @sandhujasmine:

* creates the `*_trusted` variants for `SerializationFormat` instances.
* Tests updated.
* `object_hook.py`, `json_dumps.py`, and their `_trusted` variants created as well.
* all `_trusted` dispatching is implemented as a superset of their underlying untrusted counterparts to keep it DRY.

### DIFF
--- a/blaze/json_dumps_trusted.py
+++ b/blaze/json_dumps_trusted.py
@@ -1,0 +1,32 @@
+from __future__ import absolute_import, division, print_function
+
+import types
+from collections import Callable
+
+from .dispatch import dispatch
+from functools import partial, wraps
+
+from .json_dumps import json_dumps
+
+
+json_dumps_trusted_ns = dict()
+dispatch = partial(dispatch, namespace=json_dumps_trusted_ns)
+
+@dispatch(types.BuiltinFunctionType)
+def json_dumps_trusted(f):
+    return {'__!builtin_function': f.__name__}
+
+@dispatch(Callable)
+def json_dumps_trusted(f):
+    # let the server serialize any callable - this is only used for testing
+    # at present - do the error handling when json comes from client so in
+    # object_hook, catch anything that is not pandas_numpy
+    fcn = ".".join([f.__module__, f.__name__])
+    return {'__!numpy_pandas_function': fcn}
+
+
+for types, func in json_dumps.funcs.items():
+    json_dumps_trusted.register(types)(func)
+
+print(json_dumps_trusted.funcs)
+
--- a/blaze/json_dumps_trusted.py
+++ b/blaze/json_dumps_trusted.py
@@ -12,9 +12,6 @@ from .json_dumps import json_dumps
 json_dumps_trusted_ns = dict()
 dispatch = partial(dispatch, namespace=json_dumps_trusted_ns)
 
-@dispatch(types.BuiltinFunctionType)
-def json_dumps_trusted(f):
-    return {'__!builtin_function': f.__name__}
 
 @dispatch(Callable)
 def json_dumps_trusted(f):
@@ -22,11 +19,8 @@ def json_dumps_trusted(f):
     # at present - do the error handling when json comes from client so in
     # object_hook, catch anything that is not pandas_numpy
     fcn = ".".join([f.__module__, f.__name__])
-    return {'__!numpy_pandas_function': fcn}
+    return {'__!callable': fcn}
 
 
 for types, func in json_dumps.funcs.items():
     json_dumps_trusted.register(types)(func)
-
-print(json_dumps_trusted.funcs)
-
--- a/blaze/object_hook.py
+++ b/blaze/object_hook.py
@@ -1,0 +1,139 @@
+from __future__ import absolute_import, division, print_function
+
+import datetime
+
+try:
+    from cytoolz import curry
+except ImportError:
+    from toolz import curry
+
+from datashape import dshape
+import numpy as np
+# these are used throughout blaze, don't remove them
+import pandas as pd
+import werkzeug.exceptions as wz_ex
+
+# Imports that replace older utils.
+from .compatibility import PY2, builtins, reduce
+
+# dict of converters. This is stored as a default arg to object hook for
+# performance because this function is really really slow when unpacking data.
+# This is a mutable default but it is not a bug!
+_converters = {}
+
+
+if PY2:
+    _keys = dict.keys
+else:
+    def _keys(d, _dkeys=dict.keys, _list=list):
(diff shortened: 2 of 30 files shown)


kwmsmith:
@llllllllll @sandhujasmine you mind taking a once-over?

llllllllll:
I can take a look a little later today

blaze/json_dumps_trusted.py
```diff
@@ -0,0 +1,32 @@
...
+import types
+from collections import Callable
+
+from .dispatch import dispatch
+from functools import partial, wraps
+
+from .json_dumps import json_dumps
+
+
+json_dumps_trusted_ns = dict()
```
llllllllll: instead of the loop below this might be more clear as `json_dumps_trusted_ns = jsun_dumps.funcs.copy()`

blaze/json_dumps_trusted.py
```diff
@@ -0,0 +1,32 @@
...
+dispatch = partial(dispatch, namespace=json_dumps_trusted_ns)
+
+@dispatch(types.BuiltinFunctionType)
```
llllllllll: why is this different from the callable case?

blaze/object_hook.py
```diff
@@ -0,0 +1,109 @@
...
+import datetime
+
+try:
+    from cytoolz import curry
+except ImportError:
+    from toolz import curry
+
+from datashape import dshape
+import numpy as np
+# these are used throughout blaze, don't remove them
```
llllllllll: this comment is for utils

blaze/json_dumps_trusted.py
```diff
@@ -0,0 +1,32 @@
...
+    # at present - do the error handling when json comes from client so in
+    # object_hook, catch anything that is not pandas_numpy
+    fcn = ".".join([f.__module__, f.__name__])
+    return {'__!numpy_pandas_function': fcn}
+
+
+for types, func in json_dumps.funcs.items():
+    json_dumps_trusted.register(types)(func)
+
+print(json_dumps_trusted.funcs)
```
llllllllll: dev cruft

blaze/object_hook_trusted.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def object_hook_trusted(ob, _converters=_converters_trusted):
+    return object_hook(ob, _converters=_converters)
+object_hook_trusted._converters = _converters_trusted
+object_hook_trusted.register = register(converters=_converters_trusted)
+del _converters_trusted
+
+
+@object_hook_trusted.register('builtin_function')
```
llllllllll: this will miss any C(ython) functions written in nump/pandas.

```
In [1]: type(np.array)
Out[1]: builtin_function_or_method
```

blaze/json_dumps_trusted.py
kwmsmith: Yes, I tried that initially, but the dispatching doesn't work in that case.

blaze/json_dumps_trusted.py
kwmsmith: It isn't, I removed this case and collapsed it into the `Callable` dispatch.

blaze/json_dumps_trusted.py
llllllllll: how so, iterating over the dict and updating each k/v pair should be functionally equivalent? At the end of this module would the `func` maps be different?

blaze/server/serialization.py
```diff
@@ -227,9 +241,24 @@ def fastmsgpack_materialize(data, dshape, odo_kwargs):
...
+    'fastmsgpack_trusted',
+    loads=partial(fastmsgpack_loads, object_hook=object_hook_trusted),
+    dumps=partial(fastmsgpack_dumps, default=json_dumps_trusted),
+    data_loads=fastmsgpack_data_loads,
+    data_dumps=fastmsgpack_data_dumps,
+    materialize=fastmsgpack_materialize,
 )
 
 
+most_formats = frozenset(g for _, g in globals().items()
```
llllllllll: can we give this a more descriptive name? what would the user want to use this set for?

blaze/object_hook_trusted.py
```diff
@@ -0,0 +1,38 @@
...
+def builtins_function_from_str(f):
+    if f in ("eval", "exec"):
```
llllllllll: if the user is trusted should we bother blocking these?

llllllllll:
Can we move the serialization code to a new package: `blaze.server.serialization`? I think this makes it easier to navigate to this code.

blaze/json_dumps_trusted.py
kwmsmith: Using `json_dumps.funcs.copy()` still retains the `json_dumps` function name, which ultimately triggers an exception in multipledispatch's inner dispatching logic; I have to do `json_dumps_trusted.register(types)(func)` to ensure the `json_dumps_trusted` function name is updated.

blaze/json_dumps_trusted.py
```diff
@@ -0,0 +1,26 @@
+from __future__ import absolute_import, division, print_function
+
+import types
+from collections import Callable
+
+from .dispatch import dispatch
+from functools import partial, wraps
+
```
sandhujasmine: wraps appears to be an used import

blaze/object_hook.py
```diff
@@ -9,7 +9,6 @@
 
 from datashape import dshape
 import numpy as np
-# these are used throughout blaze, don't remove them
 import pandas as pd
```
sandhujasmine: Unused imports below `builtins, reduce`

Question: So `object_hook` no longer supports map/apply - Is that right? Looks like the json for following objects correctly converts them back to python objects but these wouldn't be used by map/apply

```
In [2]: object_hook._converters
Out[2]: 
{'__!bytes': <function blaze.object_hook._read_bytes>,
 '__!datashape': <function datashape.util.dshape>,
 '__!datetime': pandas.tslib.Timestamp,
 '__!frozenset': frozenset,
 '__!mono': <function blaze.object_hook._read_mono>,
 '__!timedelta': <function blaze.object_hook._read_timedelta>}
```

kwmsmith:
> Can we move the serialization code to a new package: blaze.server.serialization? I think this makes it easier to navigate to this code.

Good idea -- done.

kwmsmith:
@llllllllll @sandhujasmine I think this one is ready to merge.

llllllllll:
lgtm. Thanks!

Issue #1508, opened 2016-05-25
ENH: MSSQL Compatibility


thequackdaddy:
Hello,

My company uses MSSQL, so that's the box I need to work with.

I'm interested in making blaze work on it for a project/idea of mine. But it seems some functionality is missing/broken. 

First, to get MSSQL to work, I noticed that some datatypes have to be mapped. I *think* the correct place to do this is in `odo`. I've filed a PR [here](https://github.com/blaze/odo/pull/430). 

So now I'm trying to understand the mechanics of blaze. I noticed that most things work, (sums, avg, split-apply-combine sum/avg). But I ran into an issue getting power (i.e., x ** 2) to work. It seems blaze is trying to use the function `pow`, but on MSSQL, the function is actually `Power` (case insensitive). 

Anyone have suggestions on how to fix? I'm not sure if this is a SQL Alchemy thing or something else. 

On another note, I don't have any idea how I can write a unit test that would work for this. I presume we'd have to use Appveyor or something else that supports MSSQL as travis-ci doesn't (AFAIK). Setting up that might be a little hard... But I'm willing to try...

PR #1514, opened 2016-05-28
ENH/WIP: MSSQL Compatibility

thequackdaddy:
Hello,

This enables blaze to correctly use `power`, `std`, and `var` math on MSSQL. There are probably other functions that need to be renamed, but this is a good start, I think. 

I don't have any unittests (not sure if its possible to use MSSQL tests on travis-ci). But if you could give me some ideas, I'll be happy to write them. To be clear, I'm still a little lost on exactly how all the internal mechanics work here, so I'm not sure what's the best approach.

Thanks for your consideration.

See #1508

### DIFF
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -23,6 +23,7 @@ import itertools
 from itertools import chain
 from operator import and_, eq, attrgetter
 import warnings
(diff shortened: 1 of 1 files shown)


postelrich:

Issue #1506, opened 2016-05-19
Methods in the `column.str` namespace


kwmsmith:
The  kwmsmith/blaze:str-namespace branch has started the work to implement Pandas' `df.column.str.upper()` API in Blaze.

This issue captures the relevant methods and breaks them out approximately according to difficulty.

The initial effort will implement these for Pandas first, and then extend to SQL given time, resources, and overlap with SQL support.

Completed:
* upper
* lower
* cat
* len
* find
* `__getitem__`

Low-hanging:
* isalnum isalpha isdecimal isdigit islower isnumeric isspace istitle isupper 
* reverse
* replace
* capitalize
* pad
* strip / lstrip / rstrip

regex-related (not low-hanging):
* slice_replace
* extract w/ single group

For reference, in both pandas and str:
* capitalize
* center
* count
* encode
* endswith
* find
* index
* isalnum
* isalpha
* isdecimal
* isdigit
* islower
* isnumeric
* isspace
* istitle
* isupper
* join
* ljust
* lstrip
* partition
* replace
* rfind
* rindex
* rjust
* rpartition
* rsplit
* rstrip
* split
* startswith
* strip
* swapcase
* title
* translate
* zfill

In pandas, not in str:
* contains
* decode
* extract
* extractall
* findall
* get
* get_dummies
* normalize
* pad
* repeat
* slice
* slice_replace
* wrap

stefanseefeld:
All but two of the above are part of https://github.com/blaze/blaze/pull/1515.
The two missing ones:
* reverse (should that be added to pandas first ?)
* extract (I need to  look into how to generate a new DataFrame from a single Series)

kwmsmith:
Closing, as the vast majority are now implemented.  Thanks @stefanseefeld

PR #1515, opened 2016-06-02
Str namespace

stefanseefeld:
This PR contains an accumulation of patches in support of a range of string functions / methods, corresponding to the work outlined in https://github.com/blaze/blaze/issues/1506.


blaze/compute/pandas.py
```diff
@@ -288,21 +293,59 @@ def compute_up(expr, data, **kwargs):
...
 @dispatch(UnaryStringFunction, Series)
 def compute_up(expr, data, **kwargs):
     name = type(expr).__name__
     return getattr(data.str, string_func_names.get(name, name))()
 
+@dispatch(Replace, Series)
+def compute_up(expr, data, **kwargs):
+    max = expr.max is None and -1 or expr.max
+    return data.str.replace(expr.old, expr.new, max)
+
```
kwmsmith: style nit: 2 lines between top-level functions.

blaze/expr/strings.py
```diff
@@ -83,6 +110,66 @@ class str_lower(UnaryStringFunction):
...
+
+@copydoc(StrFind)
+def str_find(col, sub):
+    if not isinstance(sub, basestring):
+        raise TypeError("'sub' argument must be a String")
+    return StrFind(col, sub)
+
+class Replace(ElemWise):
+    __slots__ = '_hash', '_child', 'old', 'new', 'max'
+    schema = datashape.string
```
kwmsmith: These `schema`s should be `datashape.Option(datashape.string)`, shouldn't they?

stefanseefeld: Isn't `replace` unconditionally returning a string ? I thought `datashape.Option()` was only required if the result may be `None`...

kwmsmith: If the input is `None`, then pandas' `replace` will pass it through, and the result will be `None`; that implies that if the input schema is `Option(string)`, the result schema should be, too.

blaze/compute/tests/test_pandas_compute.py
```diff
@@ -709,6 +709,106 @@ def test_str_lower():
     result = compute(expr, df).reset_index(drop=True)
     assert_series_equal(expected, result)
 
+@pytest.mark.parametrize('old,new,max,expected', (
+    ('o', 'x', None, 'Sxmexne'),
+    ('o', 'x', 1, 'Sxmeone')))
+def test_str_replace(old, new, max, expected):
+    df = pd.Series(['Someone'], name='name')
```
kwmsmith: Here and below:  could you expand the test data to include a `None` element, to test the logic on missing values?

So instead of `df = pd.Series(['Someone'], name='name')`, we'd have `df = pd.Series(['Someone', None], name='name')`.

blaze/expr/tests/test_strings.py
```diff
@@ -40,13 +42,43 @@ def test_like(ds):
...
+    assert bzs.str_isalpha(t.name).isidentical(t.name.str.isalpha())
+    assert bzs.str_isdecimal(t.name).isidentical(t.name.str.isdecimal())
+    assert bzs.str_isdigit(t.name).isidentical(t.name.str.isdigit())
+    assert bzs.str_islower(t.name).isidentical(t.name.str.islower())
+    assert bzs.str_isnumeric(t.name).isidentical(t.name.str.isnumeric())
+    assert bzs.str_isspace(t.name).isidentical(t.name.str.isspace())
+    assert bzs.str_istitle(t.name).isidentical(t.name.str.istitle())
+    assert bzs.str_isupper(t.name).isidentical(t.name.str.isupper())
+
+    assert bzs.Replace(t.name, 'A', 'a').isidentical(t.name.str.replace('A', 'a'))
```
kwmsmith: Reason for not providing `str_replace()`, `str_capitalize()`, `str_strip()`, etc. factory functions?  If the input validation is trivial, then we can just implement the class versions with lowercased names.

blaze/compute/tests/test_pandas_compute.py
stefanseefeld: I have done that, and the tests still pass, without any changes to the functions' `schema` member.

blaze/expr/strings.py
```diff
@@ -83,6 +113,86 @@ class str_lower(UnaryStringFunction):
...
+class str_istitle(UnaryStringFunction): schema = bool_
+class str_isupper(UnaryStringFunction): schema = bool_
+
+class StrFind(ElemWise):
+    """
+    Find literal substring in string column.
+
+    """
+
+    __slots__ = '_hash', '_child', 'sub'
```
llllllllll: In a recent change this has changed to `_arguments` and should not include the `_hash`. These are just the names of the arguments to the node. More info is here: https://github.com/blaze/blaze/blob/master/docs/source/expr-compute-dev.rst#adding-expressions

stefanseefeld:
I'v rebased this branch onto the new master, and made the required adjustments (`__slots__` -> `_arguments`), so the branch is working again.

blaze/expr/strings.py
```diff
@@ -83,6 +113,86 @@ class str_lower(UnaryStringFunction):
...
+
+def str_replace(col, old, new, max=None):
+    _validate(old, 'old', basestring, 'string')
+    _validate(new, 'new', basestring, 'string')
+    _validate_optional(max, 'max', int, 'integer')
+    return Replace(col, old, new, max)
+
+class Pad(ElemWise):
+    _arguments = '_child', 'width', 'side', 'fillchar'
+    schema = datashape.Option(datashape.string)
```
kwmsmith: I think this schema should be, instead:

```python
@property
def schema(self):
    self._child.schema
```

as in the `str_lower()` class above.

The reason is that `pad` passes through `None`s; if there are no `None`s in the input (i.e., non-option type), then `pad` will have no `None`s in the result either.

blaze/expr/strings.py
```diff
@@ -83,6 +113,86 @@ class str_lower(UnaryStringFunction):
...
+
+def str_pad(col, width, side=None, fillchar=None):
+    _validate(width, 'width', int, 'integer')
+    if side not in (None, 'left', 'right'):
+        raise TypeError('"side" argument must be either "left" or "right"')
+    _validate_optional(fillchar, 'fillchar', basestring, 'string')
+    return Pad(col, width, side, fillchar)
+
+class str_capitalize(UnaryStringFunction):
+    schema = datashape.Option(datashape.string)
```
kwmsmith: These schema should also be defined as properties and pass through `_child.schema`, I think.

blaze/expr/strings.py
stefanseefeld: OK, so to confirm my understanding: `Option(string)` would be used if the operation would possibly return `None` even if the argument is a `string`, while the above simply returns the argument type, i.e. maps `string` to `string` and `None` to `None` ?

blaze/expr/strings.py
stefanseefeld: OK, I think I have made all required adjustments.

blaze/compute/tests/test_pandas_compute.py
```diff
@@ -709,6 +710,90 @@ def test_str_lower():
...
+          ('string', 'pad', [10], ['someone'], ['   someone']),
+          ('string', 'pad', [10, 'left'], ['someone'], ['   someone']),
+          ('string', 'pad', [10, 'right'], ['someone'], ['someone   ']),
+          ('string', 'pad', [10, None, 'x'], ['someone'], ['xxxsomeone']),
+          ('string', 'slice_replace', [1, 3, 'x'], ['someone'], ['sxeone']),
+          ('string', 'slice_replace', [None, 3, 'x'], ['someone'], ['xeone']),
+          ('string', 'slice_replace', [1, None, 'x'], ['someone'], ['sx']),
+          ('string', 'slice_replace', [None, None, 'x'], ['someone'], ['x']),
+          ('string', 'slice_replace', [1, None, None], ['someone'], ['s']),
+          # TODO: slice
```
kwmsmith: Remove these TODO comments, as they require special casing...

kwmsmith:
LGTM, @llllllllll comments?

blaze/compute/pandas.py
```diff
@@ -288,21 +293,59 @@ def compute_up(expr, data, **kwargs):
...
 
 
 @dispatch(UnaryStringFunction, Series)
 def compute_up(expr, data, **kwargs):
     name = type(expr).__name__
     return getattr(data.str, string_func_names.get(name, name))()
 
+@dispatch(Replace, Series)
+def compute_up(expr, data, **kwargs):
+    max = expr.max is None and -1 or expr.max
```
llllllllll: what is the `-1 or expr.max` here?

blaze/expr/strings.py
```diff
@@ -83,6 +113,110 @@ class str_lower(UnaryStringFunction):
...
+
+class Replace(ElemWise):
+    _arguments = '_child', 'old', 'new', 'max'
+
+    @property
+    def schema(self):
+        return self._child.schema
+
+def str_replace(col, old, new, max=None):
+    _validate(old, 'old', basestring, 'string')
```
llllllllll: basestring doesn't exist in py3, is this compat?

blaze/compute/pandas.py
```diff
@@ -288,21 +293,59 @@ def compute_up(expr, data, **kwargs):
...
 
 
+@dispatch(StrFind, Series)
+def compute_up(expr, data, **kwargs):
+    return data.str.find(expr.sub)
+
+
+@dispatch(StrSlice, Series)
+def compute_up(expr, data, **kwargs):
+    if isinstance(expr.slice, tuple):
```
llllllllll: should we normalize to a slice object in the constructor?

blaze/compute/pandas.py
stefanseefeld: It translates between the blaze and pandas default values (blaze uses `None`, pandas `-1`).

blaze/expr/strings.py
```diff
@@ -165,13 +296,74 @@ def isstring(ds):
...
+        return str_slice(self.field, idx)
+
+    def slice_replace(self, start=None, stop=None, repl=None):
+        return str_slice_replace(self.field, start, stop, repl)
+
+class str(object):
+
+    __name__ = 'str'
+
+    def __get__(self, obj, type=None):
```
llllllllll: if `obj`  is none we should return `self`, also, `type` doesn't need a default argument.

blaze/expr/strings.py
stefanseefeld: I just cloned existing logic (and didn't observe any failures when testing with either Python 2 or 3).

blaze/compute/pandas.py
stefanseefeld: IIRC, blaze needs expressions to be hashable, which slices are not, so we use tuples instead.

llllllllll: oh weird, I thought they were immutable. Thanks for clarifying

blaze/expr/strings.py
kwmsmith: > if obj is none we should return self

+1

>  type doesn't need a default argument.

I thought that was part of the [descriptor protocol](https://docs.python.org/3.5/howto/descriptor.html#descriptor-protocol).

llllllllll: those docs are misleading: https://docs.python.org/3/reference/datamodel.html#object.__get__

kwmsmith:
@stefanseefeld Thanks for sticking with this PR, it's a big one!

LGTM, merging.

PR #1517, opened 2016-06-02
sort input validation

kwmsmith:


blaze/expr/collections.py
```diff
@@ -100,10 +100,32 @@ def sort(child, key=None, ascending=True):
     ascending : bool, optional
         Determines order of the sort
     """
+    if ascending not in (True, False):
```
llllllllll: why are we not just calling bool on the object?

blaze/expr/collections.py
```diff
@@ -100,10 +100,32 @@ def sort(child, key=None, ascending=True):
...
+        msg = "ascending must be True or False, given {}"
+        raise ValueError(msg.format(ascending))
     if not isrecord(child.dshape.measure):
-        key = None
+        if key is None or isinstance(key, _strtypes):
+            # Handle this case separately.
+            return Sort(child, None, ascending)
```
llllllllll: why is the key None here?

kwmsmith: When the datashape is array-like and `.sort()` is called with a string value for `key`, I've found no straightforward way to validate that the string is the name of the column being sorted.  We can pass in `key` here instead of `None`, but `Sort()` doesn't do any validation here either.

For example:

```python
In [5]: t
Out[5]: <`t` symbol; dshape='var * {name: string}'>

In [6]: t.name.dshape
Out[6]: dshape("var * string")  # not a record dshape, no `name` column.

In [7]: t.name.sort('name')
Out[7]: t.name.sort(None, ascending=True)

In [8]: t.name.sort()
Out[8]: t.name.sort(None, ascending=True)

In [9]: t.name.sort('foobar')
Out[9]: t.name.sort(None, ascending=True)
```

blaze/expr/collections.py
kwmsmith: To guard against calls like:

```
t.sort('a', 'b')
```

When the user should have done

```
t.sort(['a', 'b'])
```

blaze/expr/collections.py
llllllllll: if it is not a record should we just raise if we sort with a key, like, `TypeError('Cannot sort a non-record by a column name')`?

blaze/expr/collections.py
llllllllll: Ah, this is really helpful, I have tried to do that in the past. Can you just put a small comment in the source here that explains why we are doing this so I remember in the future?

blaze/expr/collections.py
kwmsmith: I thought of doing that and I see your case.  What made me go this route was the many spots in the testsuite where we sort like:

```
t.name.sort('name')
```

And if that now gives an exception like the above, I can imagine it would be surprising to users.  I'm undecided here.  Ideally we'd be able to extract the column name from `child` and compare to `key` and validate, but that seems like it would require a lot of work with expression dshapes, and would probably break other code that depends on `t.name` being non-record dshape.  Thoughts?

llllllllll: The code you showed is sort of ambigious with a dshape like: `var * {name: var * {name: A}}`. Do you know if there is any code in the wild that does this with `sort`?

Side note, it looks like `pd.Series.sort` just ignores the first argument (though it is named 'axis'.

kwmsmith: >  Do you know if there is any code in the wild that does this with sort?

I've not come across any, no.

> it looks like pd.Series.sort just ignores the first argument

It's raising `ValueError` for me, pandas version `0.18.0`.

I updated the docstring to describe the single-column semantics.

PR #1521, opened 2016-06-08
Improved server logging

necaris:
Takes #1492 and adds to it, defaulting logging to `sys.stdout` and adding documentation strings for the new parameters. 

As far as basic logging of output this is already done, however we will probably want to add more specific logging for different scenarios -- this can be in this PR or in a future update.

### DIFF
--- a/blaze/server/server.py
+++ b/blaze/server/server.py
@@ -21,7 +21,7 @@ import flask
 from flask import Blueprint, Flask, Response
 from flask.ext.cors import cross_origin
 from werkzeug.http import parse_options_header
-from toolz import valmap
+from toolz import valmap, compose
 
 import blaze
 from blaze import compute, resource
@@ -71,9 +71,11 @@ def _logging(func):
     @wraps(func)
     def _logger(*args, **kwargs):
         logger = flask.current_app.logger
-        logger.debug("Calling %s" % func.__name__)
-        ret = func(*args, **kwargs)
-        logger.debug("Leaving %s" % func.__name__)
+        try:
+            logger.debug("Calling %s" % func.__name__)
+            ret = func(*args, **kwargs)
+        finally:
+            logger.debug("Leaving %s" % func.__name__)
         return ret
     return _logger
 
@@ -96,6 +98,10 @@ def ensure_dir(path):
     except OSError as e:
         if e.errno != errno.EEXIST:
             raise
+
+# Default for logging exception tracebacks is to simply use the standard
+# `traceback.format_tb`
+_default_log_exception_formatter = compose(''.join, traceback.format_tb)
 
 
 def _register_api(app, options, first_registration=False):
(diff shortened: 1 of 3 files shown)


necaris:
@kwmsmith Without knowing more about what users want to track in the server this is pretty much good to go -- I just added a couple of debug logging calls and defaulting to `stdout` to your initial work.

kwmsmith:
@necaris great.

@llllllllll thoughts?

llllllllll:
I like it! ping @richafrank

blaze/server/server.py
```diff
@@ -62,6 +67,17 @@ class RC(object):
 _no_default = object()  # sentinel
 
 
+def _logging(func):
+    @wraps(func)
+    def _logger(*args, **kwargs):
+        logger = flask.current_app.logger
+        logger.debug("Calling %s" % func.__name__)
+        ret = func(*args, **kwargs)
+        logger.debug("Leaving %s" % func.__name__)
```
richafrank: Would it be useful to log "Leaving" in the exception case, i.e. should this be in a `finally`?

richafrank:
I was looking for any danger in returning exception tracebacks to endpoint consumers, but since those consumers would have passed `authorization` already, I think any such danger is mitigated.

blaze/server/server.py
```diff
@@ -562,10 +600,13 @@ def log_time(start=time()):
...
                                         odo_kwargs)
         except NotImplementedError as e:
-            return ("Computation not supported:\n%s" % e, RC.NOT_IMPLEMENTED)
+            error_msg = "Computation not supported:\n%s\n%s" % (e, traceback.format_exc())
+            app.logger.error(error_msg)
+            return (error_msg, RC.NOT_IMPLEMENTED)
         except Exception as e:
-            return ("Computation failed with message:\n%s: %s" % (type(e).__name__, e),
-                    RC.INTERNAL_SERVER_ERROR)
+            error_msg = "Computation failed with message:\n%s: %s\n%s" % (type(e).__name__, e, traceback.format_exc())
```
llllllllll: Could we make the `format_exc` function configurable? The use case that we have is that we run an extension of blaze server which may potentially have sensitive information in the traceback, we would like to filter the information the client sees before sending it over. I was thinking a function from traceback to str.

To pass the traceback to the function in a py2/3 compat way we could do:

```python
    tbmsg = serialize_traceback(sys.exc_info()[2])  # it is important not to store exc_info or the tb in a local
```

By default it could be `toolz.compose(''.join, traceback.format_tb)` which will do the same as it is currently doing.

Also, how do you feel about printing the traceback and then the  error. Right now the format will look like:

```
client tb
server  error message
server tb
```

and I think it may be nicer as:

```
client tb
server tb
server error message
```

necaris:
@kwmsmith @llllllllll Added requested configurability in https://github.com/blaze/blaze/pull/1521/commits/be85e1aea9551623b1af8d3293ff2017b6ad25da

kwmsmith:
@necaris thoughts on the travis failures?  They aren't obvious to me...

necaris:
@kwmsmith Thanks for pointing them out! Looking into them now, I'm doing some refactoring to make them pass and adding some tests too.

necaris:
@kwmsmith Apologies for the half-baked PR. It should be good to go now -- refactored the configurable exception logging and added tests for it.

Issue #1518, opened 2016-06-06
Sort API with `asc` and `desc`


kwmsmith:
This is an expanded description of the improvement to `sort`'s API mentioned in #1307.

Currently, `sort()` separates the columns to sort in the `key` argument, and only supports sorting ascending or descending for all columns specified via the `ascending` boolean argument.

The improvement here is to allow the user to specify ascending and descending sort on a per-column basis.

The API will look something like:

```python
expr.sort(bz.asc(expr.a), bz.desc(expr.b))
```

Equivalent variants:

```python
expr.sort(bz.asc('a'), bz.desc('b')) # strings supported
expr.sort(bz.asc(expr.a), bz.desc('b')) # mixed mode supported
expr.sort('a', bz.desc('b')) # if not specified, defaults to ascending.
```

Tasks:
* [ ] implement new `asc` and `desc` expressions.
* [ ] implement validation logic to ensure that  `asc` and `desc` are only used in a sorting context.
* [ ] re-implement `sort()` to accept `asc` and `desc` expression objects.
* [ ] re-implement `sort()` on pandas
* [ ] re-implement `sort()` on SQL.
* [ ] tests...

llllllllll:
since it is a free function I take it that it will also be available as a method of `Field` like: `expr.col.asc()`?

kwmsmith:
> since it is a free function I take it that it will also be available as a method of Field like: expr.col.asc()?

That makes sense from a consistency point-of-view...

kwmsmith:
@llllllllll BTW, I'm teeing-up this issue for a new developer to the project, who will be starting on things tomorrow.

llllllllll:
Wonderful! I look forward to working with them!

skrah:
https://github.com/skrah/blaze/commit/1697e015182188301bc04275b8df638ad4bb988e is the first attempt.

Not working yet:

  - Usub etc. expressions.

  - Generated sql that involves tail() is different from the one that sort() generates (but I think it is correct).

llllllllll:
`Tail` for sql is a funny thing, we actually rewrite it into a different blaze expression with `optimize`:

```python
@dispatch(Tail)
def _subexpr_optimize(expr):
    child = sorter = expr._child
    while not isinstance(sorter, Sort):
        try:
            sorter = sorter._child
        except AttributeError:
            break
    else:
        # Invert the sort order, then take the head, then re-sort based on
        # the original key.
        return child._subs({
[...]

PR #1531, opened 2016-06-20
Memoize subs.

skrah:
This patch improves the performance of `subs()` for large or more complex trees (see  #1518).

Simple example (10x speedup):

``` python
from blaze.expr.core import subs
import time

def mktree(d):
    if d <= 0:
        return []
    return [1, 2, [mktree(d-1)], [mktree(d-1)]]

t1 = mktree(16)

start = time.time()
t2 = subs(t1, {1: 2})
end = time.time()
print(end-start)
```

### DIFF
--- a/blaze/expr/core.py
+++ b/blaze/expr/core.py
@@ -15,7 +15,7 @@ from toolz import unique, concat, first
 
 from ..compatibility import _strtypes
 from ..dispatch import dispatch
-from ..utils import ordered_intersect
+from ..utils import ordered_intersect, memoize_mutable
 
 __all__ = ['Node', 'path', 'common_subexpression', 'eval_str']
 
@@ -404,6 +404,7 @@ def subterms(x):
     yield x
 
 
+@memoize_mutable
 def subs(o, d):
     """ Substitute values within data structure
 
--- a/blaze/expr/core.py
+++ b/blaze/expr/core.py
@@ -15,7 +15,7 @@ from toolz import unique, concat, first
 
 from ..compatibility import _strtypes
 from ..dispatch import dispatch
-from ..utils import ordered_intersect, memoize_mutable
+from ..utils import ordered_intersect, memoize_obj_mapping
 
 __all__ = ['Node', 'path', 'common_subexpression', 'eval_str']
 
@@ -404,7 +404,7 @@ def subterms(x):
     yield x
 
 
-@memoize_mutable
+@memoize_obj_mapping
 def subs(o, d):
     """ Substitute values within data structure
 
--- a/blaze/utils.py
+++ b/blaze/utils.py
@@ -274,18 +274,25 @@ def parameter_space(*args):
     )))
 
 
-def memoize_mutable(fn):
+def memoize_obj_mapping(fn):
+    """Memoize fn with the argument types (object, Mapping). The arguments
+       may be mutable.
+    """
     memo = {}
     @functools.wraps(fn)
-    def inner(*args):
+    def inner(o, d):
+        key = (o, tuple(d.items()))
         try:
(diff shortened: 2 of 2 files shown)


blaze/expr/core.py
```diff
@@ -404,6 +404,7 @@ def subterms(x):
     yield x
 
 
+@memoize_mutable
 def subs(o, d):
```
kwmsmith: @llllllllll to your knowledge in Blaze, is `subs()` ever used for anything besides expressions?  IOW, even though `subs` is fairly general and can handle mutable nested data structures, in actuality is it ever used for (mutable) data structures that aren't expressions?


I ask because having mutable memoization starts adding complexity that we may not need if we constrain `subs` to immutable things.

blaze/utils.py
```diff
@@ -266,3 +272,20 @@ def parameter_space(*args):
     return tuple(product(*(
         arg if isinstance(arg, tuple) else (arg,) for arg in args
     )))
+
+
+def memoize_mutable(fn):
```
kwmsmith: @skrah can you test this implementation on an expression tree that uses a `blaze.Data` object that's referring to a large  array or dataframe, and check and see if the entire contents of the array / dataframe are serialized?  I want to make sure that we don't serialize or memoize GBs of data with this implementation.

blaze/expr/core.py
llllllllll: I am pretty sure there is a dispatch for `Mono` to swap out types in a datashape, mainly type or dim vars. Is subs that expensive that we want to add this kind of memoization? The serialization/deserialization seem much more expensive than a tree walk.

blaze/utils.py
llllllllll: Why not make the function returrn a function to defer construction. pickle is not designed for speed / data efficiency, it is just supposed to work. imagine something like:

```python
@dec
def f():
    a = 1
    b = 2
    return lambda: {'a': a, 'b': b}
```

`dec` could call the nullary function every time the value is requested to rebuild a new value.

I am not sure I love the name memoize mutable since the function returns a new object and I normally think of memoize as `f(x) is f(x)`

blaze/expr/core.py
llllllllll: Sorry, I saw the OP and the other pr, seems like this is actually a big win. I still wonder if we can speed it up further by avoiding pickle.

skrah:
@kwmsmith Yes, in the first implementation large objects were serialized.  I've pushed a new custom decorator that avoids serializing for the common case.

The performance is good: It speeds up the ```tail()``` example from #1518 for the existing ```sort``` by about 80% and for the new ```sort2``` by a huge amount.

The overhead for trivial cases should be minimal since the hash is mostly cached.

Statistics for running the test suite for the paths taken in the decorator:

    * hashable: 25919
    * pickled: 49
    * unpicklable: 0
    * pickle maxlen: 1048718

So large mutable objects are still pickled, however there are very few of them. One additional thing to note is that pickling performance is vastly better in Python 3, even compared to cPickle in Python 2.

skrah:
The test failure in r3192686 is due to #1541.

skrah:
As to the specific question about numpy arrays. *If* they are wrapped in ```data```, they count as hashable and the new ```subs``` is faster than the old in this example:

```python
import numpy as np
import blaze as bz
import time

x = np.zeros(10000000)
d = bz.data(x)

start = time.time()
for i in range(100000):
    bz.expr.core.subs(d, {d: "XXX"})
end = time.time()
print(end-start)
```

Existing subs: 0.71s

New subs: 0.46s

skrah:
Sorry, in the example ```d = bz.data(x)``` should be in the loop. The new ```subs``` is still a bit faster.

postelrich:

PR #1533, opened 2016-06-20
Add datetime functions to 'dt' namespace.

stefanseefeld:
This PR adds various functions to the 'dt' namespace.
To mimic pandas, there are actually two distinct namespaces, one for `datetime` types, the other for `timedelta` types, both named `dt`.

### DIFF
--- a/blaze/expr/datetime.py
+++ b/blaze/expr/datetime.py
@@ -30,7 +30,8 @@ class DateTime(ElemWise):
         return '%s.%s' % (str(self._child), type(self).__name__.lower())
 
     def _schema(self):
-        return dshape(self._dtype)
+        ds = dshape(self._dtype)
+        return ds if not isinstance(self._child.schema.measure, datashape.Option) else datashape.Option(ds)
 
     @property
     def _name(self):
--- a/blaze/expr/datetime.py
+++ b/blaze/expr/datetime.py
@@ -4,6 +4,8 @@ from .expressions import ElemWise, schema_method_list, method_properties
 
 import datashape
 from datashape import dshape, isdatelike, isnumeric
+from datashape.coretypes import timedelta_
+from ..compatibility import basestring
 
 
 __all__ = ['DateTime', 'Date', 'date', 'Year', 'year', 'Month', 'month', 'Day',
@@ -13,6 +15,11 @@ __all__ = ['DateTime', 'Date', 'date', 'Year', 'year', 'Month', 'month', 'Day',
            'date', 'Time', 'time', 'week', 'nanoseconds', 'seconds', 'total_seconds',
            'UTCFromTimestamp', 'DateTimeTruncate',
            'Ceil', 'Floor', 'Round', 'strftime']
+
+
+def _validate(var, name, type, typename):
+    if not isinstance(var, type):
+        raise TypeError('"%s" argument must be a %s'%(name, typename))
 
 
 class DateTime(ElemWise):
@@ -332,14 +339,18 @@ class dt_ns(object):
     def days_in_month(self):
         return days_in_month(self.field)
     def strftime(self, format):
+        _validate(format, 'format', basestring, 'string')
         return strftime(self.field, format)
     def truncate(self, *args, **kwargs):
         return truncate(self.field, *args, **kwargs)
     def round(self, freq):
+        _validate(freq, 'freq', basestring, 'string')
         return Round(self.field, freq)
     def floor(self, freq):
+        _validate(freq, 'freq', basestring, 'string')
         return Floor(self.field, freq)
     def ceil(self, freq):
+        _validate(freq, 'freq', basestring, 'string')
         return Ceil(self.field, freq)
     def week(self):
         return week(self.field)
@@ -358,7 +369,7 @@ class seconds(DateTime): _dtype = datashape.int64
 class total_seconds(DateTime): _dtype = datashape.int64
 
 
-class td_ns(object):
+class timedelta_ns(object):
 
     def __init__(self, field):
         self.field = field
@@ -369,18 +380,17 @@ class td_ns(object):
     def total_seconds(self): return total_seconds(self.field)
 
 
-class td(object):
+class timedelta(object):
 
     # pandas uses the same 'dt' name for
     # DateTimeProperties and TimedeltaProperties.
     __name__ = 'dt'
 
     def __get__(self, obj, type=None):
-        return td_ns(obj) if obj is not None else self
+        return timedelta_ns(obj) if obj is not None else self
 
 
 def isdeltalike(ds):
-    from datashape.coretypes import timedelta_
     return ds == timedelta_
 
 schema_method_list.extend([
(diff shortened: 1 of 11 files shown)


stefanseefeld:
Here are some missing functions, for which I either didn't find any docs (in pandas), or found the docs to be not very clear.

* `to_pydatetime()`
  Does this take a string and parses it into a datetime instance ?
* `to_pytimedelta()`
  Likewise, what is the input, what the output type ?
* `normalize()`
  Not sure what this is doing.
* `tz`
  This may return `None` (if no timezone info is present in the input string), and I'm not sure how to handle that.

kwmsmith:
> `to_pydatetime()`

I imagine that this takes the underlying numpy datetime type and converts it to the equivalent python `datetime.datetime` instance, but I'd have to check.  Same for `to_pytimedelta()`.

> normalize() Not sure what this is doing.

Agreed.  Options are to:
* leave it out of Blaze for now
* look at its implementation and divine its purpose from that, and make a Pandas documentation PR :)


For `tz`, it is an attribute I take it, correct?  Currently Blaze allows you to do things like

```python
t.col.hour
t.col.minute
# etc ...
```

For datetime columns -- could we do something similar for `.tz`?  And its datashape would be an option type to handle the missing timezones.  Am I misunderstanding something?

stefanseefeld:
In contrast to `t.col.hour`, which will always return a value, `t.col.tz` may return `None`, and I'm not sure how to convert that into a valid item.

blaze/expr/datetime.py
```diff
@@ -222,10 +246,149 @@ def truncate(expr, *args, **kwargs):
...
+        return dt_ns(obj) if obj is not None else self
+
+
+class days(DateTime): _dtype = datashape.int64
+class nanoseconds(DateTime): _dtype = datashape.int64
+class seconds(DateTime): _dtype = datashape.int64
+class total_seconds(DateTime): _dtype = datashape.int64
+
+
+class td_ns(object):
```
kwmsmith: Can we rename this to something like `timedelta_ns`?  Using `td` as the abbreviation isn't obvious at first glance.

blaze/expr/datetime.py
```diff
@@ -222,10 +246,149 @@ def truncate(expr, *args, **kwargs):
...
+    def __init__(self, field):
+        self.field = field
+
+    def days(self): return days(self.field)
+    def nanoseconds(self): return nanoseconds(self.field)
+    def seconds(self): return seconds(self.field)
+    def total_seconds(self): return total_seconds(self.field)
+
+
+class td(object):
```
kwmsmith: Same as above -- can we call this class `timedelta` or something similar?

blaze/expr/datetime.py
```diff
@@ -222,10 +246,149 @@ def truncate(expr, *args, **kwargs):
...
+    # pandas uses the same 'dt' name for
+    # DateTimeProperties and TimedeltaProperties.
+    __name__ = 'dt'
+
+    def __get__(self, obj, type=None):
+        return td_ns(obj) if obj is not None else self
+
+
+def isdeltalike(ds):
+    from datashape.coretypes import timedelta_
```
kwmsmith: Move `import` to top of module.

blaze/expr/datetime.py
stefanseefeld: sure.

blaze/expr/tests/test_datetime.py
```diff
@@ -1,5 +1,6 @@
 from blaze.expr import symbol
-from blaze.expr.datetime import isdatelike
+import blaze.expr.datetime as bdt
+#from blaze.expr.datetime import isdatelike
```
kwmsmith: Dev cruft.

blaze/expr/datetime.py
```diff
@@ -222,10 +246,149 @@ def truncate(expr, *args, **kwargs):
...
+        return is_quarter_start(self.field)
+    def is_quarter_end(self):
+        return is_quarter_end(self.field)
+    def is_year_start(self):
+        return is_year_start(self.field)
+    def is_year_end(self):
+        return is_year_end(self.field)
+    def days_in_month(self):
+        return days_in_month(self.field)
+    def strftime(self, format):
```
kwmsmith: For the methods that take a `format` string argument, when is validation done?  At a minimum we want to raise a `TypeError` as early as possible if `format` isn't a string; ideally we'd validate that `format`'s value is correct, if possible.

stefanseefeld: I agree on the early validation. But I'm not sure what constitutes a "correct" value, and thus I'm not sure we can do more than make sure it's a string.

kwmsmith: It's likely not feasible to validate the value of `format` for `strftime()`, but even just checking its type beforehand is useful -- the more we can verify at expression-creation time the better.

blaze/expr/datetime.py
```diff
@@ -119,6 +129,27 @@ def utcfromtimestamp(expr):
     return UTCFromTimestamp(expr)
 
 
+class nanosecond(DateTime): _dtype = datashape.int64
```
kwmsmith: This is probably getting annoying by now, but we need to think about Option types and datetimes, so I think we need to do something like:

```python
class week(DateTime):
    @property
    def _dtype(self):
        return datashape.int64 if self._child._dtype == datashape.datetime_ else Option(datashape.datetime_)
```

And similar for the `string` and `bool_` `_dtype`s here.

stefanseefeld: True. I had simply cloned from the existing classes above. I assume I should fix all of those as well while I'm at it ?

kwmsmith: Yes, looks like they all need to be consistent.

stefanseefeld: I'm testing what I thought to be the obvious solution: https://github.com/stefanseefeld/blaze/commit/99c1bc637cda7a05c523d75e0f6c4943d450df79, which results in test failures in blaze/server/tests/test_server.py::test_minute_query (https://travis-ci.org/stefanseefeld/blaze/jobs/143455858). Trying to debug this now...

stefanseefeld: It turns out the mismatch is caused by https://github.com/blaze/odo/blob/master/odo/backends/pandas.py#L25, were `datetime_` is detected as "possibly_missing". What is the rationale for this ?
Should I adjust the failing test (by marking the expected schema as '?int64'), or does this require fixing in odo ?
(It would be good to have at least a bit of inline comments or even documentation in odo's `dshape_from_pandas` to help people understand the intent...)

stefanseefeld: OK, the last commit addresses the above and adjusts a test to use a changed expectation.

stefanseefeld:
I noticed an issue with the documentation build, but am not sure whether that's caused by a misconfiguration or something else: None of the cross references I checked (e.g. `:func:...`) are compiled into links. I didn't see anything wrong, so perhaps that's a problem with my local sphinx install.

kwmsmith:
> I noticed an issue with the documentation build, but am not sure whether that's caused by a misconfiguration or something else: None of the cross references I checked (e.g. :func:...) are compiled into links. I didn't see anything wrong, so perhaps that's a problem with my local sphinx install.

Yes, am working to carve out time for a documentation overhaul over the next quarter or so.

Issue #1532, opened 2016-06-20
vectorized string function naming scheme


kwmsmith:
I like the `t.col.str.upper()` API (and likewise for datetime, with `t.col.dt.zzz()`).

I'm not wild about our current string function naming scheme:
* `str_find`
* `str_len`
* `str_upper`
* etc.

I'd much rather we recommend something like:

```python
from blaze import string as bstr

bstr.len(t.col) == t.col.str.len()
bstr.upper(t.col) == t.col.upper()
# etc...
```

That way the function-level API is much more streamlined, and we let the user determine how they are to be namespaced.

@llllllllll @stefanseefeld Thoughts?

stefanseefeld:
I fully agree. I find the 'str_' prefix quite redundant, given that these symbols already come from the 'strings' module.

llllllllll:
Is this just for the functional form of the expressions? I think the `expr.str.*` is nice for methods but I would support this for the functional versions.

kwmsmith:
> Is this just for the functional form of the expressions?

Yes, exactly -- the method form would remain `expr.str.zzz()`, which would be equivalent to `bstr.zzz(expr)`.

stefanseefeld:
Do we need to support the existing API (and deprecate it) ?

stefanseefeld:
This is being addressed by https://github.com/blaze/blaze/pull/1538.

Note that there are two distinct types of redundancy:

* The naming of functions in the blaze.expr.strings module.
* The appearance of certain methods both within the `str` namespace as well as outside it. (Example: the `s.name.str.len()` method can also be spelled `s.name.str_len()`. The latter is now implicitly deprecated by virtue of the stand-alone `str_len()` function being deprecated.

kwmsmith:
Closed by #1538.

PR #1538, opened 2016-06-30
Clean up redundant naming.

stefanseefeld:
This addresses https://github.com/blaze/blaze/issues/1532.

Note that there are two distinct sets of redundancies, both of which are addressed:

* functions within `blaze.strings` using a 'str_' prefix
* string functions appearing both within the `str` attribute as well as outside it (e.g. `s.name.str.len()` vs. `s.name.str_len()`).

Both have been addressed with this PR, with the old names being deprecated.
(There appears to be a bug in `py.test` that triggered the error "Recursion detected..." when the old functions were used, so I had to update all tests to use the new functions instead. See https://github.com/pytest-dev/pytest/issues/70. I couldn't reproduce any of those recursions when testing the old / new functions manually.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,6 @@ install:
 
   - if [[ $TRAVIS_PYTHON_VERSION == '2.7' || $TRAVIS_PYTHON_VERSION == '3.4' ]]; then conda install spark=$SPARK_VERSION -c blaze -c anaconda-cluster; fi
   - if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then conda install pyhive -c blaze; fi
-  - if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then pip install future; fi
 
   # blaze required deps
   - pip install git+https://github.com/blaze/datashape
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,7 @@ install:
 
   - if [[ $TRAVIS_PYTHON_VERSION == '2.7' || $TRAVIS_PYTHON_VERSION == '3.4' ]]; then conda install spark=$SPARK_VERSION -c blaze -c anaconda-cluster; fi
   - if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then conda install pyhive -c blaze; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then pip install future; fi
 
   # blaze required deps
   - pip install git+https://github.com/blaze/datashape
--- a/blaze/compute/pandas.py
+++ b/blaze/compute/pandas.py
@@ -290,16 +290,10 @@ def compute_up(expr, data, **kwargs):
     return compute_up(expr._child.distinct().count(), data, **kwargs)
 
 
-# No mapping needed - the original names work just fine.
-string_func_names = {
-    # <blaze function name>: <pandas function name>
-}
-
-
 @dispatch(UnaryStringFunction, Series)
 def compute_up(expr, data, **kwargs):
     name = type(expr).__name__
-    return getattr(data.str, string_func_names.get(name, name))()
+    return getattr(data.str, name)()
 
 @dispatch(Replace, Series)
 def compute_up(expr, data, **kwargs):
--- a/blaze/compute/pandas.py
+++ b/blaze/compute/pandas.py
@@ -290,23 +290,10 @@ def compute_up(expr, data, **kwargs):
     return compute_up(expr._child.distinct().count(), data, **kwargs)
 
 
-string_func_names = {'str_len': 'len',
-                     'strlen': 'len',
-                     'str_upper': 'upper',
-                     'str_lower': 'lower',
-                     'str_capitalize': 'capitalize',
-                     'str_strip': 'strip',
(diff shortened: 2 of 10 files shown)


stefanseefeld:
This currently fails on Python 2.7 in the `blaze/compute/tests/test_sparksql.py` tests. Unfortunately I'm unable to run those as I'm unable to install the required packages on my Fedora 23 laptop: `pyhive` needs an external `libsasl2.so.2` library that is not available for my platform.

stefanseefeld:
Problem fixed.

stefanseefeld:
The recently introduced functions using the `str_` prefix have simply been renamed. The ones already present in the previous release have been formally deprecated.

blaze/compute/pandas.py
```diff
@@ -290,23 +290,10 @@ def compute_up(expr, data, **kwargs):
...
-                     'str_isalnum': 'isalnum',
-                     'str_isalpha': 'isalpha',
-                     'str_isdecimal': 'isdecimal',
-                     'str_isdigit': 'isdigit',
-                     'str_islower': 'islower',
-                     'str_isnumeric': 'isnumeric',
-                     'str_isspace': 'isspace',
-                     'str_istitle': 'istitle',
-                     'str_isupper': 'isupper'}
+# No mapping needed - the original names work just fine.
```
kwmsmith: Is there any reason to keep this now-empty mapping around anymore?  My vote is to remove it if it isn't necessary.

blaze/compute/sql.py
```diff
@@ -1198,9 +1197,10 @@ def compute_up(t, s, **kwargs):
     return s.like(t.pattern.replace('*', '%').replace('?', '_'))
 
 
-string_func_names = {# <blaze function name>: <SQL function name>
-                     'str_upper': 'upper',
-                     'str_lower': 'lower'}
+# No mapping needed - the original names work just fine.
```
kwmsmith: Same comment as above...

blaze/expr/strings.py
```diff
@@ -87,41 +90,32 @@ class UnaryStringFunction(ElemWise):
...
-class str_isalnum(UnaryStringFunction): schema = bool_
-class str_isalpha(UnaryStringFunction): schema = bool_
-class str_isdecimal(UnaryStringFunction): schema = bool_
-class str_isdigit(UnaryStringFunction): schema = bool_
-class str_islower(UnaryStringFunction): schema = bool_
-class str_isnumeric(UnaryStringFunction): schema = bool_
-class str_isspace(UnaryStringFunction): schema = bool_
-class str_istitle(UnaryStringFunction): schema = bool_
-class str_isupper(UnaryStringFunction): schema = bool_
+class isalnum(UnaryStringFunction): schema = bool_
```
kwmsmith: Are these `schema` correct?  Shouldn't they be `@property` like the `class lower` and `class upper`, since `isalpha` on an `Option(string)` column should return an `Option(bool_)` dshape, correct?

blaze/compute/pandas.py
stefanseefeld: I was wondering that, too. We can reintroduce it if we have to...
I'll remove it.

.travis.yml
```diff
@@ -45,6 +45,7 @@ install:
 
   - if [[ $TRAVIS_PYTHON_VERSION == '2.7' || $TRAVIS_PYTHON_VERSION == '3.4' ]]; then conda install spark=$SPARK_VERSION -c blaze -c anaconda-cluster; fi
   - if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then conda install pyhive -c blaze; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then pip install future; fi
```
kwmsmith: Where is the `future` package used in the code below?  I'm not able to find it.

stefanseefeld: The `builtins` package (used in the `slice` function) is available in Python 2 only  if 'future' is installed.

llllllllll: in py2 slice is in `__builtin__`, instead of adding a new package can we just put slice in compat?

blaze/expr/strings.py
```diff
@@ -87,41 +90,32 @@ class UnaryStringFunction(ElemWise):
...
-class str_isalnum(UnaryStringFunction): schema = bool_
-class str_isalpha(UnaryStringFunction): schema = bool_
-class str_isdecimal(UnaryStringFunction): schema = bool_
-class str_isdigit(UnaryStringFunction): schema = bool_
-class str_islower(UnaryStringFunction): schema = bool_
-class str_isnumeric(UnaryStringFunction): schema = bool_
-class str_isspace(UnaryStringFunction): schema = bool_
-class str_istitle(UnaryStringFunction): schema = bool_
-class str_isupper(UnaryStringFunction): schema = bool_
+class isalnum(UnaryStringFunction): schema = Option(bool_)
```
kwmsmith: Shouldn't these schema be something like:

```python
@property
def schema(self):
    return datashape.bool_ if self._child.schema == datashape.string else datashape.Option(datashape.bool_)
```

?

kwmsmith:
@stefanseefeld looks good.  I'll merge and add a `whatsnew` entry.

PR #1540, opened 2016-07-05
Issue #1518: Add a new sort API.

skrah:
I'm not sure how to attach a pull request to an issue, so this supersedes #1518.

The name `sort2` isn't very nice, but I think there is no good way of adding the new functionality to the existing `sort` function.


kwmsmith:
@skrah Apologies for taking so long to get to this PR.

I see that you allow the the `asc` and `desc` expressions to be `compute`d directly and act, essentially, an alias to `sort(asc(col))` and `sort(desc(col))`.

API-wise, I was thinking more that `asc` and `desc` are meant to be auxiliary expressions that flag the column's sort order, and don't do anything when `compute`d outside the context of a containing `sort`.

One way to handle this would be to absorb the `asc` and `desc` exprs into the `sort2` object's attributes itself, so that by the time `compute` is called, no `asc` or `desc` objects remain.  With that API, `asc` and `desc` are meant to make the `sort` API easier to use, but do not show up at `compute` time at all.  Any attempt to `compute` `asc` or `desc` would be an error.

That way we don't have to implement `compute_up` for `Key` expressions at all.

Looking over the rest of the PR currently.

skrah:
@kwmsmith I think I took the idea of (optionally) computing ```asc()``` directly from https://github.com/blaze/blaze/issues/1518#issuecomment-224066762 .

I got used to it pretty quickly (i.e. I no longer wanted to type ```sort(asc('a'))``` when I just wanted to sort a column), but it *does* of course add complexity to the implementation.

kwmsmith:
I see, and I see how you could come to that conclusion from the comment without more context :)

Unless I misunderstand @llllllllll, what I understood there was just that we would support two ways of creating an `asc` expression object: via the function call interface `asc(t.col)` or the method call interface `t.col.asc()`.  It was not meant to imply that the `asc` expression was itself a computable thing.

The underlying inspiration for the interface comes from SQL dialects, like PostGreSQL, and the desire to keep the `asc` and `desc` attributes close to the columns they're describing, rather than in a separate boolean array, as in Pandas (currently).

To keep things simpler, please remove the `compute_up` implementations for `Key` expressions.

llllllllll:
I agree that `Key` should not be computable, we should probably have `compute_up` raise an exception that says "You probably needed a `sort` expr"

skrah:
Okay, I have one more question here: Is ```asc(t.col)``` allowed to created a ```Sort``` expression under the hood?  I somehow got the impression that the names of the factory functions always must match the names of the expression objects.

llllllllll:
No, I think `asc(expr)` should be the same as `expr.asc()`, it is just nice for things like `map`

kwmsmith:
> No, I think asc(expr) should be the same as expr.asc(), it is just nice for things like map
 
Agreed.

llllllllll:
The primary reason for supporting both apis is to allow people to write code in whatever way makes sense for them. The methods feel very natural for people coming from pandas/numpy; however, for many pure functional operations (which blaze promotes) a free functions is more natural. This is also tied to the implementation where we will write the method as a free function and then dynamically register it to types using the dshape or schema maps.

skrah:
I've taken out the ```Key``` computations.  I'm not sure what to do about ```Key``` inheriting from ```Expr```.   If it can't be computed, should it be an expression at all? Currently it still has a dshape.

blaze/expr/collections.py
```diff
@@ -31,19 +33,200 @@
...
+        try:
+            child = obj._child
+        except AttributeError:
+            return None, None
+        if isinstance(obj, Field) and _isrecord(child):
+            return child, obj
+        obj = child
+
+
+def _extract_key_args(first, second, func):
```
kwmsmith: There is a lot of logic in here, it looks like you're being very thorough, which is excellent.  Can you summarize the cases that you're covering?  I see things like `asc(data, 'a')`, etc, which at first glance, does not strike me as the API we want to expose to the user.

The main cases to cover would be:
* `t.sort(asc(t.a), desc(t.b), ...)`
* `t.sort(asc('a'), desc('b'), ...)`

When would the multi-argument version come in to play?

blaze/compute/sql.py
```diff
@@ -952,6 +953,42 @@ def compute_up(expr, data, **kwargs):
                               group_by=grouper)
```
kwmsmith: Can you refactor the three `compute_up()` implementations to call a common internal function, since they all have the same core logic?

llllllllll:
sorry for not commenting on this sooner, but why do we have a `sort2`, why can't sort just start accepting `asc` or `desc` objects. This should be backwards compatible.

kwmsmith:
> sorry for not commenting on this sooner, but why do we have a sort2, why can't sort just start accepting asc or desc objects. This should be backwards compatible.

The current signature for `sort` is:

```python
sort(child, key=None, ascending=True)
```

Once `sort2` becomes `sort`, its sig will be:

```python
sort(child, *keys)
```

Until we're at that point, mixing `asc`, `desc`, and the `ascending` argument in the current `sort` can lead to confusion and ambiguities, and more logic to implement and test.  For example:

```python
t.sort([asc(t.a), desc(t.b)], ascending=['b'])
```

 I'd rather keep the new stuff separate, and eventually remove the old `sort` after deprecating.

blaze/expr/collections.py
skrah: Good point.  Things like ```asc(data, 'a')``` are actually leftovers from the previous version that could compute ```data.a.asc()```.

I was waiting with this particular cleanup because I wasn't sure if ```Key``` should be an expression and have a ```dshape``` etc.  Can/should expressions have a dshape even if they aren't computable?

llllllllll: I don't event think `asc` and `desc` need to be subclasses of `Expr`. They are not expressions. It could even be that `asc(a.b)` returns a `SortKey(field=a.b, ascending=True)` object and `desc(a.b)` returns a `SortKey(field=a.b, ascending=False)` object. I think that making them exprs will only complicate things. How does this sound?

blaze/expr/collections.py
```diff
@@ -31,19 +33,200 @@
...
+
+    @property
+    def _name(self):
+        return self._child._name
+
+    def __str__(self):
+        return "%s.sort2(%s)" % (self._child, ", ".join(str(x) for x in self.keys))
+
+
+def sort2(first, *rest):
```
llllllllll: How do you feel about just calling this `sort_values` or `sorted`. I would not think to use `sort2` over sort.

llllllllll: also, should we make `sort` raise a deprecation warning?

blaze/expr/collections.py
```diff
@@ -831,7 +1014,7 @@ def shift(expr, n):
     return Shift(expr, n)
 
 
-dshape_method_list.extend([(iscollection, set([sort, head, tail, sample])),
+dshape_method_list.extend([(iscollection, set([asc, desc, sort, sort2, head, tail, sample])),
```
llllllllll: we only want `asc` and `desc` on `Field` objects, right?

blaze/expr/collections.py
```diff
@@ -31,19 +33,200 @@
...
+
+    """Expression sorted in ascending or descending order"""
+
+    _arguments = '_child', 'table', '_key', 'ascending'
+
+    def _dshape(self):
+        return self._child.dshape
+
+    @property
+    def key(self):
```
llllllllll: why do we return a copy in this property?

blaze/expr/collections.py
```diff
@@ -31,19 +33,200 @@
...
+
+    @property
+    def _name(self):
+        return self._child._name
+
+    def _copy(self):
+        """Create a copy of self, detached from any specific table."""
+        return Key(None, None, self._key, self.ascending)
+
+    def _invert(self):
```
llllllllll: I don't think we should allow `~a.b.asc()` to mean `a.b.desc()`. I have no intuition for the unary invert meaning flip the order.

blaze/expr/collections.py
kwmsmith: I don't have a strong opinion on the name -- eventually the current `sort` will be replaced with this new thing, so in the long run, we'll only have `sort`.

+1 on raising a deprecation warning in current `sort`.

blaze/expr/collections.py
skrah: Sounds good.   Do you prefer ```SortKey```?  I have no preference.

blaze/expr/collections.py
kwmsmith: Keeping it narrow to begin with seems good to me.

blaze/expr/collections.py
kwmsmith: Agreed.

blaze/expr/collections.py
skrah: Some test case expected a list instead of a tuple.  Multiple keys are again a leftover from ```data.asc() == data.asc(keys=['a', 'b', ...])```.

blaze/expr/collections.py
kwmsmith: > How does this sound?

+1

blaze/expr/collections.py
skrah: Ah, ```_invert``` is a reserved name.  I just used it here as an ad-hoc name to be used in ```_subexpr_optimize()```.  Perhaps ```_reverse_copy```.

llllllllll: it is only reserved for blaze exprs to dynamically update the `__invert__` method.

blaze/expr/collections.py
skrah: @kwmsmith @llllllllll I think I've addressed all comments except for this one.

 We currently allow ```data.sort_values('a', 'b')```,  which is probably a good shorthand for ```data.sort_values(data.a.asc(), data.b.desc())```.

Do I understand correctly that ```data.sort_values(desc('a'), desc('b'))``` should be disallowed?

skrah: There may be another problem with desc(). With the existing API it is simple to sort by all columns in descending order:

```python
str(bz.compute(data.sort(ascending=False)))
'SELECT numeric_table2.id, numeric_table2.alpha, numeric_table2.beta, numeric_table2.gamma, numeric_table2.delta \nFROM numeric_table2 ORDER BY numeric_table2.id DESC, numeric_table2.alpha DESC, numeric_table2.beta DESC, numeric_table2.gamma DESC, numeric_table2.delta DESC'
```

Now all columns have to be enumerated -- I think that was part of the reason why the first patch allowed ```data.desc()```.

datnamer:
> This is also tied to the implementation where we will write the method as a free function and then dynamically register it to types using the dshape or schema maps.

@llllllllll can you elaborate  on this. Is it currently implemented?

postelrich:

PR #1543, opened 2016-07-13
MNT: Remove blaze/expr/table.py

pratapvardhan:
This PR removes `blaze/expr/table.py` -- #1441 (Remove Table and TableSymbol) removed this file, but was later re-introduced by older PR (#1348)


llllllllll:
Oh, that must have snuck back in when I rebased, that was a large PR. Thanks for removing this again.

PR #1547, opened 2016-07-21
Python 3.5 support

kwmsmith:


kwmsmith:
@llllllllll this is ready for you to review.

* I removed the `blaze/expr/parse.py` since no one is using `exprify()` and its tests were failing under py 3.5.
* There is no pymongo==2.8 package for Py35 in anaconda, so this PR bumps the pymongo version to 3.0.  The testsuite did not have to change to support this.

blaze/compute/tests/test_mongo_compute.py
```diff
@@ -8,12 +8,19 @@
 from toolz import pluck, reduceby, groupby
 
 from datashape import Record
-from blaze import into, compute, compute_up, discover, dshape, data
+from blaze import into as bz_into, compute, compute_up, discover, dshape, data
```
llllllllll: Can we try to keep imports from their canonical path, like `from odo import into` or `from datashape import discover, dshape`

blaze/expr/tests/test_symbol.py
```diff
@@ -805,9 +805,12 @@ def myfunc(x, y, z):
         f = partial(partial(myfunc, 2), 1)
         expr = t.amount.map(f)
         s = str(expr)
-        assert s == ("Map(_child=t.amount, func=partial(partial(%s, 2), 1),"
-                     " _asschema=None, _name0=None)" %
-                     funcname('test_nested_partial', 'myfunc'))
+        fn = funcname('test_nested_partial', 'myfunc')
+        assert s in ("Map(_child=t.amount, func=partial(partial(%s, 2), 1),"
+                     " _asschema=None, _name0=None)" % fn,
+                     # Py3.5 version does partial() folding!
```
llllllllll: This is a nice feature, didn't know about the folding here.

llllllllll:
Looks like solid cleanup, thanks for adding the 3.5 support. I left a few small comments but other than that it lgtm

blaze/compute/tests/test_mongo_compute.py
kwmsmith: Agreed in general; in this case we redefine and wrap `blaze.into` to detect Mongo connection failures, defined below.

PR #1549, opened 2016-07-22
Replace chunks backend with dask for CSV reading.

stefanseefeld:
This implements the simple backend replacment. It was tested against dask master. While explicitly specifying a blocksize affects performance, even the default is (now) better than performance with the chunks backend.

This change may obsolete a number of other functions that are no longer reachable, so a little more cleanup may be warranted.
I'm also not quite sure how much testing this needs.

### DIFF
--- a/blaze/compute/tests/test_csv_compute.py
+++ b/blaze/compute/tests/test_csv_compute.py
@@ -11,7 +11,7 @@ from toolz import first
 from collections import Iterator
 from odo import odo
 from odo.chunks import chunks
-
+import dask.dataframe
 
 def test_pre_compute_on_small_csv_gives_dataframe():
     csv = CSV(example('iris.csv'))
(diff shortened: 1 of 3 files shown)


blaze/compute/tests/test_csv_compute.py
```diff
@@ -37,25 +37,23 @@ def test_compute_chunks_on_single_csv():
...
+def test_compute_calls_lean_projection():
     csv = CSV(example('iris.csv'))
     s = symbol('s', discover(csv))
-    result = pre_compute(s.sort('sepal_length').species,
-                         csv, comfortable_memory=10)
-    assert set(first(result).columns) == \
-            set(['sepal_length', 'species'])
+    result = compute(s[s.sepal_length > 5.0].species,
+                     csv, comfortable_memory=10)
+    assert len(result) == 118
```
kwmsmith: To verify we're doing the lean projection, this needs to be something like:

```python
result = pre_compute(s[s.sepal_length > 5.0].species, csv, comfortable_memory=10)
assert set(result.columns) == set(['sepal_length', 'species'])
```

Specifically, we need to test the `pre_compute()` stage explicitly.

stefanseefeld: Fixed.

kwmsmith:
We can remove these functions now that we no longer need chunks with CSV:

    https://github.com/blaze/blaze/blob/master/blaze/compute/csv.py#L76-L97

Looking at the coverage reports, these functions aren't hit anymore during testing.

llllllllll:
what version of pandas is required by dask.dataframe?

stefanseefeld:
pandas>=0.18.0

llllllllll:
I was going to say that I am using blaze on a project that supported pandas 0.17, but we just had a talk and decided to drop pandas 0.17 support so that is no longer an issue.

kwmsmith:
@llllllllll great -- any other comments on this PR?  Otherwise will merge.

llllllllll:
We need to add `dask[dataframe]` to the requirements, other than that it lgtm

Issue #1557, opened 2016-08-01
Strange things happen when mixing different instances of a bz.Data object


dhirschfeld:
In my larger example I get incorrect SQL issued and an SQL error. In my repro below it throws a ValueError.

It may be the case that the answer is simply *"Don't Do That"* but in that case it may still be nice if it's possible to detect it to then throw a more meaningful error.

dhirschfeld:
Using the same test case as in https://github.com/blaze/blaze/issues/1552#issuecomment-235126768 I observe the following behaviour:
```python
In [8]: def get_data():
   ...:     """Returns a *new* `bz.Data` object wrapping the same table"""
   ...:     return bz.Data(T1)

In [9]: mask = get_data().label_id.isin({2})

In [10]: odo(get_data()[mask], pd.DataFrame)
Traceback (most recent call last):

  File "<ipython-input-10-340392c503d0>", line 1, in <module>
    odo(get_data()[mask], pd.DataFrame)

  File "c:\dev\packages\blaze\blaze\expr\expressions.py", line 132, in __getitem__
    return selection(self, key)

  File "c:\dev\packages\blaze\blaze\expr\expressions.py", line 523, in selection
    subexpr = common_subexpression(table, predicate)

  File "c:\dev\packages\blaze\blaze\expr\core.py", line 502, in common_subexpression
    'No common leaves found in expressions %s' % list(exprs)

ValueError: No common leaves found in expressions [<'Table' data; _name='_3', dshape='var * {  timestamp: string,  label_id: map[int32, ...'>,    label_id
0     False
1     False
2      True
3      True
4     False
5      True
6     False
7     False
8      True
9     False
...]
```

dhirschfeld:
In my larger example I get an even more opaque DatabaseError but I haven't been able to isolate a test case to repro it exactly.

As you can see below though it looks like its raising an exception in the same `common_subexpression` but somehow that's being caught and actual incorrect SQL is issued to the DB
```
  File "c:\dev\packages\blaze\blaze\expr\core.py", line 502, in common_subexpression
    'No common leaves found in expressions %s' % list(exprs)
```

```python
In [5]: mask = schema.HOLIDAY_SCHEDULE.hs_hc_id.isin({1})

In [6]: schema.HOLIDAY_SCHEDULE[mask]
Traceback (most recent call last):

  File "<ipython-input-6-5927c2648912>", line 1, in <module>
    schema.HOLIDAY_SCHEDULE[mask]

[...]

PR #1558, opened 2016-08-01
Fixed possible exception when repr'ing an expression

dhirschfeld:
Partially fixes an issue raised in #1557

### DIFF
--- a/blaze/interactive.py
+++ b/blaze/interactive.py
@@ -441,7 +441,10 @@ def _choose_repr(self):
         return new_repr(self)
     else:
         warnings.warn(_warning_msg, DeprecationWarning, stacklevel=2)
-        return expr_repr(self)
+        try:
+            return expr_repr(self)
+        except Exception as exc:
+            return new_repr(self)
 
 
 def _warning_repr_html(self):
@@ -449,7 +452,10 @@ def _warning_repr_html(self):
         return new_repr(self)
     else:
         warnings.warn(_warning_msg, DeprecationWarning, stacklevel=2)
-        return to_html(self)
+        try:
+            return to_html(self)
+        except Exception as exc:
+            return new_repr(self)
 
 
 def new_repr(self):


dhirschfeld:
Not sure if this is the best fix - if it can more generally happen that str/repr raises an exception then it may be best to protect the whole thing with a try/except

blaze/expr/core.py
```diff
@@ -499,7 +499,7 @@ def common_subexpression(expr, *exprs):
     # no common leaves therefore no common subexpression
     if not leaves:
         raise ValueError(
-            'No common leaves found in expressions %s' % list(exprs)
+            'No common leaves found in expressions %s' % list(map(str, exprs))
```
dhirschfeld: I'm not sure this is the correct fix since in general it doesn't give as much info as the repr does.

I see however that blaze is moving towards a repr which doesn't compute and therefore won't be prone to throwing an exception so this won't be an issue in future.

dhirschfeld:
I just force pushed a better, more general fix which is to protect the repr's themselves with a try/except.

Although this behaviour will change in future this fix gives a vastly (IMHO) improved user experience right now by returning something sensible in the case where an exception would have otherwise been thrown.

The current behaviour whereby a repr can raise can be a source of significant confusion where unprotected reprs are used in the codebase  (c.f. #1557) as it will raise a completely unrelated exception.

kwmsmith:
FYI, we've deprecated the automatic-compute-during-repr behavior.  In the next release, I will be flipping the `use_new_repr` flag found here (currently False):

https://github.com/blaze/blaze/blob/master/blaze/interactive.py#L436

Could you try flipping that bit and see if that fixes the issue you're seeing here?

dhirschfeld:
Setting `use_new_repr = True` works fine since that is effectively what this PR is doing (conditional on the existing repr raising an exception).

This PR was predicated on the fact that you didn't what change to using the new repr just yet (for whatever reason). That being the case, the reprs need to be protected against raising an exception since it can cause significant issues - hence the solution in this PR.

I'd be perfectly happy with just setting `use_new_repr = True` or equivalently removing the old repr entirely.

kwmsmith:
> That being the case, the reprs need to be protected against raising an exception since it can cause significant issues.

Yes, we feel this pain *frequently* with Blaze, given that it's not uncommon for a malformed expression or poorly supported backend to raise an exception when computing.  The Python runtime gets very confused when `repr` raises an exception.  This is one reason why we decided to make blaze's expression repr less sophisticated and more standard.

> I'd be perfectly happy with just setting use_new_repr = True or equivalently removing the old repr entirely.

Yes, we'll be removing the old `__repr__` behavior in the next release, so that should resolve the issue you're seeing here.

FYI, to get a quick look at the computed results of an expression, you can use the `.peek()` method on an expression object to trigger computation -- if the computation raises an exception, you'll get a standard traceback.

postelrich:

Issue #1551, opened 2016-07-25
Switch to `return_type='core'` in `compute`


kwmsmith:
We agreed to deprecate the `'native'` return type in favor of `'core'` for `compute` after 0.11.

Initial attempts to switch over to `return_type='core'` as the default revealed more than 100 test failues, so there's some work to be done here.

PR #1559, opened 2016-08-01
compute(): change default of 'return_type' parameter to 'core'.

skrah:
Implementation for #1551.

### DIFF
--- a/blaze/compute/core.py
+++ b/blaze/compute/core.py
@@ -367,12 +367,12 @@ def compute(expr, d, return_type='core', **kwargs):
         The blaze expression to compute.
     d : any
         The data source to compute expression on.
-    return_type : {'native', 'core', type}, optional
-        Type to return data as. Defaults to 'native' but will be changed
-        to 'core' in version 0.11.  'core' forces the computation into a core
-        type. 'native' returns the result as is from the respective backend's
-        ``post_compute``. If a type is passed, it will odo the result into the
-        type before returning.
+    return_type : {'core', 'native', type}, optional
+        Type to return data as.  Defaults to 'core' since version 0.11.1.
+        'core' forces the computation into a core type. 'native' returns
+        the result as is from the respective backend's ``post_compute``.
+        If a type is passed, it will odo the result into the type before
+        returning.
 
     Examples
     --------
--- a/blaze/compute/core.py
+++ b/blaze/compute/core.py
@@ -358,7 +358,7 @@ def swap_resources_into_scope(expr, scope):
 
 
 @dispatch(Expr, Mapping)
-def compute(expr, d, return_type=no_default, **kwargs):
+def compute(expr, d, return_type='core', **kwargs):
     """Compute expression against data sources.
 
     Parameters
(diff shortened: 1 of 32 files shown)


skrah:
@kwmsmith I think this is ready for review.  I had to tweak ```kwargs``` in two places, which is a bit ugly, but also happens elsewhere.

blaze/compute/core.py
```diff
@@ -367,20 +367,20 @@ def compute(expr, d, return_type=no_default, **kwargs):
...
+        returning.
 
     Examples
     --------
     >>> t = symbol('t', 'var * {name: string, balance: int}')
     >>> deadbeats = t[t['balance'] < 0]['name']
 
     >>> data = [['Alice', 100], ['Bob', -50], ['Charlie', -20]]
-    >>> list(compute(deadbeats, {t: data}))
+    >>> list(compute(deadbeats, {t: data}, return_type='native'))
```
kwmsmith: is `return_type = 'native'` necessary here?  I would have expected `return_type = 'core'` to return a `list`.

skrah: @kwmsmith Currently the return type is necessary:

```python
>>> list(compute(deadbeats, {t: data}))
Traceback (most recent call last):
...
AttributeError: 'NoneType' object has no attribute 'measure'
```

In many places ```return_type='native'``` is hardcoded (for lack of a better word)
when ```compute``` is used internally.  Example:

https://github.com/blaze/blaze/blob/master/blaze/compute/sql.py#L125

Should that always be the case?  If so, perhaps the argument is missing in 
some places.

skrah: @kwmsmith Currently the return type is necessary:

```python
>>> list(compute(deadbeats, {t: data}))
Traceback (most recent call last):
...
AttributeError: 'NoneType' object has no attribute 'measure'
```

In many places ```return_type='native'``` is hardcoded (for lack of a better word)
when ```compute``` is used internally.  Example:

https://github.com/blaze/blaze/blob/master/blaze/compute/sql.py#L125

Should that always be the case?  If so, perhaps the argument is missing in 
some places.

kwmsmith:
There are a number of spots when computing against the `python` backend where `compute(..., return_type='native')` is specified and there's an explicit `list()` coercion applied to the result.  The `list()` call shouldn't be necessary, unless there's something I'm misunderstanding.  Can you address those?

blaze/compute/core.py
skrah: Just for reference, the commit that added the arg was 185cc7be7.

kwmsmith: I see -- `compute(deadbeats, {t: data})` triggers an odo conversion from `itertools.chain` to a DataFrame, and that path appears to be broken in odo (separate issue).

One reason for the `return_type` kwarg was to reduce the amount of explicit `odo`ing and type conversions when calling `bz.compute()`.  To make it more streamlined, please convert

```
list(compute(deadbeats, {t: data}, return_type='native')
```

to

```
compute(deadbeats, {t: data}, return_type=list)
```

skrah:
This works: ```compute(t['name'], data, return_type=list) == [x[0] for x in data]```.

Omitting ```return_type``` entirely doesn't work and ```return_type='native'``` needs the ```list()```.


So I guess I have to figure out the exact return types.

kwmsmith:
> So I guess I have to figure out the exact return types.

Not everywhere -- only in the places where there's a need for an explicit coercion that isn't already a 'native' or 'core' type, or where the `odo` conversion logic needs help.  So the SQL backends can use `return_type='native'`, while the `list` docstring examples should use:

```
compute(expr, data, return_type=list)
```

skrah:
Okay, great.  That saves some work.

kwmsmith:
related: https://github.com/blaze/odo/issues/465

postelrich:

Issue #1552, opened 2016-07-25
Filtering a table using `==`  doesn't work when a ForeignKey is involved (and can silently return incorrect results)


dhirschfeld:
As shown below when using the `==` operator to filter rows only a single match is returned. In comparison, the `isin` operator correctly returns the full set of matching rows:
```python
In [16]: import blaze as bz

In [17]: bz.__version__
Out[17]: '0.10.1'

In [19]: T[T.cce_contractid.isin({12199})].nrows
Out[19]: 35088

In [20]: T[T.cce_contractid == 12199].nrows
Traceback (most recent call last):

  File "<ipython-input-20-22f63c35d212>", line 1, in <module>
    T[T.cce_contractid == 12199].nrows

  File "C:\Python\lib\site-packages\blaze\expr\expressions.py", line 222, in __getattr__
    result = object.__getattribute__(self, key)

AttributeError: 'Slice' object has no attribute 'nrows'


In [21]: odo(T[T.cce_contractid == 12199], pd.DataFrame).shape
C:\Python\lib\site-packages\blaze\compute\sql.py:1414: UserWarning: The order of the result set from a Slice expression computed against the SQL backend is not deterministic.
  warnings.warn('The order of the result set from a Slice expression '
Out[21]: (1, 20)
```

dhirschfeld:
It appears that this only happens when the column used as a filter has a foreign key constraint and is thus defined as a map in datashape
**Repro:**
```python
from itertools import chain

import blaze as bz
import numpy as np
from numpy.random import choice, randint
from odo import odo
import pandas as pd
import sqlalchemy as sa

engine = sa.create_engine('sqlite://')
metadata = sa.MetaData(bind=engine)
T1 = sa.Table('demo', metadata,
   sa.Column('timestamp', sa.String, primary_key=True),
   sa.Column('label_id', sa.Integer, sa.ForeignKey("labels.label_id"), nullable=False),
)
T2 = sa.Table('labels', metadata,
   sa.Column(
       'label_id',
       sa.Integer,
       primary_key=True,
       nullable=False,
   ),
   sa.Column('label', sa.String),
)

metadata.create_all()


labels = pd.DataFrame(
    list(zip(range(4), ['A','B','C','D'])),
    columns=['label_id', 'label']
)
odo(labels, T2)


dates = pd.date_range('01-Feb-2016', '01-Mar-2016', freq='D')
df = pd.DataFrame({'timestamp': list(map(str, dates))})
[...]

PR #1561, opened 2016-08-02
Issue #1552: Fix method discovery (foreign key)

skrah:
This fixes the method discovery if the datashape represents a foreign key.  It further entrenches the assumption that datashape.Map *always* represents a foreign key, see e.g.

https://github.com/blaze/blaze/blob/master/blaze/expr/expressions.py#L176

Compare with the datashape docs ("meant to be useful outside of that context "):

https://github.com/blaze/datashape/blob/master/docs/source/types.rst#maps

### DIFF
--- a/blaze/expr/expressions.py
+++ b/blaze/expr/expressions.py
@@ -220,9 +220,13 @@ class Expr(Node):
         except AttributeError:
             fields = dict(zip(map(valid_identifier, self.fields), self.fields))
 
+            measure = self.dshape.measure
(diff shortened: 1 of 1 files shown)


kwmsmith:
Thanks @skrah, and thanks for documenting the assumptions made here.  The map-in-datashape-represents-foreign-key assumption is pervasive throughout Blaze, which is acceptable given Blaze's data analysis and tabular data focus.

PR #1562, opened 2016-08-02
Issue #1546: Fix equality failure for simple types.

### LINKED ISSUE #1546: Data constructor requires 'name' to be set
If the optional ```name``` argument in the ```data``` function isn't set, some operations behave unexpectedly:

```python
>>> from blaze import data, compute
>>> 
>>> xd = data(157)
>>> yd = data(101)
>>> compute(xd == yd)
True
```

This could be prevented by automatically generating unique symbols:

```python

>>> xd = data(157, name="__v0001")
>>> yd = data(101, name="__v0002")
>>> compute(xd == yd)
False
```

skrah:
This fixes #1546 until a way is found to support name=None for equality.

### DIFF
--- a/blaze/interactive.py
+++ b/blaze/interactive.py
@@ -94,10 +94,7 @@ class _Data(Symbol):
             cls,
             data,
             dshape,
-            name or (
-                next(names)
-                if isrecord(dshape.measure) else None
-            ),
+            name or next(names),
         )
 
     def _resources(self):
--- a/blaze/interactive.py
+++ b/blaze/interactive.py
@@ -90,11 +90,15 @@ class _Data(Symbol):
     _arguments = 'data', 'dshape', '_name'
 
     def __new__(cls, data, dshape, name=None):
+        if name is None:
+            name = next(names)
+        elif not isinstance(name, _strtypes):
+            raise TypeError("'name' argument must be a string")
         return super(Symbol, cls).__new__(
             cls,
             data,
             dshape,
-            name or next(names),
+            name,
         )
 
     def _resources(self):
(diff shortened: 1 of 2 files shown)


kwmsmith:
So the result of this change is that all `_Data` objects will have a non-None `name` attribute, which will disambiguate them when equality testing?  Makes sense to me.  Should we go a step further and validate the `name` type if passed in -- something like:

```
if not isinstance(name, str):
   raise ValueError(...)
```

Even more thorough would be to disallow `name` to be the name of an existing `_Data` object, but that's probably outside the scope of this PR.

@llllllllll do you know why the original logic had the `isrecord()` special case?

skrah:
Yes, checking for uniqueness would be good:

```
>>> x = data(1, name='_4')
>>> y = data(2, name='_4')
>>> 
>>> x == y
True
```

But I also think it's better for a separate issue.

llllllllll:
Why does the name need to be unique, if we just make `data` hash like the other expr types then `data(1, name='_4') is data(1, name='_4')`

llllllllll:
Also the record stuff is old, that is just so that tables automagically get named things like `_%d`. We could just always do that.

skrah:
The underlying problem is here:

https://github.com/blaze/blaze/blob/master/blaze/compute/core.py#L330s

toolz.merge() eliminates keys with the same name.


I'm still unsure what namespace guarantees blaze makes, but as a new user I'd probably expect this:

All names of the form "_123" are unique and reserved for internal use (we could actually check for that).

If a non-reserved name is assigned and re-used: Is there some sort of name-binding model that says that old names stay bound?

All names seem global to me, in which case I'd either expect that the new object with the same name overrides the old one entirely or that an error is raised.

postelrich:

Issue #1564, opened 2016-08-05
Some dask tests cannot fail


skrah:
`dask.Scalar` equality returns a `dask.Scalar` instead of a `bool`.  Using this scalar in `assert` is always true.

Example:

 https://github.com/blaze/blaze/blob/master/blaze/compute/tests/test_dask_dataframe.py#L124

```python
from blaze import compute, symbol
from blaze.expr import mean
import pandas as pd
import dask.dataframe as dd
import dask


t = symbol('t', 'var * {name: string, amount: int, id: int}')
df = pd.DataFrame([['Alice', 100, 1], ['Bob', 200, 2], ['Alice', 50, 3]], columns=['name', 'amount', 'id'])
ddf = dd.from_pandas(df, npartitions=2)
x = compute(mean(t['amount']), ddf, return_type='native')

>>> x
<dask.dataframe.core.Scalar object at 0x7f06ad06d048>
>>> assert x == 10
>>> dask.compute(x == 10)
(False,)
```

PR #1565, opened 2016-08-05
Issue #1564: compute dask.dataframe.core.Scalar all the way to a native type.

skrah:


skrah:
@kwmsmith This adresses #1564.

Synopsis: `dask.dataframe.core.Scalar` wasn't fully computed, so some `asserts` always passed in the tests.

blaze/compute/tests/test_dask_dataframe.py
```diff
@@ -121,13 +121,13 @@ def test_selection_inner_inputs():
 
 
 def test_reductions():
```
llllllllll: should we check the type here to make sure it is normalized to float

skrah: It isn't: `count, sum, min, max` return integers for both pandas and dask dataframes.  `mean, var, std` return floats here, probably since the exact values happen to be floats.

Dask dataframe results also end up as `numpy.int64` and `numpy.float64`.

PR #1567, opened 2016-08-08
Incorrect results computing against interactive dict data

necaris:
This introduces a failing test for interactive computation on a list-of-dictionaries. No solution as yet but opening the PR should allow the Travis-CI tests to complete, and provide a place for discussion. The test is in `blaze/compute/tests/test_python_compute.py` and named `test_selection_interactive`.

The test can be run on its own with:

```
py.test blaze/compute/tests/test_python_compute.py::test_selection_interactive
```

### DIFF
--- a/blaze/compute/core.py
+++ b/blaze/compute/core.py
@@ -139,8 +139,8 @@ def top_then_bottom_then_top_again_etc(expr, scope, **kwargs):
     top_to_bottom -- older version
     bottom_up -- older version still
     """
-    rl = kwargs.get('_recurse_level', 0)
-    kwargs['_recurse_level'] = rl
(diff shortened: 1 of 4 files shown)


necaris:
Having done some more work digging into this, as far as I can tell the issue _seems_ to be that in some cases when we are mapping functions across iterators, an iterator's `__next__` is being called while somewhere else there is another reference to the same iterator, which is not getting its state restored appropriately. (My best guess is in the `pre_compute` function [here](https://github.com/blaze/blaze/blob/master/blaze/compute/python.py#L110)). Thus the iterator is advancing twice as fast as desired.

postelrich:

PR #1577, opened 2016-08-22
Issue1513

### LINKED ISSUE #1513: sort should accept a `Field` object
`expr.sort(expr.field)` should be the same as `expr.sort('field')`

ywang007:
closes #1513

### DIFF
--- a/blaze/expr/collections.py
+++ b/blaze/expr/collections.py
@@ -123,10 +123,8 @@ def sort(child, key=None, ascending=True):
     if key is None and isrecord(child.dshape.measure):
         key = child.dshape.measure.names
     if isinstance(key, (list, tuple)):
-        key = keys_to_validate = tuple([k._name if isinstance(k, Field) else k for k in key])
+        key = keys_to_validate = tuple(key)
     else:
-        if isinstance(key, Field):
-            key = key._name
         keys_to_validate = (key,)
     for k in keys_to_validate:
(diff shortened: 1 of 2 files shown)


blaze/expr/collections.py
```diff
@@ -123,8 +123,10 @@ def sort(child, key=None, ascending=True):
     if key is None and isrecord(child.dshape.measure):
         key = child.dshape.measure.names
     if isinstance(key, (list, tuple)):
-        key = keys_to_validate = tuple(key)
```
llllllllll: if the key is an instance of field then we should verify that the child of the field is compatible with the eexpression being sorted.

We should not allow things like:
```python
a = bz.data(...)
b = bz.data(...)
a.sort(b.field)
```

blaze/expr/collections.py
```diff
@@ -134,9 +134,17 @@ def sort(child, key=None, ascending=True):
             if k not in child.dshape.measure.names:
                 msg = "sort key {} is not a column of schema {}"
                 raise ValueError(msg.format(k, child.dshape.measure))
+        elif isinstance(k, Field):
+            if k._resources() != child._resources():
```
llllllllll: I think the better check is just `not k._child.isindentical(child)`, we can always be more relaxed later.

postelrich:

PR #1581, opened 2016-08-29
Correctly reconstruct SQL Selects with Joins

necaris:
Ensure that if a select statement is a JOIN, calling `reconstruct_select` correctly reconstructs the JOIN statement instead of dropping it into a multi-table select.


postelrich:
+1 on merging this. Prevents ending up with a multitable select which is SLOW.

kwmsmith:
LGTM.  Thanks for the test as well.

llllllllll:
Thank you!

Issue #1542, opened 2016-07-08
Turn off eager expression reordering


kwmsmith:
```python
In [2]: t = bz.symbol('t', 'var * {name: string, id: int}')

In [6]: t[['name']][t.name == 'Ying']
Out[6]: t[t.name == 'Ying'][['name']]

In [7]: t[['name', 'id']][['name']]
Out[7]: t[['name']]

In [8]: t[['name', 'id']][['id']]
Out[8]: t[['id']]

In [9]: t[['name', 'id']][t.id < 100][['id']]
Out[9]: t[t.id < 100][['id']]
```

We have usecases that build up expressions incrementally.  The final composite expression needs to be able to indicate which smaller expressions it is derived from.  Blaze's eager expression rewriting makes it impossible to do this at times.

One solution would be to turn off any eager expression rewriting or reordering, and instead do rewriting / reordering as a `compute()` time optimization.

skrah:
@kwmsmith If eager rewriting is turned off, are expressions that are not commutative allowed to fail?

Example:

```psql
blaze_test=# select * from alice_bob;
 id | name  | amount 
----+-------+--------
  1 | Alice |    100
  2 | Bob   |    200
  3 | Alice |     50
(3 rows)

blaze_test=# select * from (select name from alice_bob) as t where amount < 100;
ERROR:  column "amount" does not exist
LINE 1: ...ct * from (select name from alice_bob) as t where amount < 1...
                                                             ^
blaze_test=# select name from (select * from alice_bob where amount < 100) as t;
 name  
-------
 Alice
(1 row)
```

Eager rewriting currently lets some computations succeed that would otherwise fail naturally like above.  The test suite contains some examples (e.g. test_spark).

kwmsmith:
For this specific example, what is the blaze expression that's being rewritten / reordered here?

Is it something like:

```
t[t.amount < 100].name
```

When I repr this expression, it isn't being reordered -- it's symbolic representation looks identical to the expr as input.

The expression reordering I'm referring to here is entirely at the symbolic Blaze level *before* any `compute` happens.  It's fine if Blaze reorders an expression as an optimization pass, but that has to happen *during* `compute`, not when the original symbolic expr is constructed.

So, for instance, when I do a projection followed by a selection:
```
t[['name']][t.amount < 100]
```

The resulting symbolic expression is constructed as a selection, *then* a projection:

```
t[t.amount < 100][['name']]
```

[...]

PR #1582, opened 2016-08-30
[WIP] Issue #1542: Turn off eager expression reordering.

skrah:


skrah:
@kwmsmith @llllllllll 

This is a rough initial sketch for #1542 to highlight a couple of problems.

1) Eager rewriting occurs everywhere, also implicitly, in special methods
   like `_project()` and `_get_field_name()`.

2) Sometimes lookahead is required to get the same results as before,
   see https://github.com/blaze/blaze/pull/1582/files#diff-17b7f303fd5bc74676d737b3cab1bd09R55 .

3) I did not manage to integrate rewriting into `optimize()`, since `optimize()`
   is called after `bottom_up_until_type_break()`.


So I added a new `rewrite.py` file and `rewrite()` is called at the beginning
of `compute()`.


It would be great to hear if this is similar to what you had in mind before
I proceed (it isn't finished at all).

skrah:
I forgot: The `transform` name is already taken (see `collections.py`).

postelrich:

PR #1585, opened 2016-09-08
ENH: fix dayof(week|year) for sql and update dtypes for datetime exprs

llllllllll:
We were using int64 where we could store the result in a much smaller type. We cannot go all the way down to i8 for these because sql small integer maps to an i16 so it is just easier to leave this here than to make i8 map to small integer also.

This mainly cleans up that `extract` uses the names `dow` and `doy` for `dayofweek` and `dayofyear`.

I also correct that sql dow uses sunday=0 but `datetime.datetime` and `pandas` use `monday=0`

### DIFF
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -94,6 +94,7 @@ from ..expr import (
     StrSlice,
     var,
 )
+from ..expr.datetime import dayofweek
 from ..expr.strings import len as str_len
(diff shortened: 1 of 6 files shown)


llllllllll:
ping @stefanseefeld this updates the new datetime stuff you worked on, can you take a look at this?

llllllllll:
also, I think the doc test issue is around the dask change in odo

Issue #1584, opened 2016-09-08
Select one raises coerce_core error


postelrich:
When I try to do a core compute on sqlalchemy connected dataset, `coerce_core` fails when selecting a single row.  If I set `return_type=pd.DataFrame` it works fine.  Seems to be a missed edge case with `coerce_core`.

```
In [3]: data[0].compute()
/home/myuser/miniconda/envs/myenv/lib/python3.5/site-packages/blaze/compute/sql.py:1414: UserWarning: The order of the result set from a Slice expression computed against the SQL backend is not deterministic.
  warnings.warn('The order of the result set from a Slice expression '
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-67ecd54b524c> in <module>()
----> 1 data[0].compute()

---> 72         return bz.compute(self.compile(), return_type='core')


/home/myuser/miniconda/envs/myenv/lib/python3.5/site-packages/multipledispatch/dispatcher.py in __call__(self, *args, **kwargs)
    162             self._cache[types] = func
    163         try:
--> 164             return func(*args, **kwargs)
    165
    166         except MDNotImplementedError:

/home/myuser/miniconda/envs/myenv/lib/python3.5/site-packages/blaze/interactive.py in compute(expr, **kwargs)
    193         raise ValueError("No data resources found")
    194     else:
--> 195         return compute(expr, resources, **kwargs)
    196
    197

/home/myuser/miniconda/envs/myenv/lib/python3.5/site-packages/multipledispatch/dispatcher.py in __call__(self, *args, **kwargs)
    162             self._cache[types] = func
    163         try:
--> 164             return func(*args, **kwargs)
    165
    166         except MDNotImplementedError:

/home/myuser/miniconda/envs/myenv/lib/python3.5/site-packages/blaze/compute/core.py in compute(expr, d, return_type, **kwargs)
    419     # (python type, pandas Series/DataFrame, numpy array)
    420     elif return_type == 'core':
--> 421         result = coerce_core(result, expr.dshape)
    422     # user specified type
    423     elif isinstance(return_type, type):

/home/myuser/miniconda/envs/myenv/lib/python3.5/site-packages/blaze/interactive.py in coerce_core(result, dshape, odo_kwargs)
    314     else:
    315         msg = "Expr does not evaluate to a core return type"
--> 316         raise ValueError(msg)
    317
    318     return result

ValueError: Expr does not evaluate to a core return type
```

[...]

PR #1586, opened 2016-09-09
#1584 coerce record to df

postelrich:


postelrich:

blaze/compute/core.py
```diff
@@ -544,7 +544,7 @@ def coerce_core(result, dshape, odo_kwargs=None):
         return result
     elif isscalar(dshape):
         result = coerce_scalar(result, dshape, odo_kwargs=odo_kwargs)
-    elif istabular(dshape) and isrecord(dshape.measure):
+    elif istabular(dshape) or isrecord(dshape):
```
llllllllll: I don't think this is correct, a scalar record is _not_ a dataframe, it is a single row of a dataframe.

postelrich: Depends on how we want to treat the selection of a single row. If it's a scalar then `isscalar` can be extended to handle it.

PR #1589, opened 2016-09-26
FIX: tweak `data()` logic to handle dshape overrides.

kwmsmith:


blaze/tests/test_interactive.py
```diff
@@ -201,6 +201,19 @@ def test_table_resource():
         assert isinstance(t.data, CSV)
         assert into(list, compute(t)) == into(list, csv)
 
+def test_explicit_dshape():
+    ds = dshape("var * {a: ?float64, b: ?string, c: ?float32}")
+    s = """a,b,c
```
llllllllll: could we use `textwrap.dedent` on this? having strings go back to 0 indentation makes it harder to follow the flow of code when scrolling though a file.

blaze/tests/test_interactive.py
```diff
@@ -201,6 +201,19 @@ def test_table_resource():
         assert isinstance(t.data, CSV)
         assert into(list, compute(t)) == into(list, csv)
 
+def test_explicit_dshape():
+    ds = dshape("var * {a: ?float64, b: ?string, c: ?float32}")
```
llllllllll: can you put a comment saying that if this isn't overridden it will return `c: int64` instead. This will help debug this later if it ever changes.

llllllllll:
two small comments but lgtm, thanks for this!

blaze/tests/test_interactive.py
kwmsmith: I never knew there was a `textwrap` module -- that's awesome, thanks for the pointer.

blaze/tests/test_interactive.py
kwmsmith: Yes.

PR #1591, opened 2016-10-04
Support joins with foreign keys represented as `map[]` datashapes.

### LINKED ISSUE #1568: DataShape not NumPy-compatible prevents joining tables
Using foreign-keys in a database is a very common thing to do however it results in a map type in datashape which apparently prevents performing even a simple join

kwmsmith:
Fixes #1568.


dhirschfeld:
Thanks @postelrich - good to see the project moving forward!

PR #1593, opened 2016-10-07
Fix dask `chunks` args to match length of `shape`.

kwmsmith:
Exposed when updating to dask `0.11.1`.


kwmsmith:
This unblocks #1592.

llllllllll:
Nice! This is blocking other work of mine so I am going to merge.

PR #1597, opened 2016-11-01
When converting `return_type` in `compute`, make use of known datashape

necaris:
Currently, `odo` is not able to make use of the right data-types for its return values because it does not have knowledge of the result expression's datashape -- this passes the shape into `odo.into` to ensure that's present. 

Actual code by @mcg1969, I'm just the courier ;-)


necaris:
Closed by https://github.com/blaze/blaze/commit/5a16c68517799a3e82a3440c736ed7073825ba89#diff-7072562c8b67248d9974b044058a91d4R427

PR #1601, opened 2016-11-03
BLD: add extra install targets and ci

llllllllll:


llllllllll:
this is covered in: #1598

PR #1605, opened 2016-11-18
Fix Odo dependency for 0.11.3 release

necaris:
This should provide the basis for an `0.11.3` release containing critical bugfixes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ install:
   - conda update conda
 
   # Install dependencies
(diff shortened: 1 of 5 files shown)


necaris:
At the moment this seems to be breaking the build -- since this is happening across branches, and since `odo` is installed from Git for CI purposes, my hypothesis is that it's because of conflicting `dask` versions here vs transitively from `odo`.

necaris:
Note that this explicitly _excludes_ Odo 0.5.2 as a dependency until we figure out why the CI builds using `master` are failing, and fix them if possible.

necaris:
This doesn't seem to pass CI no matter what, even when it's reduced to updating Dask dependencies (for the transitive `odo` dependency) -- so going ahead and closing this in favor of a more careful examination of upstream + errors down the road.

PR #1611, opened 2016-12-20
Variadic expressions

llllllllll:
Adds a new VarArgsExpr/VarArgs pair of objects for representing
sequences of known types but unknown length in the expression tree. This
allows us to use multipledispatch to define compute handlers for
expressions that conceptually accept a sequence like `Merge`.

This also implements `Merge` as a variadic expression.

Future work will be porting `Concat`.

### DIFF
--- a/blaze/compute/core.py
+++ b/blaze/compute/core.py
@@ -6,11 +6,11 @@ import itertools
 import numbers
 import warnings
 
+from odo import odo
+import pandas as pd
 import toolz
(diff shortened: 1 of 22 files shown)


llllllllll:
NOTE: I didn't implement this for the spark backend because I have no idea how to fix that. I will leave it unimplemented for now.

llllllllll:
ping @necaris. Do you have any thoughts on the implementation of this?

PR #1612, opened 2016-12-21
ENH: add drop_fields method for tabular expressions

llllllllll:
simple helper for dropping fields of a table


llllllllll:
ping @mtydykov

blaze/expr/expressions.py
```diff
@@ -949,15 +955,60 @@ def ndim(expr):
...
 
 
+def drop_field(expr, field, *fields):
+    """Drop a field or fields from a tabular expression.
+
+    Parameters
+    ----------
+    expr : Expr
+        A tabular expression to drop columns from.
+    *fields
```
mtydykov: `field` is missing from docstring

llllllllll: This is intentional, this pattern is to make a *args which requires at least one argument.

blaze/expr/tests/test_expr.py
```diff
@@ -331,3 +331,44 @@ def test_cast():
...
+        " (['a', 'b', 'c', 'd'])"
+    )
+
+    # test dropping a mix of correct and missing fields
+    with pytest.raises(ValueError) as e:
+        s.drop_field('a', 'fake')
+    assert (
+        str(e.value) ==
+        "fields ['fake'] were not in the fields of expr (['a', 'b', 'c', 'd'])"
+    )
```
mtydykov: Should we add an assertion after each of these failed drops to ensure that the expr still has the fields we expect?

mtydykov:
One last comment, but otherwise 👍

blaze/expr/tests/test_expr.py
llllllllll: expressions are immutable so this just returns the expression with less fields. The name is based on pandas `drop(axis=1)` which is not inplace by default.

PR #1625, opened 2017-04-04
DOC: Fix docstrings for expressions module.

ehebert:
Fix the reference to `projection` function, which was a documentation build
error.

Also, change docstring for `Coalesce` to clean up warnings on indentation and
lines between examples.

Discovered while building docs via `make info`.

### DIFF
--- a/blaze/expr/expressions.py
+++ b/blaze/expr/expressions.py
@@ -857,19 +857,24 @@ def binop_inputs(expr):
 class Coalesce(Expr):
     """SQL like coalesce.
 
-    coalesce(a, b) = {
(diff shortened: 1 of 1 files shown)


blaze/expr/expressions.py
```diff
@@ -857,19 +857,24 @@ def binop_inputs(expr):
 class Coalesce(Expr):
     """SQL like coalesce.
 
-    coalesce(a, b) = {
-        a if a is not NULL
-        b otherwise
-    }
+    .. code-block:: python::
```
llllllllll: just `.. code-block python`, no extra colons

PR #1636, opened 2017-05-05
MAINT: Move type predicates into dedicated module.

ehebert:
Break out `iscorescalar`, `iscoresequence`, and `iscoretype` into a module
dedicated to type checking.

Putting these functions in their own module helps avoid an import cycle while
breaking out other components from the interactive module.


ehebert:
@llllllllll I fixed up the `__future__` imports (for Python 2.7), and rebased against HEAD master (which contained #1637). This is green now, and ready for review.

PR #1638, opened 2017-05-08
ENH: Add `literal`, analogous to `data`.

ehebert:
The purpose of the `literal` function is to create leafs which delegate the repr to
the underlying data; e.g. `repr(frozenset([1, 2, 3]))` and
`repr(literal(frozenset([1, 2, 3]))` would be equivalent.

Change the `Data` class to `Literal`, and change the repr logic so that if
`_name` is `None` the repr of the `data` object is used.

Support for generated names which include the data type, dshape and name generated
by autoincrementing is retained by giving the `data` function a new default of
`name=generated`.

This patch also moves some tests from `test_interactive` to test suites
dedicated to `literal` and `data`.


blaze/expr/literal.py
```diff
@@ -34,11 +35,20 @@
     not_an_iterator.append(pymongo.database.Database)
 
 
-class Data(Symbol):
+class generate(object):
+    """A sentinel value, indicating whether or not `literal` should
+    should generate a name for the returned `Literal`.
```
llllllllll: should repeated

blaze/expr/tests/test_literal_expr.py
```diff
@@ -0,0 +1,89 @@
...
+
+def test_literal_name_repr():
+    assert "amounts" == repr(nl)
+
+
+def test_data_repr():
+    def replace_name(s):
+        # Account for autogenerated name.
+        return re.sub(r"name='(.*)'", "name='_0'", s)
+    expected = "<'tuple' data; _name='.+', dshape='2 * {name: string, amount: int64}'>" # noqa
```
llllllllll: two spaces before noqa, can you also put an empty line after the replace_name function?

blaze/expr/literal.py
```diff
@@ -84,7 +84,7 @@ class Literal(Symbol):
     """
     _arguments = 'data', 'dshape', '_name'
 
-    def __new__(cls, data, dshape, name=None):
+    def __new__(cls, data, dshape, name=None, _repr=False):
```
ehebert: I'm removing this.

ehebert: `This` being the added kwarg.

PR #1639, opened 2017-05-09
ENH: Allow to_tree to render a literal.

ehebert:
So that input literals which are not bound to a data resource, can be sent to
the server. e.g. a frozenset which is the `_keys` of an `IsIn` statement.

### DIFF
--- a/blaze/compute/__init__.py
+++ b/blaze/compute/__init__.py
@@ -1,3 +1,5 @@
 from __future__ import absolute_import, division, print_function
 
 from .core import compute, compute_up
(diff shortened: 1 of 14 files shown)


blaze/compute/__init__.py
```diff
@@ -1,3 +1,4 @@
 from __future__ import absolute_import, division, print_function
 
 from .core import compute, compute_up
+from . import literal  # noqa
```
ehebert: Is this to ensure that the dispatches are registered?

PR #1640, opened 2017-05-12
ENH: Allow an expr as the `_keys` of an IsIn.

ehebert:
The goal is to support SQL IsIn statements which use a `SELECT` from another
table as the definition of the `IN` group.

### DIFF
--- a/blaze/compute/sql.py
+++ b/blaze/compute/sql.py
@@ -1521,32 +1521,53 @@ def post_compute(_, s, **kwargs):
     return select(s)
 
 
-@dispatch(IsIn, ColumnElement, (Iterable, Selectable))
+@dispatch((Iterable, Selectable))
+def _coerce_op_input(i):
+    """Make input to SQLAlchemy operator an amenable type.
+
+    Parameters
+    ----------
+    i : (Iterable, Selectable)
+       The iterable or selectable to coerce.
+
+    Returns
+    -------
+    coerced_input : (Iterable, Selectable)
+       The iterable or selectable passed to the function.
+       These types are already the amenable types.
+    """
+    return i
+
+
+@dispatch(ColumnElement)
+def _coerce_op_input(i):
+    """Make input to SQLAlchemy operator an amenable type.
+
+    Parameters
+    ----------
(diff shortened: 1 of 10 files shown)


blaze/server/tests/test_server.py
```diff
@@ -694,6 +694,24 @@ def test_isin(test, serial):
...
+                       headers=mimetype(serial),
+                       data=serial.dumps(query))
+    expected = {'data': [False, True],
+                'names': ['name'],
+                'datashape': '2 * bool'}
+    assert result.status_code == RC.OK
+    resp = serial.loads(result.data)
+    assert list(serial.data_loads(resp['data'])) == expected['data']
+    assert list(resp['names']) == expected['names']
+    assert resp['datashape'] == expected['datashape']
```
ehebert: Will change.

blaze/compute/sql.py
```diff
@@ -1520,18 +1521,32 @@ def post_compute(_, s, **kwargs):
...
+
+
+@dispatch(IsIn, Selectable, ColumnElement)
+def compute_up(expr, data, keys, **kwargs):
     assert len(data.columns) == 1, (
         'only 1 column is allowed in a Select in IsIn'
     )
     col, = unsafe_inner_columns(data)
-    return reconstruct_select((col.in_(expr._keys),), data)
+    return col.in_(Select([keys]))
```
ehebert: This wasn't using `select` due to unfamiliarity on my part. I'm now aware of `compute.sql.select` and I'll use that.

Going to add a test case for `expr[expr > 5]).isin([1, 2, 3]` before resubmitting.

blaze/compute/tests/test_sql_compute.py
```diff
@@ -1691,6 +1692,118 @@ def test_isin():
     assert normalize(result_sql_expr) == normalize(expected)
 
 
+def test_not_isin_literal():
+    result = t[~t.name.isin(['foo', 'bar'])]
```
ehebert: While creating the failing test for an `isin` of a single column select, I noticed that this `result` should be `expr`. Will change.

PR #1643, opened 2017-05-22
fix doc errors

electronwill:
fix doc errors

### DIFF
--- a/blaze/expr/expressions.py
+++ b/blaze/expr/expressions.py
@@ -876,7 +876,7 @@ def binop_inputs(expr):
 class Coalesce(Expr):
     """SQL like coalesce.
 
-    .. code-block python
+    .. code-block:: python
 
         coalesce(a, b) = {
             a if a is not NULL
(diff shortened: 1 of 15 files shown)


llllllllll:
Why did you change all of the underline levels?

electronwill:
We're copying the Blaze docs into the overall Anaconda docs at https://docs.continuum.io/docs_oss/blaze/ so that customers with an offline installation of Anaconda Repository or Anaconda Enterprise Notebooks will have them. Until now, the Sphinx documentation build system has been producing lots of errors, partly because the underlines in these release notes are inconsistent with each other and with the Sphinx preferred underlines at https://docs.python.org/devguide/documenting.html#sections . This PR should fix all of these errors.

electronwill:
@postelrich , would you please merge this one or let me know what else I should do to make this ready to merge?

necaris:
@electronwill it looks like this is failing because of failures on master -- #1656 needs to be merged first before that will be good.

@postelrich note that these are _just_ docs changes so I'd have no objection to merging them despite CI failure -- I'll do this tomorrow unless you'd rather I don't.

electronwill:
Oh, OK, that makes sense. Thanks!

PR #1656, opened 2017-08-22
MAINT: Fix all tests

thequackdaddy:
Hello,

So I really like this project but it seems to have been quite quiet lately. 

I spent some time today un-pinning all the packages and getting the tests to pass again. One caveat is that `odo` has a [problem](https://github.com/blaze/odo/issues/567). For this, I've just `odo` from a repository I built. Obviously, this is not ideal. 

Let me know if this is helpful. This is an amazing set of tools.


thequackdaddy:
cc @postelrich

etc/requirements.txt
```diff
@@ -1,6 +1,6 @@
 flask>=0.10.1
 flask-cors
-odo>=0.5.0
+# odo>=0.5.1
```
postelrich: This would have to be uncommented before merge

thequackdaddy: I've made the pull request (see below). 

Once you change this, I'll revert this change and push the update. 

(For reasons I don't fully understand, pulling from my repo kept the version number below 0.5.0 so this kept failing :-/)

blaze/expr/datetime.py
```diff
@@ -131,14 +131,14 @@ def utcfromtimestamp(expr):
 
 
 class nanosecond(DateTime): _dtype = datashape.int64
-class week(DateTime): _dtype = datashape.int16
+class week(DateTime): _dtype = datashape.int64
```
ehebert: This change makes the tests pass; however, https://github.com/blaze/blaze/pull/1585 intentionally made the datetime types as narrow as possible.

I ran across your proposed fix while looking at the following failure when running the test `py.test -s -x blaze/compute/tests/test_postgresql_compute.py::test_datetime_access[week]` against current master of `blaze`, `odo`, and `datashape`:

```
obj = 'Attributes', message = 'Attribute "dtype" are different', left = dtype('int16')
right = dtype('int64'), diff = None

    def raise_assert_detail(obj, message, left, right, diff=None):
        if isinstance(left, np.ndarray):
            left = pprint_thing(left)
        if isinstance(right, np.ndarray):
            right = pprint_thing(right)
    
        msg = """{0} are different
    
    {1}
    [left]:  {2}
    [right]: {3}""".format(obj, message, left, right)
    
        if diff is not None:
            msg = msg + "\n[diff]: {diff}".format(diff=diff)
    
>       raise AssertionError(msg)
E       AssertionError: Attributes are different
E       
E       Attribute "dtype" are different
E       [left]:  int16
E       [right]: int64

/mnt/envs/blaze/lib/python3.4/site-packages/pandas/util/testing.py:1018: AssertionError
```

(It might make sense to break off making the aforementioned test pass into a separate issue/PR.)

ehebert: This patch may checks the values, but ignores the dtypes.

```
$ git diff
diff --git a/blaze/compute/tests/test_postgresql_compute.py b/blaze/compute/tests/test_postgresql_compe.
py                                                                                                     
index 6c15703..608e8fb 100644
--- a/blaze/compute/tests/test_postgresql_compute.py
+++ b/blaze/compute/tests/test_postgresql_compute.py
@@ -886,4 +886,5 @@ def test_datetime_access(attr, sql_with_dts):
         compute(expr, sql_with_dts, return_type=pd.Series),
         getattr(compute(s.A, sql_with_dts, return_type=pd.Series).dt, attr),
         check_names=False,
+        check_dtype=False,
     )
```

`compute(s.A, sql_with_dts, return_type=pd.Series).dt` is an instance of `pandas.tseries.common.DatetimeProperties` which is going to return a `Series` with a dtype of `int64` when accessing the `week` attribute.

ehebert: https://github.com/blaze/blaze/pull/1660 applies the patch in the previous comment.

blaze/compute/tests/test_mongo_compute.py
```diff
@@ -271,31 +271,36 @@ def test_datetime_handling(events):
...
-    assert result == [('Bob', 200.0, 600), ('Alice', 150.0, 300)]
+    assert ((result == [('Bob', 200.0, 600), ('Alice', 150.0, 300)]) or
+            (result == [('Alice', 150.0, 300), ('Bob', 200.0, 600)]))
 
 
 def test_summary_count(bank):
     expr = by(t.name, how_many=t.amount.count())
     result = compute(expr, bank)
-    assert result == [('Bob', 3), ('Alice', 2)]
+    assert ((result == [('Bob', 3), ('Alice', 2)]) or
```
llllllllll: We could just sort both the result and expectation

etc/requirements_ci.txt
```diff
@@ -1,38 +1,38 @@
...
-psycopg2==2.6.2
-PyHive==0.2.1
-pymongo==2.9.4
-PyMySQL==0.7.9
-pytest==3.0.3
-pytest-cov==2.0.0
-python-dateutil==2.5.3
-pytz==2016.7
-pyyaml==3.11
+bcolz
```
llllllllll: the point of this file was to pin the requirements we test with. Tests kept failing because downstream packages were being updated. I don't want to fix the whole world with every PR and it put a huge burden on whoever wanted to open the next PR.

requirements-rtd.txt
```diff
@@ -1,7 +1,7 @@
 numpy >= 1.7
 pandas >= 0.15.0
 git+git://github.com/blaze/datashape.git
-git+git://github.com/blaze/odo.git
+git+git://github.com/thequackdaddy/odo.git@reversion
```
llllllllll: ?

postelrich:
@thequackdaddy @llllllllll Two issues I've noticed in trying to get this PR in. One is starting at dask 0.15.2, `test_dask.py` tests start failing.  There is now a separate `dask-core` package starting at that version, don't know if its related. Second is the tests are failing for 3.4 only on travis mysql tests:

```
sqlalchemy.exc.InternalError: (pymysql.err.InternalError) (13, "Can't get stat of '/home/travis/build/blaze/blaze/blaze/examples/data/nyc.csv' (Errcode: 13 - Permission denied)") [SQL: 'LOAD DATA  INFILE %(path)s\n            INTO TABLE nyc\n            CHARACTER SET %(encoding)s\n            FIELDS\n                TERMINATED BY %(delimiter)s\n                ENCLOSED BY %(quotechar)s\n                ESCAPED BY %(escapechar)s\n            LINES TERMINATED BY %(lineterminator)s\n            IGNORE %(skiprows)s LINES\n            '] [parameters: {'quotechar': '"', 'escapechar': '\\', 'path': '/home/travis/build/blaze/blaze/blaze/examples/data/nyc.csv', 'lineterminator': '\n', 'skiprows': 1, 'encoding': 'utf8', 'delimiter': ','}]
../../../miniconda/envs/blaze/lib/python3.4/site-packages/pymysql/err.py:107: InternalError
```

Will continue to investigate.

ehebert:
@postelrich the MySQL `Can't get stat` error may be the only error that is keeping all current tests on master from passing. (See, https://travis-ci.org/blaze/blaze/jobs/288742182).

I had dug into this a little, and the following fixture in `test_mysql_compute` is failing with that fixture.
```
odo(
            example('nyc.csv'),
            'mysql+pymysql://%s@localhost/test::nyc' % getuser()
        )
```

The `LOAD DATA INFILE` command is performed by the mysql process as the `mysql` user. The `mysql` user does not have sufficient access to the `blaze` source checkout on the Travis server (nor does it have access on a local checkout.)
Should this be a `LOAD DATA LOCAL INFILE` command instead?

These tests appear to have used to work, and started failing with this commit. https://github.com/blaze/blaze/commit/e0e63677358f1547abb2a07ab41ea71a57de1bcd. The relevant change there is that the versions of requirements started to be constrained by `requirements_ci.txt`.

postelrich:
@ehebert i think its related to `LOAD DATA LOCAL` as well re: https://dba.stackexchange.com/questions/55960/mysql-import-csv-file-error-13-hy000-cant-get-stat-of-path-file-csv-errcod and similar questions. I checked and the file does have universal read perms. Looks like the last successful build was on this pr https://travis-ci.org/blaze/blaze/jobs/267031371. The build info doesn't state the mysql version but the current failing ones show `mysql  Ver 14.14 Distrib 5.6.33`. My guess is that travisci is now using a newer version that requires the `LOCAL` or something else.

postelrich:
Though on second though, how could it be the db if it works fine for 2.7 and 3.5.

postelrich:
@thequackdaddy @ehebert @llllllllll I'm at a loss for why only 3.4 is failing with that mysql error. I tried setting `LOCAL` as well as `set global local_infile=OK`. I also made sure they're all running the latest pymysql, yet it still fails. I'm open to suggestions.

Since we're adding 3.6, any objection to removing testing for 3.4? I see a lot of projects don't make builds for 3.4.

llllllllll:
We are still using 3.4 for our blaze servers at Quantopian. We don't use mysql so I would be fine with skipping those tests on 3.4 while we migrate to 3.6.

ehebert:
@llllllllll, I agree, skipping the mysql tests for 3.4 seems is a good interim solution.

postelrich:
@llllllllll @ehebert @necaris Someone want to review?

llllllllll:
@postelrich I will review this on monday

postelrich:
Also need to do an odo py3.6 release too.

etc/requirements_ci.txt
llllllllll: This comment still stands

llllllllll:
I am fine with everything except the change to requirements_ci.txt. in the past not installing the same test requirements has wasted so much time and discouraged people from contributing.

postelrich:
@llllllllll doesn't look like there are a lot of packages built for all 2.7, 3.4, 3.5, and 3.6. If you know a good way to come up with fixed dependencies, please go for it. Otherwise, we can get this PR merged in without pegged deps, and add it back in another PR when we drop 3.4 support.

PR #1660, opened 2017-10-12
TST: Fix failing datetime access tests.

ehebert:
The results of the expressions compared in the test_datetime_access test do not
share the same dtype.

The computation which does the attribute lookup as part of the expression uses
the dtypes specified in `blaze.expr.datetime`.

The computation which does not do the attribute lookup returns a
`DatetimeProperties` instance; attribute lookups on those instances are all
`int64`.

Fix by setting the check dtype flag to `False`.

Add expected dtypes to the tests fixture and assert that the computation which
does the attribute lookup returns the correct type.


llllllllll:
lgtm

PR #1661, opened 2017-10-16
BUG: Prevent truncation of datetime second attribute.

ehebert:
The `datetime.second` attribute was truncating to second resolution, instead of
including subsecond values.

i.e. `second` attribute of a datetime `2014-01-01 00:00:00.042` would return `0`
instead of `0.042`

Fix by changing the type of `datetime.Second` from `int64` to `float64`.

Returning a float was the behavior before the `compute_up` from `DateTime` ->
`ColumnElement` begain enforcing the specified type via a `CAST`.

### DIFF
--- a/blaze/compute/tests/test_postgresql_compute.py
+++ b/blaze/compute/tests/test_postgresql_compute.py
@@ -168,6 +168,26 @@ def sql_with_dts(url):
 
 
 @pytest.yield_fixture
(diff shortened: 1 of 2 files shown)


llllllllll:
lgtm

llllllllll:
test_datetime_access needs to be updated

ehebert:
Fixing `test_datetime_access` now.

PR #1671, opened 2018-07-31
WIP: separate non core backends

### LINKED ISSUE #1670: Separate backends from core
As discussed in #1669, want to split the backends into their own separately installable packages. This would allow blaze development to continue unhindered by the task of keeping the different backends up to date.

I propose a simple naming schema such as `blaze-{backend}` be chosen. The idea would be that a user could `pip/conda install blaze-sqlalchemy` and use it such as:

```
import blaze as bz
import blaze_sqlalchemy
bz.data(...)
```

This involves moving most of the modules in https://github.com/blaze/blaze/tree/master/blaze/compute. 

Modules to move:
- [ ] pandas
- [ ] sql
- [ ] numpy
- [ ] mongo
- [ ] bcolz
- [ ] csv
- [ ] json
- [ ] dask
- [ ] h5py
- [ ] hdfstore
- [ ] spark

Should trivial ones like csv or json be kept in core?

postelrich:
Resolves #1670 

This PR removes non-core backends from the blaze package.


blaze/__init__.py
```diff
@@ -76,29 +76,8 @@
...
 with ignoring(ImportError):
     from .compute.chunks import *
-with ignoring(ImportError):
-    from .compute.bcolz import *
-with ignoring(ImportError):
-    from .mongo import *
-    from .compute.mongo import *
-with ignoring(ImportError):
-    from .pytables import *
-    from .compute.pytables import *
```
dhirschfeld: This probably improves things enormously wrt import times:
```python
In [1]: %time import blaze as bz
Wall time: 4.6 s
```

I had wondered if some sort of lazy import mechanism might be useful, but with this change perhaps not?

If there's a plugin mechanism for backends though it may still be useful so users don't have to pay for what they're not using.

dhirschfeld: I had thought to use the [`setuptools` plugin capability](http://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins) but your solution is simpler and more explicit!

postelrich:
pandas-msgpack, which blaze server depends on, is a blocker since that package isn't capable with newer pandas. I made a quick fix but tests on that are failing for one of the builds.

postelrich:
Nope not working on this.

PR #1682, opened 2019-08-14
ENH: Adds ability to register hook to compute endpoint.

ernestoeperez88:


blaze/server/tests/test_server.py
```diff
@@ -129,6 +129,21 @@ def temp_server_with_excfmt(request):
         yield (client, stream)
 
 
+@pytest.yield_fixture(params=[None, tdata])
+def temp_server_with_custom_hook(request):
+    """Server with custom compute hook"""
+    data = request.param
+
+    def custom_compute_hook(expr):
```
llllllllll: Can we test passing some `compute_hook_kwargs` through here?

ernestoeperez88: Changed the test to set the status of the response based on the status passed to `compute_hook_kwargs`

